PR #46, opened 2024-06-07
added more verbose submitter installation instructions.

andymanaws:
Customers were failing to understand how to setup the submitter.
Some versions of C4D were also causing issues due to missing libraries.

Just updated the README file to elevate the instructions more verbosely.

Customer should be able to setup C4d submitters.

In consulation with service team.

This is a doc change.

NO
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,48 +34,8 @@ This library requires:
 
 This package provides a Cinema 4D plugin that creates jobs for AWS Deadline Cloud using the [AWS Deadline Cloud client library][deadline-cloud-client]. Based on the loaded scene it determines the files required, allows the user to specify render options, and builds an [OpenJD template][openjd] that defines the workflow.
 
-### Submitter Python environment 
-The submitter requires the following modules to be installed:
+For installation instructions, refer to [Submitter Environment Instructions](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/DEVELOPMENT.md#submitter-environment).
 
-```sh
-pip install deadline-cloud-for-cinema-4d
-pip install deadline[gui]
-```
-
-### Cinema 4D Plugin Extension Setup 
-
-The Cinema 4D plugin extension can be downloaded from the git repo:
-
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -34,6 +34,49 @@ This library requires:
...
+A missing library error will manifest in errors that can be visible from the **Python** section of the **Extensions > Console** UI. These typically look like:
+
+```
+PySide6/__init__.py: Unable to import Shiboken from  ...
+```
+
+To remedy these errors, you can switch to a later version of Cinema 4D which resolves the missing libraries, or you can manually add them specifically to the Cinema 4D python module, e.g.
+
+```
+& "C:\Program Files\Maxon Cinema 4D 2024\resource\modules\python\libs\win64\python.exe"-m ensurepip
```
karthikbekalp: I personally used these commands and I find these to be incredibly useful. Thanks for writing these! 

But I wonder if this should be in `DEVELOPMENT.md` instead or if we should just have all the instructions from that page over here? 
https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/DEVELOPMENT.md

For example, in blender (https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/DEVELOPMENT.md) and other repos we have installation instructions for submitters in `DEVELOPMENT.md`.

karthikbekalp (commented):
A little clarification regarding the location for installation instructions, but otherwise the PR looks great!

README.md
andymanaws: Sure - its a little detailed to be on the main page.

Can we add a line in 'Getting Started' to say something like 'For more detailed installation of the submitter and adaptor, please see the development.md' ....

karthikbekalp: Yup that makes sense to me.

crowecawcaw:
PR looks great! There are a couple checks are failing that are easy to fix:
1. The PR title or commit name needs to follow the Conventional Commit standard. You can add `docs: ` to the beginning of the PR title.
2. The commits need to be signed. There are instructions for signing the commits you've pushed here: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/46/checks?check_run_id=30059458571

Issue #55, opened 2024-06-25
Bug: Adaptor is not pathmapping the output path.


yuanmich2:
### Expected Behaviour

When using job attachments or when specifying a custom output path, the adaptor should overwrite any settings saved in the scene file. When using job attachments, the output path should be somewhere in the assetroot directory in the openjd session folder. When a custom output path and storage profile are used, the custom output path should be used.

### Current Behaviour

The render will always export to the path defined in the render settings.

### Reproduction Steps

Create a new scene. Define an output path that is accessible by your render job user in your render settings. Save your scene.

Bring up the submitter and change the output directory or submit to a queue that uses job attachments.

After the render completes, DCM will report no output files to download and your output will appear in the original output directory defined in the render settings.

### Code Snippet

n/a

yuanmich2:
It looks like the adaptor isn't setting the path at all.

These couple of lines set the frame range of the render by editing the render data: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/9f3813c65451f073cea4d256817dd2f927b906c0/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py#L42C1-L44C50

To edit the path, you'd just edit the RDATA_PATH key in the render data. Docs on RDATA_PATH here: https://developers.maxon.net/docs/py/2024_2_0/modules/c4d.documents/RenderData/index.html#c4d.documents.RenderData

yuanmich2:
Confirmed that adding a line like
`self.render_data[c4d.RDATA_PATH] = "C:/some/random/path/output_filename"`
Added to the section I linked in my last message successfully redirects the output. I'd make a PR, but I don't know how to extract the pathmapped output directory from the openjd description.

epmog:
This is partially resolved by https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/57

However, that's a quick fix to get the existing output paths in the scene path mapped. The longer term solution is for us to pass along the output directory and multipass directory through to the adaptor such that users can override the path from the submitter without being forced to save it in the scene file.

epmog:
To do the actual fix, we'll need to: 
1. Add the options to the submitter. Not sure how we want to present the multipass path. Optional checkbox to enable to match how C4D does it?
[...]

PR #57, opened 2024-06-27
fix: Add path mapping to outputs

evanspearman-a:

Output paths were not path mapped.

Call `map_path` on the output and multipass output paths in the adaptor.

Outputs will now be path mapped.

I installed the changes on a CMF instance and rendered a scene that had both normal and multi-pass outputs. I observed that I was able to download the output and that the logs recorded the path mapping happening.

No

No

----


evanspearman-a:
Partially fixes https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/55

epmog (approved):

ddneilson (approved):

Issue #53, opened 2024-06-25
Feature request: Submitter should force a scene save before submission


yuanmich2:
### Use Case

I often forget to save changes to my scene before hitting the submit button and would like the submitter to force me to save before the render job goes out.

I've attached a log of one example where I had never saved a scene so the job attachments failed to upload anything which resulted in a "scene does not exist" error.

[Deadline Cloud Farm_Customer Managed Jobs_session-027787349ca84e769f62f8a6e6b24acb.log](https://github.com/user-attachments/files/15974129/Deadline.Cloud.Farm_Customer.Managed.Jobs_session-027787349ca84e769f62f8a6e6b24acb.log)

### Proposed Solution

The submitter should be able to check if there are any unsaved changes and prompt the user to save before submission. This was the default behavior in Deadline 10.

karthikbekalp:
Closing this issue as we have merged a fix for it. https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/69

PR #69, opened 2024-10-11
feat: prompt save before submit (#53)

gp-dan:

#53 prompt save scene changes before submitting

Check for scene changes before submit dialog is shown and prompt user

Scene changes properly persisted before submit

C4D 2025

No

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -237,17 +237,42 @@ def _get_job_template(
     return job_template
 
 
+def _prompt_save_current_document():
+    doc = c4d.documents.GetActiveDocument()
+    if not doc.GetChanged():
+        # Document has no unsaved changed
+        return
+    if not c4d.gui.QuestionDialog("Save scene changes before submission?"):
+        # User selected No
+        return
+    filepath = doc.GetDocumentPath()
+    filename = doc.GetDocumentName()
+    if filepath:
+        # Document save path exists
+        save_path = os.path.join(filepath, filename) + ".c4d"
+    else:
+        # Prompt with Save As to set path for Untitled document
+        save_path = c4d.storage.SaveDialog(c4d.FILESELECTTYPE_ANYTHING, "Save As", "c4d")
+        # Handle user cancels document save
+        if not save_path:
+            return
+        # Set document path and name
+        path = os.path.dirname(save_path)
+        base_name = os.path.basename(save_path)
+        name, _ = os.path.splitext(base_name)
+        doc.SetDocumentPath(path)
+        doc.SetDocumentName(name)
+    # Save document to disk
+    c4d.documents.SaveDocument(doc, save_path, c4d.SAVEDOCUMENTFLAGS_0, c4d.FORMAT_C4DEXPORT)
+    # Ensure document is active
+    c4d.documents.InsertBaseDocument(doc)
+    # Update UI
+    c4d.EventAdd()
+
+
 def _show_submitter(parent=None, f=Qt.WindowFlags()):
 
-    if Scene.changed():
-        prompt = QtWidgets.QMessageBox.warning(
-            None,
-            "Deadline Cloud",
-            "Save scene changes before submission?",
-            QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel
-        )
-        if prompt == QtWidgets.QMessageBox.Ok:
-            Scene.save()
+    _prompt_save_current_document()
 
     render_settings = RenderSubmitterUISettings()
 
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -239,6 +239,16 @@ def _get_job_template(
 
 def _show_submitter(parent=None, f=Qt.WindowFlags()):
 
+    if Scene.changed():
+        prompt = QtWidgets.QMessageBox.warning(
+            None,
+            "Deadline Cloud",
+            "Save scene changes before submission?",
+            QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel
+        )
+        if prompt == QtWidgets.QMessageBox.Ok:
+            Scene.save()
+
     render_settings = RenderSubmitterUISettings()
 
     # Set the setting defaults that come from the scene
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -245,11 +245,11 @@ def _prompt_save_current_document():
     if not c4d.gui.QuestionDialog("Save scene changes before submission?"):
         # User selected No
         return
-    filepath = doc.GetDocumentPath()
-    filename = doc.GetDocumentName()
-    if filepath:
+    file_path = doc.GetDocumentPath()
+    file_name = doc.GetDocumentName()
+    if file_path:
         # Document save path exists
-        save_path = os.path.join(filepath, filename) + ".c4d"
+        save_path = os.path.join(file_path, file_name)
     else:
         # Prompt with Save As to set path for Untitled document
         save_path = c4d.storage.SaveDialog(c4d.FILESELECTTYPE_ANYTHING, "Save As", "c4d")
@@ -257,11 +257,10 @@ def _prompt_save_current_document():
         if not save_path:
             return
         # Set document path and name
-        path = os.path.dirname(save_path)
+        doc_path = os.path.dirname(save_path)
         base_name = os.path.basename(save_path)
-        name, _ = os.path.splitext(base_name)
-        doc.SetDocumentPath(path)
-        doc.SetDocumentName(name)
+        doc.SetDocumentPath(doc_path)
+        doc.SetDocumentName(base_name)
     # Save document to disk
     c4d.documents.SaveDocument(doc, save_path, c4d.SAVEDOCUMENTFLAGS_0, c4d.FORMAT_C4DEXPORT)
     # Ensure document is active
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -255,7 +255,10 @@ def _prompt_save_current_document():
         save_path = c4d.storage.SaveDialog(c4d.FILESELECTTYPE_ANYTHING, "Save As", "c4d")
         # Handle user cancels document save
         if not save_path:
-            return
+            c4d.gui.MessageDialog(
+                "Submission canceled. File must be saved to disk before submission."
+            )
+            return False
         # Set document path and name
         doc_path = os.path.dirname(save_path)
         base_name = os.path.basename(save_path)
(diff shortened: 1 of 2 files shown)


leongdl (dismissed):
Thank you for the contribution!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -239,6 +239,16 @@ def _get_job_template(
 
 def _show_submitter(parent=None, f=Qt.WindowFlags()):
 
+    if Scene.changed():
```
leongdl: Side question for the deadline team, should there be a default setting to always save?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -239,6 +239,16 @@ def _get_job_template(
...
+        prompt = QtWidgets.QMessageBox.warning(
+            None,
+            "Deadline Cloud",
+            "Save scene changes before submission?",
+            QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel
+        )
+        if prompt == QtWidgets.QMessageBox.Ok:
+            Scene.save()
```
jusiskin: Should we `return` here if the user picks cancel?

Looking back at the source issue (#53), the same problem would occur:

>   I had never saved a scene so the job attachments failed to upload anything which resulted in a "scene does not exist" error.

Another possible outcome is that the job is submitted with the previously-saved copy of the scene while the scene has unsaved changes.

jusiskin (changes requested):
Thanks for the contribution here. This looks like it will help solve an important issue with the Cinema4D submitter.

I have a small suggestion to improve this even further.

jusiskin:
@gp-dan , it also looks like the "DCO" pull request check failed. Apologies, it looks like the `CONTRIBUTING.md` documentation in this repository is not complete.

Please refer to the contributing documentation in the `deadline-cloud-for-maya` repository as a reference:

https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/CONTRIBUTING.md#contributing-via-pull-requests

This documents how to sign-off the commit. Our team plans to eventually migrate that same documentation over to this repository to better help contributors in the future.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: I agree with @jusiskin's suggestion to return. Additionally, I propose that we should update the prompt to suggest that the submitter quits if user picks cancel. 


Suggested prompt: "Press 'OK' to save scene changes before submission. Selecting 'Cancel' will stop the submission."

gp-dan: thanks @jusiskin 
This is consistent with the UX flow of other submitters such as [maya](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/maya_render_submitter.py#L534). The prompt is a narrow choice to save or ignore any changes since the last save.

gp-dan:
@jusiskin to pass DCO should I force push a signed off commit here or create a new PR?

jusiskin:
@gp-dan, yes you can force-push to this same branch. No need for another pull request.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
jusiskin: Oh, thanks for surfacing that Maya behaves this way. Consistency aside, I'm trying to think of scenarios where that behavior makes sense.

What are your thoughts?

gp-dan: Hi @jusiskin  

FYI - The Houdini submitter also follows this warning style UX flow. 

I think guiding the user to save with a warning is better than forcing the user to save. Sometimes saving a big scene can take a long time and the user knows the changes since last save will have no effect on the result so they want to continue on to submit without further friction. Another issue is GetChanged() can return false positives in some cases and the user wants to skip past the save. 

From what I can tell most other render farm products use this flow so it shouldn't surprise the typical DCC user.

It's ultimately a subjective user experience preference about what the product should value more: strictness or flexibility.

I'm happy to revise the code based on consensus of the maintainers. Thanks!

jusiskin: @gp-dan that makes sense, thanks for the context. I also realize that this function only displays the submitter dialog and doesn't submit it. So the person using this can still cancel their submission by closing the submitter dialog.

On that basis, this behavior seems okay to me.

jusiskin:
@gp-dan, this looks good to go. We just need the DCO signed-off commit to be pushed.

karthikbekalp (dismissed):
Looks good to me, should be safe to merge.

jusiskin (dismissed):

jusiskin:
@gp-dan, it looks like the [format linter failed](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/actions/runs/11394608038/job/31736473957?pr=69) the pull request checks:

<details>
    <summary>Lint failure message</summary>

```
--- /home/runner/work/deadline-cloud-for-cinema-4d/deadline-cloud-for-cinema-4d/src/deadline/cinema4d_submitter/scene.py	2024-10-18 14:09:54.026706+00:00
+++ /home/runner/work/deadline-cloud-for-cinema-4d/deadline-cloud-for-cinema-4d/src/deadline/cinema4d_submitter/scene.py	2024-10-18 14:10:17.155592+00:00
@@ -99,15 +99,18 @@
would reformat /home/runner/work/deadline-cloud-for-cinema-4d/deadline-cloud-for-cinema-4d/src/deadline/cinema4d_submitter/scene.py
 
 class Scene:
     """
     Functionality for retrieving settings from the active scene
     """
+
     @staticmethod
     def save() -> bool:
         """Save active document"""
-        return c4d.documents.SaveDocument(doc, doc.GetDocumentName(), c4d.SAVEDOCUMENTFLAGS_DIALOGSALLOWED, c4d.FORMAT_C4DEXPORT)
+        return c4d.documents.SaveDocument(
+            doc, doc.GetDocumentName(), c4d.SAVEDOCUMENTFLAGS_DIALOGSALLOWED, c4d.FORMAT_C4DEXPORT
+        )
 
     @staticmethod
     def changed() -> bool:
         """
         Returns True if Active scene has unsaved changes
--- /home/runner/work/deadline-cloud-for-cinema-4d/deadline-cloud-for-cinema-4d/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py	2024-10-18 14:09:54.026706+00:00
+++ /home/runner/work/deadline-cloud-for-cinema-4d/deadline-cloud-for-cinema-4d/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py	2024-10-18 14:10:17.233799+00:00
@@ -242,11 +242,11 @@
     if Scene.changed():
         prompt = QtWidgets.QMessageBox.warning(
             None,
would reformat /home/runner/work/deadline-cloud-for-cinema-4d/deadline-cloud-for-cinema-4d/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
             "Deadline Cloud",
             "Save scene changes before submission?",
-            QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel
+            QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel,
         )
         if prompt == QtWidgets.QMessageBox.Ok:
             Scene.save()
 
     render_settings = RenderSubmitterUISettings()

Oh no! 💥 💔 💥
2 files would be reformatted, 25 files would be left unchanged.
```

</details>

These are blocking the merge. The simplest way to fix is to run:

```sh
hatch run fmt
```

then:

```sh
# Add the formatted files
git add ...

# Amend your commit
git commit --amend

# Force push the amended commit
git push --force-with-lease <remote_name> feat-prompt-save-before-submit-gh53
```

karthikbekalp:
@gp-dan and I tried manually testing this feature and it looks like the scene is saved after and not before the submission. We would have to fix that before we merge this PR.

gp-dan:
@karthikbekalp I've just pushed a commit to address the bug you noticed. I've simplified the feature to a single function. It uses c4d.gui instead of qt message box and handles cases where the scene is untitled or previously saved with unsaved changes. @jusiskin I ran the lint function before submitting. To make the review easier I've kept the previous commit instead of force pushing, If this looks good you can rebase and squash to get rid of the superfluous previous commit. Thanks

karthikbekalp:
> @karthikbekalp I've just pushed a commit to address the bug you noticed. I've simplified the feature to a single function. It uses c4d.gui instead of qt message box and handles cases where the scene is untitled or previously saved with unsaved changes. @jusiskin I ran the lint function before submitting. To make the review easier I've kept the previous commit instead of force pushing, If this looks good you can rebase and squash to get rid of the superfluous previous commit. Thanks

Thanks for your contribution. I checked with the latest changes in the PR but I still face the issue with saving the scene. :(

karthikbekalp:
Thanks for fixing this @gp-dan! I checked the changes locally and I can confirm that it works!

karthikbekalp (dismissed):
After running this locally, I can confirm that the fix works! Thanks for the contribution.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -237,8 +237,42 @@ def _get_job_template(
...
+    file_name = doc.GetDocumentName()
+    if file_path:
+        # Document save path exists
+        save_path = os.path.join(file_path, file_name)
+    else:
+        # Prompt with Save As to set path for Untitled document
+        save_path = c4d.storage.SaveDialog(c4d.FILESELECTTYPE_ANYTHING, "Save As", "c4d")
+        # Handle user cancels document save
+        if not save_path:
+            return
```
jusiskin: If the program goes to line 258, does that mean the scene is not saved at all? Wouldn't that guarantee that the submission will not work? If so, maybe it would be better to fail fast than proceed to the submission dialog?

jusiskin (changes requested):
I have one more suggestion based on the revisions.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gp-dan: Submission would work but render would fail. We can short circuit at this point.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -237,6 +241,41 @@ def _get_job_template(
...
 
 
+def _prompt_save_current_document():
+    doc = c4d.documents.GetActiveDocument()
+    if not doc.GetChanged():
+        # Document has no unsaved changed
+        return
+    if not c4d.gui.QuestionDialog("Save scene changes before submission?"):
+        # User selected No
+        return
```
karthikbekalp: Missing return? Same above and at the end of this function.

karthikbekalp (changes requested):
I think there are some missing returns that need to be fixed. :)

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for fixing this.

karthikbekalp (approved):
Thanks the changes now look good to me.

jusiskin (approved):

Issue #51, opened 2024-06-25
Bug: Adaptor doesn't recognize "CRITICAL stop" error


yuanmich2:
### Expected Behaviour

I expect the job status in the DCM to accurately reflect when a render has failed.

### Current Behaviour

I just ran a job where the logs reported a "CRITICAL stop" error, but the DCM marked the job as a success. The fix would probably just require adjusting the way we parse the log output of C4D.

I've attached a log of the job.
[Deadline Cloud Farm_Customer Managed Jobs_session-7faa4fd9cf2a4dfeabd65ee9a1c2ac25.log](https://github.com/user-attachments/files/15973717/Deadline.Cloud.Farm_Customer.Managed.Jobs_session-7faa4fd9cf2a4dfeabd65ee9a1c2ac25.log)

### Reproduction Steps

I am not sure how to reproduce the issue, but I experience this issue on every C4D render in my farm.

### Code Snippet

n/a

yuanmich2:
To add some context, this error pops up when the user running the render doesn't have access to the output directory.

The PR to fix this issue would just append a regex that matches the error logs I attached to these few lines in the adaptor: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/9e00966d9e55fa081505c6cf3c326090f935400d/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py#L192C1-L194C87

karthikbekalp:
This issue should have been fixed by https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/72

PR #72, opened 2024-10-17
fix: handle CRITICAL error in adaptor regex callbacks (#51)

gp-dan:

#51 DCM job status would show success when task failed due to CRITICAL error

Add regex callback to handle CRITICAL log messages

Augment adaptor error_regexes pattern

By manually testing the regex against [log file in the issue ](https://github.com/user-attachments/files/15973717/Deadline.Cloud.Farm_Customer.Managed.Jobs_session-7faa4fd9cf2a4dfeabd65ee9a1c2ac25.log)

No

No
----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -191,7 +191,9 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
 
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
-            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)]
+            error_regexes = [
+                re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)
+            ]
 
             callback_list.append(RegexCallback(completed_regexes, self._handle_complete))
             callback_list.append(RegexCallback(progress_regexes, self._handle_progress))
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -191,7 +191,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
 
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
-            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*", re.IGNORECASE)]
+            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)]
 
             callback_list.append(RegexCallback(completed_regexes, self._handle_complete))
             callback_list.append(RegexCallback(progress_regexes, self._handle_progress))
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -197,8 +197,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
 
             callback_list.append(RegexCallback(completed_regexes, self._handle_complete))
             callback_list.append(RegexCallback(progress_regexes, self._handle_progress))
-            if self.init_data.get("strict_error_checking", False):
-                callback_list.append(RegexCallback(error_regexes, self._handle_error))
+            callback_list.append(RegexCallback(error_regexes, self._handle_error))
 
             callback_list.append(
                 RegexCallback(
--- a/test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
+++ b/test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
@@ -40,3 +40,20 @@ class TestCinema4DAdaptor_on_cleanup:
         # THEN
         assert match is not None
         assert str(adaptor._exc_info) == f"Cinema4D Encountered an Error: {stdout}"
+
+    def test_handle_error(self, init_data: dict) -> None:
+        """Tests that the _handle_error method throws a error runtime error correctly"""
+        # GIVEN
+        adaptor = Cinema4DAdaptor(init_data)
+
+        # WHEN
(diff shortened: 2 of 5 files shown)


src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -191,7 +191,7 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
 
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
-            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*", re.IGNORECASE)]
+            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)]
```
karthikbekalp: Can we add unit tests to confirm that any CRITICAL / ERROR in the log stops the adaptor?

karthikbekalp (changes requested):
Thanks for your contribution. 

Looks good to me although we need a few tests to confirm the behavior.

test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
```diff
@@ -0,0 +1,42 @@
...
+class TestCinema4DAdaptor_on_cleanup:
+
+    def test_handle_critical(self, init_data: dict) -> None:
+        """Tests that the _handle_error method throws a critical runtime error correctly"""
+        # GIVEN
+        adaptor = Cinema4DAdaptor(init_data)
+
+        # WHEN
+        error_regex = re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*")
+        stdout = "CRITICAL: Stop [ge_file.cpp(1172)]"
```
karthikbekalp, line 34: Nit: Can we add a test for error as well?

karthikbekalp (dismissed):
Thanks for your contribution! I see that the SonarCloud check has failed. I believe this should be harmless considering DoS is not possible as we are only matching with logs generated by the worker agent.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
karthikbekalp: Thanks for adding the test for Critical errors.

karthikbekalp:
The string matching process, while not entirely user-controlled, could potentially take a considerable amount of time for extremely large strings in worst-case scenarios.

I will consult with my colleagues to determine whether we should override this recommendation, or implement timeouts to halt the matching process if the regex handler fails to produce results after a predetermined period (e.g., one second).

test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
karthikbekalp: Thanks for adding the extra case!

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -191,12 +191,13 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
...
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
-            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*", re.IGNORECASE)]
+            error_regexes = [
+                re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)
+            ]
 
             callback_list.append(RegexCallback(completed_regexes, self._handle_complete))
             callback_list.append(RegexCallback(progress_regexes, self._handle_progress))
-            if self.init_data.get("strict_error_checking", False):
```
karthikbekalp, line 198: Hmm, can we modify this to return True by default instead? If customers were to pass `strict_error_checking` = False and still find renders failing that would be unexpected no? 

Also this is a different behaviour from other adaptors like [keyshot](https://github.com/aws-deadline/deadline-cloud-for-keyshot/blob/mainline/src/deadline/keyshot_adaptor/KeyShotAdaptor/adaptor.py#L208) or [blender](https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/src/deadline/blender_adaptor/BlenderAdaptor/adaptor.py#L196).

karthikbekalp (changes requested):
Thanks for the contribution, the changes look good but just asking for a clarification for 1 changed behaviour.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -191,12 +191,13 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
 
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
-            error_regexes = [re.compile(".*Error: .*|.*\\[Error\\].*", re.IGNORECASE)]
+            error_regexes = [
+                re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)
```
karthikbekalp, line 195: Also I wonder if we should add regex for [CRITICAL] as well similar to [ERROR] pattern checking?

 Something like `.*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*|.*\\[Critical\\].*`? Because if the log line does not have `CRITICAL: ` (with space at the end) then this would not fail the job.

karthikbekalp: Hmm, never mind. I searched for logs with [CRITICAL] and I couldn't find any.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
gp-dan: The current c4d adaptor [init schema](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/schemas/init_data.schema.json) doesn't support strict_error_checking so customers can't pass it. Nuke and Houdini adaptors don't support it either. My motivation to remove this line is it's the smallest impact change that will solve #51 without bringing in complexity from other issues. You can add support for strict_error_checking but it requires impact on other modules and further testing.

karthikbekalp: Ok makes sense. Thanks for cleaning this up!

karthikbekalp (approved):
Thanks for your contributions. I've also tried this locally and it works!

moorec-aws (approved):

PR #59, opened 2024-06-29
feat!: Updates to work towards a GA release

Ahuge:
### What was the problem/requirement? (What/Why)
`deadline-cloud-for-cinema-4d` is an "alpha release" and during my testing trying to get this up and running on a customer managed fleet on AWS I have had to make several changes to get it working correctly.
### What was the solution? (How)
Each of these commits are aimed to be mostly separate features or fixes however in my testing some of them do build on others.
- [Add COMMANDLINE_EXECUTABLE env var](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/commit/0c261e4efb3600cd4a554164417dc2aeeb6e8302) This implements optional environment variable overrides to point to a custom "Commandline" installation.
- [Forcefully add the dll path for C4D python](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/commit/a5dbdb10519b4b436a80db75335fcaad19a542f1) Implement some DLL patching required for versions of C4D below 2024.2. See https://github.com/danbradham/wheels/issues/4#issuecomment-1772721170
- [Creating an adaptor script wrapper for Linux](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/commit/74da6c2e121f9cccdabd4b5a87c2bb11a48368ad) On Linux, C4D requires environment variables set up and they provide a `setup_c4d_env` script which mostly works. We've built an `adaptor.sh` for Linux to run with will apply the `setup_c4d_env` script. This also introduces some environment variable changes that I found from Deadline10.
- [Include a patch for versions below 2024.4](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/commit/814dc9e77f032b69a34e93097c9290434931cce2) This patch will add `/usr/lib64` to the LD_LIBRARY path if the C4D version is below 2024.4. Versions lower than this had an issue where `c4dplugin.xso64` required libwebkit2gtk even in a headless `-nogui` mode.
- [Add a better safety around the os.environ call](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/commit/8e0d1e18810be185c1469b8cec633c985a27a474) A quick fix around querying for an environment variable that may be None.

### How was this change tested?
[...]

PR #73, opened 2024-10-18
feat!: add adaptor output path mapping and overrides (#55)

gp-dan:

#55 adaptor not path mapping or overriding output path

- Add get_output_paths() to Scene to return expanded default and multi-pass paths
- Add output and multi-pass paths to parameters, init data, ui and adaptor actions
- Remove DEADLINE_PYTHON_PATH to ensure proper loading of win32 modules

User can override output path and asset outputs are detected and downloadable

- Submitted scene with default settings Override Output Path disabled, the path was mapped by adaptor and files downloaded
- Submitted scene with Override Output Path enabled and augmented to a different name, the path was mapped by adaptor and files downloaded
 
No

Potentially -- logic relying on DEADLINE_PYTHON_PATH is removed so this may impact installation instructions 

Manually adding modules from an arbitrary DEADLINE_PYTHON_PATH location failed to properly boostrap more complex modules like win32 with extra DLLs (even when calling os.add_dll_directory)

Installing the modules to a location expected by the embedded python (such as `C:\Users\<user>\AppData\Roaming\Maxon\python\python311\libs`) solves this issue and makes the plugin a more idiomatic Cinema4D extension.

----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -38,15 +38,14 @@ hatch run all:test
 
 ## Submitter environment
 
-Cinema4D does not support PYTHONPATH. We set DEADLINE_CLOUD_PYTHONPATH which the
-submitter and adaptor uses to set sys.path explictly and load deadline modules.
+Install deadline in a [supported search path](https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#python-interpreter-bound-search-paths) or set a [custom interpreter path](https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#custom-interpreter-bound-search-paths) to the deadline install location.
 
 - install deadline-cloud with pyside
 - set env below
 
 ```
 # deadline-cloud lib with pyside
-export DEADLINE_CLOUD_PYTHONPATH="/path/to/deadline-cloud/site-packages"
+export C4DPYTHONPATH311 "/path/to/deadline-cloud/site-packages"
 # configure cinema4d to find extension entry point
 export g_additionalModulePath="/path/to/deadline-cloud-for-cinema4d/deadline_cloud_extension"
 ```
@@ -56,8 +55,7 @@ export g_additionalModulePath="/path/to/deadline-cloud-for-cinema4d/deadline_clo
 
 ## Worker adaptor environment
 
-Cinema4D does not support PYTHONPATH. We set DEADLINE_CLOUD_PYTHONPATH which the
-adaptor uses to set sys.path explictly and load deadline modules.
+Install deadline in a [supported search path](https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#python-interpreter-bound-search-paths) or set a [custom interpreter path](https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#custom-interpreter-bound-search-paths) to the deadline install location.
 
 ### Linux
 
@@ -68,16 +66,6 @@ client.
 Example linux env below:
 
 ```
-export DEADLINE_CLOUD_PYTHONPATH="/tmp/lib/python3.11/site-packages"
+export C4DPYTHONPATH311 "/path/to/deadline-cloud/site-packages"
 export DEADLINE_CINEMA4D_EXE="/opt/maxon/cinema4dr2024.200/bin/c4d"
-```
-
-### Windows
-
-To run the adaptor on Windows, you'll have to configure the environment variable `DEADLINE_CLOUD_PYTHONPATH` (like the submitter above) and install pywin32 into Cinema4D's python. Example:
-
-```
-set DEADLINE_CLOUD_PYTHONPATH="C:\path\to\deadline-cloud\site-packages"
-"C:\Program Files\Maxon Cinema 4D 2024\resource\modules\python\libs\win64\python.exe" -m ensurepip   
-"C:\Program Files\Maxon Cinema 4D 2024\resource\modules\python\libs\win64\python.exe" -m pip install pywin32  
 ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -45,7 +45,7 @@ Install deadline in a [supported search path](https://developers.maxon.net/docs/
 
 ```
 # deadline-cloud lib with pyside
-export C4DPYTHONPATH311 "/path/to/deadline-cloud/site-packages"
+export C4DPYTHONPATH311="/path/to/deadline-cloud/site-packages"
 # configure cinema4d to find extension entry point
 export g_additionalModulePath="/path/to/deadline-cloud-for-cinema4d/deadline_cloud_extension"
 ```
@@ -66,6 +66,6 @@ client.
 Example linux env below:
 
 ```
-export C4DPYTHONPATH311 "/path/to/deadline-cloud/site-packages"
+export C4DPYTHONPATH311="/path/to/deadline-cloud/site-packages"
 export DEADLINE_CINEMA4D_EXE="/opt/maxon/cinema4dr2024.200/bin/c4d"
 ```
--- a/deadline_cloud_extension/DeadlineCloud.pyp
+++ b/deadline_cloud_extension/DeadlineCloud.pyp
@@ -4,12 +4,7 @@ import sys
 root = os.path.dirname(__file__)
 sys.path.insert(0, os.path.join(root, 'modules'))
 
-if 'cinema4d_submitter' not in sys.modules.keys():
-    python_path = os.getenv('DEADLINE_CLOUD_PYTHONPATH')
-    python_paths = python_path.split(os.pathsep)
-    for n in python_paths:
-        if n not in sys.path:
-            sys.path.append(n)
+if 'deadline.cinema4d_submitter' not in sys.modules.keys():
     import deadline.cinema4d_submitter
 else:
     import importlib
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -26,7 +26,7 @@ class Cinema4DNotRunningError(Exception):
     pass
 
 
-_FIRST_CINEMA4D_ACTIONS = ["scene_file", "take"]
+_FIRST_CINEMA4D_ACTIONS = ["scene_file", "take", "output_path", "multi_pass_path"]
 _CINEMA4D_RUN_KEYS = {
     "frame",
 }
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -70,6 +70,8 @@ def _get_parameter_values(
 
     # Set the c4d scene file value
     parameter_values.append({"name": "Cinema4DFile", "value": Scene.name()})
+    parameter_values.append({"name": "OutputPath", "value": settings.output_path})
+    parameter_values.append({"name": "MultiPassPath", "value": settings.multi_pass_path})
 
     if settings.override_frame_range:
         frame_list = settings.frame_list
@@ -168,7 +170,7 @@ def _get_job_template(
         else:
             # Update the init data of the step
             init_data = step["stepEnvironments"][0]["script"]["embeddedFiles"][0]
-            init_data["data"] = "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'" % take_data.name
+            init_data["data"] = "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '{{Param.OutputPath}}'\nmulti_pass_path: '{{Param.MultiPassPath}}'" % take_data.name
 
     # If Arnold is one of the renderers, add Arnold-specific parameters
     if "arnold" in renderers:
@@ -244,10 +246,12 @@ def _show_submitter(parent=None, f=Qt.WindowFlags()):
     # Set the setting defaults that come from the scene
     render_settings.name = Path(Scene.name()).name
     render_settings.frame_list = str(Animation.frame_list())
-    render_settings.output_path = Scene.output_path()
+    default_path, multi_path = Scene.get_output_paths()
+    render_settings.output_path = default_path
+    render_settings.multi_pass_path = multi_path
 
     # Load the sticky settings
-    render_settings.load_sticky_settings(Scene.name())
+    # render_settings.load_sticky_settings(Scene.name())
 
     doc = c4d.documents.GetActiveDocument()
     take_data = doc.GetTakeData()
@@ -308,6 +312,12 @@ def _show_submitter(parent=None, f=Qt.WindowFlags()):
         if settings.take_selection == TakeSelection.CURRENT:
             submit_takes = current_data_list
 
+        # Add overrides to asset references
+        if settings.output_path:
+            asset_references.output_directories.add(os.path.dirname(settings.output_path))
+        if settings.multi_pass_path:
+            asset_references.output_directories.add(os.path.dirname(settings.multi_pass_path))
+
         # # Check if there are multiple frame ranges across the takes
         first_frame_range = submit_takes[0].frame_range
         per_take_frames_parameters = not settings.override_frame_range and any(
@@ -355,7 +365,6 @@ def _show_submitter(parent=None, f=Qt.WindowFlags()):
 
     for take_data in take_data_list:
         auto_detected_attachments.output_directories.update(take_data.output_directories)
-
     attachments = AssetReferences(
         input_filenames=set(render_settings.input_filenames),
         input_directories=set(render_settings.input_directories),
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -170,7 +170,10 @@ def _get_job_template(
         else:
             # Update the init data of the step
             init_data = step["stepEnvironments"][0]["script"]["embeddedFiles"][0]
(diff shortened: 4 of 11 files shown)


deadline_cloud_extension/DeadlineCloud.pyp
```diff
@@ -4,12 +4,7 @@ import sys
 root = os.path.dirname(__file__)
 sys.path.insert(0, os.path.join(root, 'modules'))
 
-if 'cinema4d_submitter' not in sys.modules.keys():
-    python_path = os.getenv('DEADLINE_CLOUD_PYTHONPATH')
```
karthikbekalp, line 8: Considering this has an impact on the environment variable that we set, can we update the instructions as well? I mean [over here](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/f26c9ac7f213f6be9719226bfcc8965caf328d6c/DEVELOPMENT.md?plain=1#L41)

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -26,7 +26,7 @@ class Cinema4DNotRunningError(Exception):
     pass
 
 
-_FIRST_CINEMA4D_ACTIONS = ["scene_file", "take"]
+_FIRST_CINEMA4D_ACTIONS = ["scene_file", "take", "output_path", "multi_pass_path"]
```
karthikbekalp, line 29: Nit: Should it be `multi_pass_output_path` instead so that it signifies that its an output path?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -244,10 +246,12 @@ def _show_submitter(parent=None, f=Qt.WindowFlags()):
...
     render_settings.name = Path(Scene.name()).name
     render_settings.frame_list = str(Animation.frame_list())
-    render_settings.output_path = Scene.output_path()
+    default_path, multi_path = Scene.get_output_paths()
+    render_settings.output_path = default_path
+    render_settings.multi_pass_path = multi_path
 
     # Load the sticky settings
-    render_settings.load_sticky_settings(Scene.name())
+    # render_settings.load_sticky_settings(Scene.name())
```
karthikbekalp: Hmm, any reason we commented this line out? Is it not required anymore?

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -170,6 +174,43 @@ def get_render_data(doc=None, take=None):
...
 
+    @staticmethod
+    def get_output_paths(take=None) -> str:
+        doc = c4d.documents.GetActiveDocument()
+        doc_path = doc.GetDocumentPath()
+        if not take:
+            take_data = doc.GetTakeData()
+            take = take_data.GetCurrentTake()
+        render_data = Scene.get_render_data(doc=doc, take=take)
+        rbc = render_data.GetDataInstance()
```
karthikbekalp: Nit: Can we have better variables names? It is not intuitive as what rbc or rpd mean. :)

karthikbekalp (changes requested):
Thanks for your contribution! On the whole this PR looks really good, but I've added few comments that require some clarifications/improvements. 

Also, we need to add unit tests for this.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gp-dan: My mistake, thanks for catching this

src/deadline/cinema4d_submitter/scene.py
gp-dan: I agree names could be better but they match the dict key names in the c4d api here which also uses [_rBc](https://developers.maxon.net/docs/py/2024_0_0a/modules/c4d.modules/tokensystem/index.html#:~:text=_rData%3A%20RenderData-,_rBc,-%3A%20BaseContainer) - I could rename it `render_base_container_instance` ?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: No problem.

src/deadline/cinema4d_submitter/scene.py
karthikbekalp: That would make more sense, thanks!

DEVELOPMENT.md
```diff
@@ -38,15 +38,14 @@ hatch run all:test
...
-submitter and adaptor uses to set sys.path explictly and load deadline modules.
+Install deadline in a [supported search path](https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#python-interpreter-bound-search-paths) or set a [custom interpreter path](https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#custom-interpreter-bound-search-paths) to the deadline install location.
 
 - install deadline-cloud with pyside
 - set env below
 
 ```
 # deadline-cloud lib with pyside
-export DEADLINE_CLOUD_PYTHONPATH="/path/to/deadline-cloud/site-packages"
+export C4DPYTHONPATH311 "/path/to/deadline-cloud/site-packages"
```
karthikbekalp: Missing `=`? Also this may have some merge conflicts with[ this PR](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/59/files).  Same issue below

DEVELOPMENT.md
```diff
@@ -68,16 +66,6 @@ client.
 Example linux env below:
 
 ```
-export DEADLINE_CLOUD_PYTHONPATH="/tmp/lib/python3.11/site-packages"
+export C4DPYTHONPATH311 "/path/to/deadline-cloud/site-packages"
 export DEADLINE_CINEMA4D_EXE="/opt/maxon/cinema4dr2024.200/bin/c4d"
-```
-
-### Windows
```
karthikbekalp, line 75: Hmm, so we don't need to manually install pywin32 anymore? But wouldn't this raise module not found errors by the adaptor?

karthikbekalp (changes requested):
Thanks for your contribution! Added a few comments for clarification.

deadline_cloud_extension/DeadlineCloud.pyp
karthikbekalp: Thanks for updating!

DEVELOPMENT.md
karthikbekalp: Also any reason this is suffixed with 311? I was able to get it working on a machine with Python 3.12 as well.

deadline_cloud_extension/DeadlineCloud.pyp
```diff
@@ -4,12 +4,7 @@ import sys
...
-    python_paths = python_path.split(os.pathsep)
-    for n in python_paths:
```
karthikbekalp, line 10: I was testing this and I got errors like these 

```
Traceback (most recent call last):
  File "D:\C4D_Submitter_Externsion\DeadlineCloud.pyp", line 8, in <module>
ModuleNotFoundError:No module named 'deadline'
```

DEVELOPMENT.md
gp-dan: Yes, thanks, missing = is a typo.
Reason for suffix is dev docs seem to indicate C4DPYTHONPATH is for python 2.7 only, I wouldn't be surprised if 3.12 worked too, especially pure python modules: https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#custom-interpreter-bound-search-paths:~:text=%3C%3D%20C4D_VERSION)-,C4DPYTHONPATH311,-Note%20To%20register

deadline_cloud_extension/DeadlineCloud.pyp
gp-dan: Have you installed the modules to a supported interpreter path? https://developers.maxon.net/docs/py/2024_0_0a/manuals/manual_py_libraries.html#python-interpreter-bound-search-paths

DEVELOPMENT.md
gp-dan: when I pip install deadline-cloud-for-cinema-4d to a interpreter search path supported by Cinema 4D it installs the pywin32 dependency too, both submitter and adaptor load it

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
gp-dan: My experience is the term "multi-pass" is unambiguously an output path in Cinema4D. The c4d api uses the handle `RDATA_MULTIPASS_FILENAME`. We can certainly change this line but it would also mean changing the respective action method, the adaptor schema, the init data and the template --  and probably the render settings data class and the ui widget for consistency sake too.

karthikbekalp: Ok thanks for the clarifications!

DEVELOPMENT.md
karthikbekalp: Hmm, I've had issues with this error when I don't install pywin32 separately. But I'll investigate this on my side. Thanks for the tip. 

```
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)Traceback (most recent call last):
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)  File "C:\ProgramData\Amazon\OpenJD\session-1e008b980032473da35ef8f59bb9942e09l386vr\.env\Lib\site-packages\deadline\cinema4d_adaptor\Cinema4DClient\plugin\DeadlineCloudClient.pyp", line 25, in <module>
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    from deadline.cinema4d_adaptor.Cinema4DClient.cinema4d_client import main
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)  File "C:\ProgramData\Amazon\OpenJD\session-1e008b980032473da35ef8f59bb9942e09l386vr\.env\Lib\site-packages\deadline\cinema4d_adaptor\Cinema4DClient\__init__.py", line 3, in <module>
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    from .cinema4d_client import Cinema4DClient
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)  File "C:\ProgramData\Amazon\OpenJD\session-1e008b980032473da35ef8f59bb9942e09l386vr\.env\Lib\site-packages\deadline\cinema4d_adaptor\Cinema4DClient\cinema4d_client.py", line 9, in <module>
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    from openjd.adaptor_runtime_client import (
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)  File "C:\ProgramData\Amazon\OpenJD\session-1e008b980032473da35ef8f59bb9942e09l386vr\.env\Lib\site-packages\openjd\adaptor_runtime_client\__init__.py", line 19, in <module>
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    from .win_client_interface import WinClientInterface as ClientInterface  # type: ignore
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)  File "C:\ProgramData\Amazon\OpenJD\session-1e008b980032473da35ef8f59bb9942e09l386vr\.env\Lib\site-packages\openjd\adaptor_runtime_client\win_client_interface.py", line 12, in <module>
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    from .named_pipe.named_pipe_helper import NamedPipeHelper
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)  File "C:\ProgramData\Amazon\OpenJD\session-1e008b980032473da35ef8f59bb9942e09l386vr\.env\Lib\site-packages\openjd\adaptor_runtime_client\named_pipe\named_pipe_helper.py", line 7, in <module>
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)    import win32file
2024/10/25 07:48:35+00:00 ADAPTOR_OUTPUT: STDOUT: (null)ModuleNotFoundError: No module named 'win32file'
```

deadline_cloud_extension/DeadlineCloud.pyp
karthikbekalp: Hmm, using the `C4DPYTHONPATH311` fixed my issues.

DEVELOPMENT.md
karthikbekalp: Thanks for your clarification!

karthikbekalp (dismissed):
Thanks for your contributions! The changes look good to me.

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -170,6 +174,40 @@ def get_render_data(doc=None, take=None):
             render_data = doc.GetActiveRenderData()
         return render_data
 
+    @staticmethod
+    def get_output_paths(take=None) -> str:
```
epmog: Are we duplicating all of `get_output_directories` here? Looks like the main difference is that we want the specific values rather than having the two of them in a set? Seems like we could completely remove the old usage, and just use the results from this function?

epmog (changes requested):
Aside from the one comment to discuss and perhaps de-dupe, this will have to be a breaking change (signified by `feat!`) due to changing how the installation works and changing the adaptor interface (a new submitter requires a new adaptor in this case)

Otherwise, LGTM

src/deadline/cinema4d_submitter/scene.py
gp-dan: hi @epmog , get_output_directories() is for fully expanded directories and called to build the asset references structure --  in contrast get_output_paths() is only for getting the save image and save multi-pass "path" values (not directory values) which can be subtle and often don't have a fully expanded name, for example things line frame and extension are finalized by c4d during render time. They could be combined to reduce lines of code but you might end up with more complexity in one function that must navigate corner cases of 2 different use cases. It's a subjective call on net cost of code maintenance - my instinct is to wait until later to factor out any common code as combining them now might have unintended side effects that need more testing.

epmog: Hmm... If I do a diff between the methods, that difference in functionality doesn't exist? It looks like one is missing pulling the default take data, we may want to add that? `get_output_paths` appears to otherwise return the exact same data as `get_output_directories`, just formatted differently (ie. dirname for asset references and in a set, but the underlying value being access is the same). Which if we're adding them as PATH + OUTPUT job parameters to the template will automatically handle that. 

Let me know if I'm missing something there.

If there is an actual difference between the two, then I ask that we document what that is as a doc-string, otherwise I'd err on having the one function and if we're still manually adding the asset reference for the outputs, then we can take the dirname of the result of one function. 

```diff
-    def get_output_directories(render_data=None, take=None) -> set[str]:
-        """
-        Returns a list of directories files will be output to.
-        """
+    def get_output_paths(take=None) -> str:
         doc = c4d.documents.GetActiveDocument()
         doc_path = doc.GetDocumentPath()
-        if not take:
-            take_data = doc.GetTakeData()
-            take = take_data.GetCurrentTake()
+
+
+        
         render_data = Scene.get_render_data(doc=doc, take=take)
-        rbc = render_data.GetDataInstance()
+        render_base_container_instance = render_data.GetDataInstance()
         rpd = {
             "_doc": doc,
             "_rData": render_data,
-            "_rBc": rbc,
+            "_rBc": render_base_container_instance,
             "_frame": doc.GetTime().GetFrame(doc.GetFps()),
         }
         if take:
             rpd["_take"] = take
-        image_paths = set()
+        default_out = ""
+        multi_out = ""
         if render_data[c4d.RDATA_SAVEIMAGE]:
             path = render_data[c4d.RDATA_PATH]
             xpath = c4d.modules.tokensystem.FilenameConvertTokens(path, rpd)
@@ -25,7 +23,7 @@
                 if xpath.startswith("./"):
                     xpath = xpath[2:]
                 xpath = os.path.join(doc_path, xpath)
-            image_paths.add(os.path.dirname(os.path.normpath(xpath)))
+            default_out = os.path.normpath(xpath)
         if render_data[c4d.RDATA_MULTIPASS_SAVEIMAGE]:
             path = render_data[c4d.RDATA_MULTIPASS_FILENAME]
             xpath = c4d.modules.tokensystem.FilenameConvertTokens(path, rpd)
@@ -33,5 +31,5 @@
                 if xpath.startswith("./"):
                     xpath = xpath[2:]
                 xpath = os.path.join(doc_path, xpath)
-            image_paths.add(os.path.dirname(os.path.normpath(xpath)))
-        return image_paths
\ No newline at end of file
+            multi_out = os.path.normpath(xpath)
+        return default_out, multi_out
\ No newline at end of file

```

Interestingly... we're previously completely overriding the render_data without using the parameter in get_output_directories... but that's just a musing.

gp-dan: howdy @epmog it's common in other submitters to have 2 methods for this - for example in maya [get_output_directories()](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/scene.py#L119) and [output_path()](https://github.com/aws-deadline/deadline-cloud-for-maya/blob/mainline/src/deadline/maya_submitter/scene.py#L136) are the equivalent of our get_output_directories() and get_output_paths() respectively -- the only difference is we return a tuple default_out, mulit_out here. It's a feature that get_output_paths() doesn't fall back to main take, if it did the output mapping of each take would be overridden with the main take path - this is what I mean about 2 use cases needing 2 different functions, combining them can open the door to side effects that break other features.  I'll add a doc string to get_output_paths() to make this clearer.

karthikbekalp (approved):
Thanks for your contribution. Approving the changes.

epmog (approved):

Issue #66, opened 2024-10-04
Bug: Adapter does not fail a job with "Out of memory" errors


karthikbekalp:
### Expected Behaviour

If rendering fails, we expect that the task is in FAILED status. 

### Current Behaviour

If rendering using Redshift fails with "Rendering failed: Out of memory", the job is classified as SUCCEEDED. 

### Reproduction Steps

I submitted a Redshift job to a Windows CMF instance with low GPU memory.  
[BYOLFarm_BYOL-Windows-Queue_session-c1f3baa0741f445dbae0dda6c85befa6.log](https://github.com/user-attachments/files/17263485/BYOLFarm_BYOL-Windows-Queue_session-c1f3baa0741f445dbae0dda6c85befa6.log)

### Code Snippet

I assume that this should be similar to effort for https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/51. I wonder if there is a way to generalize this so that all types of Cinema 4D failures result in tasks getting failed?

yuanmich2:
Just ran into this issue today. For me, the job never failed or completed, it just froze. on this error:
```
2024/10/07 15:16:21-07:00 ADAPTOR_OUTPUT: STDOUT: Rendering frame 0 at <Mon Oct  7 22:16:21 2024>
2024/10/07 15:16:30-07:00 ADAPTOR_OUTPUT: STDOUT: Error: RenderDocument: 1
```

The error gets printed on these few lines:
https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/01a7c3f935d3a47a50745dff9f7e726dc1d69c61/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py#L71-L74

Digging a bit deeper, I found that error code `1` comes from `c4d.RENDERRESULT_OUTOFMEMORY` which is what took me to this github issue.

There's actually a large number of possible errors that we are not checking for at that spot in the code. A more complete list is here: https://developers.maxon.net/docs/py/2024_0_0a/consts/RENDERRESULT.html

yuanmich2:
This is kind of off topic, but, in my case, the "Out of memory" error was actually caused by a licensing issue. I fixed it for my c7i.2xlarge Windows CMF machine by just setting the `redshift_LICENSE` environment variable to point to `port@host` for the port and hostname of my license server.

Either way, we need better error handling, but I just want to call out that this specific error could actually be a licensing thing.

[...]

PR #76, opened 2024-10-22
fix: adaptor handle render result failure results (#66)

gp-dan:
Check RenderDocument() result in Cinema4DHandler.start_render() against all c4d.RENDERRESULT values to ensure job status is correct.

#66 renders that fail report as SUCCEEDED in DCM  

Check RenderDocument() result against all RENDERRESULT values so errors bubble up to DCM

Errors properly caught and logged, status properly set

Windows CMF Cinema4D 2025

No

No

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -10,6 +10,19 @@ try:
 except ImportError:  # pragma: no cover
     raise OSError("Could not find the Cinema4D module. Are you running this inside of Cinema4D?")
 
+_RENDERRESULT = {
+    c4d.RENDERRESULT_OK: "Function was successful.",
+    c4d.RENDERRESULT_OUTOFMEMORY: "Not enough memory.",
+    c4d.RENDERRESULT_ASSETMISSING: "Assets (textures etc.) are missing.",
+    c4d.RENDERRESULT_SAVINGFAILED: "Failed to save.",
+    c4d.RENDERRESULT_USERBREAK: "User stopped the processing.",
+    c4d.RENDERRESULT_GICACHEMISSING: "GI cache is missing.",
+    c4d.RENDERRESULT_NOMACHINE: "Machine was not found. (Team Rendering only)",
+    c4d.RENDERRESULT_PROJECTNOTFOUND: "Project was not found.",
+    c4d.RENDERRESULT_ERRORLOADINGPROJECT: "There was an error while loading the project.",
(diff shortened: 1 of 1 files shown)


evanspearman-a (dismissed):

epmog (dismissed):

karthikbekalp (changes requested):
Hmm, I don't think this fixes the problem. The Task is stuck in Running status and the task is not classified as FAILED. 

[
Screenshot 2024-10-22 at 4 18 58 PM
](url)

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -68,8 +81,11 @@ def start_render(self, data: dict) -> None:
             c4d.RENDERFLAGS_EXTERNAL | c4d.RENDERFLAGS_SHOWERRORS,
             prog=progress_callback,
         )
-        if result != c4d.RENDERRESULT_OK and result != c4d.RENDERRESULT_USERBREAK:
-            print("Error: RenderDocument: %s" % result)
+        result_description = _RENDERRESULT.get(result, None)
+        if result_description is None:
```
karthikbekalp, line 85: Can we have a test for these?

gp-dan: We could but there's zero testing skeleton for handler at the moment. It seems like a larger project than narrowly  testing the correct handling of RenderDocument() result value. Do you see a simple way to test this code path without dragging that extra complexity and line count for this PR? Thanks!

karthikbekalp: Hmm, I was thinking of something similar to how it was done in [one of the PRs](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/72/files). 

Mocking this whole function may not be straightforward, but if we were to abstract this into a separate function then mocking that function could be simpler. 

For example, 

```
def _process_render_result(result, None):
     if result_description is None:
            raise RuntimeError("Error: unhandled render result: %s" % result)
        if result != c4d.RENDERRESULT_OK:
            raise RuntimeError("Error: render result: %s" % result_description)
        else:
            print("Finished Rendering")
```

By extracting this logic, we could more easily mock and test _process_render_result. This approach would also allow us to test different RenderResults such as AssetMissing, SavingFailed, etc.

But even after using this approach, there will still be huge gaps in testing as the handler code does not have any testing right now. So I'm fine if we cannot add tests now. :)

karthikbekalp (approved):
Thanks again for your contributions, I've also tested this locally and I can confirm that this works! 

Screenshot 2024-10-23 at 9 28 42 AM

epmog (approved):

PR #77, opened 2024-10-25
test: add plain text job_bundle_output_tests

gp-dan:
5 test scenes implemented as python scripts and respective job bundles:
- simple physical render
- simple redshift render
- redshift with multipass render
- redhsift with takes
- redshift with texture asset, bmp file generated by python script

Example scenes and respective expected job bundle outputs

Plain text py files to generate scenes and their respective job bundle outputs with normalized paths

Developers have a common set of test scenes

- Launch Cinema 4D
- Extensions > User Scripts > Run Script ...
- Select script file ./job_bundle_output_tests/{test}/scene/cube.py
- The script builds the scene and saves it to ./job_bundle_output_tests/{test}/scene/{test}.c4d
- Extensions > Deadline Cloud Submitter
- Submit or Export Bundle 

No

No

----

### DIFF
--- a/job_bundle_output_tests/redshift_textured/scene/cube.py
+++ b/job_bundle_output_tests/redshift_textured/scene/cube.py
@@ -1,0 +1,74 @@
+import os
+import struct
+import c4d
+
+
+def _checkerboard_bmp(filename):
+    width = 128
+    height = 128
+    color1 = (255, 255, 255)
+    color2 = (0, 0, 0)
+    # BMP file header
+    file_size = 14 + 40 + (width * height * 3)
+    file_header = struct.pack("<2sIHHI", b"BM", file_size, 0, 0, 54)
+    # DIB header
+    dib_header = struct.pack("<IiiHHIIiiII", 40, width, height, 1, 24, 0,
+                             width * height * 3, 2835, 2835, 0, 0)
+    pixel_data = []
+    for y in range(height):
+        row_data = []
+        for x in range(width):
+            if (x // 8 + y // 8) % 2 == 0:
+                row_data.extend(color1)
+            else:
+                row_data.extend(color2)
+        padding = (4 - (width * 3) % 4) % 4
+        row_data.extend([0] * padding)
+        pixel_data.extend(row_data)
+    with open(filename, "wb") as bmp_file:
+        bmp_file.write(file_header)
+        bmp_file.write(dib_header)
+        bmp_file.write(bytearray(pixel_data))
+
+
+def main():
+    doc = c4d.documents.GetActiveDocument()
+    doc.Flush()
+    cube = c4d.BaseObject(c4d.Ocube)
+    cube[c4d.PRIM_CUBE_LEN] = c4d.Vector(200, 200, 200)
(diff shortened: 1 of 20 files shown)


job_bundle_output_tests/redshift_textured/scene/cube.py
```diff
@@ -0,0 +1,74 @@
...
+    doc = c4d.documents.GetActiveDocument()
+    doc.Flush()
+    cube = c4d.BaseObject(c4d.Ocube)
+    cube[c4d.PRIM_CUBE_LEN] = c4d.Vector(200, 200, 200)
+    cube.SetAbsPos(c4d.Vector(0, 170, -170))
+    doc.InsertObject(cube)
+    mat = c4d.BaseList2D(c4d.Mmaterial)
+    doc.InsertMaterial(mat)
+    mat[c4d.MATERIAL_USE_REFLECTION] = False
+    bitmapShader = c4d.BaseShader(c4d.Xbitmap)
```
karthikbekalp: Please update this variable name from `bitmapShader` to `bitmap_shader` (and replace all the occurrences) to pass the SonarCloud checks.

karthikbekalp (changes requested):
Thanks for your contribution. I've been able to run the scripts and confirm that the renders work. 
Although the SonarCloud tests are failing, this can be fixed with a simple variable name change.

job_bundle_output_tests/redshift_textured/scene/cube.py
karthikbekalp: Thanks for fixing it!

karthikbekalp:
Hmm, sorry my bad. Looks like the SonarCloud checks still failed because it detected duplicate code amongst all the tests. Can you move the duplicate code into a separate util function in the test package so that we can eliminate this issue?

karthikbekalp:
But the linting has failed so there's a need to reformat the code anyway. https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/actions/runs/11564466263/job/32191091030?pr=77 
Can you please run 

```
# For fixing the formatting
hatch run fmt

# Add the formatted files
git add ...

# Amend your commit
git commit --amend

# Force push the amended commit
git push --force-with-lease <remote_name> gp-dan:job_bundle_output_tests
```

gp-dan:
Thanks @karthikbekalp I intentionally kept the duplication as I was trying to make a clean plain text equivalent of a static .c4d scene description.  if we factor out the duplicated lines the scripts will no longer be static scene descriptions but programs that require a util library to run. You'll need to add the util library to the environment before running Extensions > User Scripts > Run Script ... The scenes will no longer be strictly reproducible and self contained tests as the scene descriptions now depend on a utility function that may change and affect all scenes and complicate finding regressions. Could we configure the CI to ignore the duplication requirement on job_bundle_output_tests dir to avoid adding the extra complexity? What do you think?

karthikbekalp:
Thanks for your clarification. It looks like we can override and merge even if the SonarCloud checks failed.

karthikbekalp (approved):

lucaseck (approved):

PR #80, opened 2024-10-31
fix: Ensure only Critical stops fail the job and not others.

karthikbekalp:
Apparently, sometimes CRITICAL regex can be printed out even though the job itself is successful. 
Thanks to @Ahuge for pointing it out to us. 
For example, here's the logs for one such job. 
```
2024/10/15 16:03:39-07:00 ==============================================
2024/10/15 16:03:39-07:00 --------- Running Task
2024/10/15 16:03:39-07:00 ==============================================
2024/10/15 16:03:39-07:00 Parameter values:
2024/10/15 16:03:39-07:00 Frame(INT) = 2
2024/10/15 16:03:39-07:00 ----------------------------------------------
2024/10/15 16:03:39-07:00 Phase: Setup
2024/10/15 16:03:39-07:00 ----------------------------------------------
...

2024/10/15 16:03:40-07:00 ADAPTOR_OUTPUT: STDOUT: CRITICAL: nullptr [text_object.cpp(1082)] [objectbase1.hxx(549)]
2024/10/15 16:03:40-07:00 ADAPTOR_OUTPUT: STDOUT: CRITICAL: nullptr [text_object.cpp(1082)] [objectbase1.hxx(549)]
....
2024/10/15 16:08:42-07:00 ADAPTOR_OUTPUT: STDOUT: Progress: 100%
2024/10/15 16:08:45-07:00 ADAPTOR_OUTPUT: STDOUT: Finished Rendering
```

I modified the regex to only fail the jobs with `CRITICAL Stop` in output. 
I couldn't find any information regarding which critical errors should fail the job for Cinema 4D. Hence, I'm only checking for stops. 
If we find out that its not true and somehow `CRITICAL: nullptr` is special, we can change the error regex accordingly in the future. 

This avoids the problem of failing successful jobs. 

Added few test cases and ran `hatch run all:test` and ensured that the tests all passed. 

No

No

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -193,7 +193,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
             error_regexes = [
-                re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)
+                re.compile(r"(.*Error: .*)|(.*\[Error\].*)|(.*CRITICAL: Stop.*)", re.IGNORECASE)
             ]
 
             callback_list.append(RegexCallback(completed_regexes, self._handle_complete))
--- a/test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
+++ b/test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
@@ -1,7 +1,6 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 from __future__ import annotations
 
-import re
 import pytest
 from deadline.cinema4d_adaptor.Cinema4DAdaptor import Cinema4DAdaptor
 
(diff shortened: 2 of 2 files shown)


test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
```diff
@@ -24,36 +23,36 @@ def init_data() -> dict:
...
-        assert str(adaptor._exc_info) == f"Cinema4D Encountered an Error: {stdout}"
-
-    def test_handle_error(self, init_data: dict) -> None:
+    @pytest.mark.parametrize(
+        "stdout,error_expected",
+        [
+            # Only critical stop errors should fail the job
+            ("CRITICAL: Stop [ge_file.cpp(1172)]", True),
+            # Any errors should fail the job
+            ("Redshift Error: Maxon licensing error: User not logged in (7)", True),
```
joel-wong-aws, line 32: nit: should we also add an example for the `.*\\[Error\\].*` regex?

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -192,7 +192,7 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
             completed_regexes = [re.compile(".*Finished Rendering.*")]
             progress_regexes = [re.compile(".*Progress ([0-9]+)%.*")]
             error_regexes = [
-                re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: .*", re.IGNORECASE)
+                re.compile(".*Error: .*|.*\\[Error\\].*|.*CRITICAL: Stop.*", re.IGNORECASE)
```
joel-wong-aws: Can we use a negative lookahead for `CRITICAL: nullptr` messages instead of just looking for `CRITICAL: Stop` messages?

Also, as a general note, should we use a raw string (often used with regexes) and put parenthesis around each of the three cases to make the regexes easier to understand?

joel-wong-aws (changes requested):

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
karthikbekalp: Yeah, I touched upon this in the description : https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/80#issue-2627958789  and I wasn't sure if we should check for negative look ahead. 

I couldn't find any information regarding which critical errors should fail the job for Cinema 4D. Hence, I'm only checking for stops.
If we find out that its not true and somehow CRITICAL: nullptr is special, we can change the error regex accordingly in the future.

test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
karthikbekalp: Sure, I can add it in a follow up PR.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
karthikbekalp: > Also, as a general note, should we use a raw string (often used with regexes) and put parenthesis around each of the three cases to make the regexes easier to understand?

How about I add this as a follow up? This is one of the fixes that we need to get in as quickly as possible to ensure that we don't have customers that have successful jobs that fail because of this.

joel-wong-aws: Discussed with Karthik. Cinema4D uses a different error convention than other software. Approving as this is the most reasonable regex we can come up with right now and can always update it if we find additional cases for which to fail the render.

joel-wong-aws (dismissed):

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
karthikbekalp: > should we use a raw string (often used with regexes) and put parenthesis around each of the three cases to make the regexes easier to understand?

Looks like this PR was not reviewed by others yet so I might as well update it. So I updated according to your suggestions. :)

joel-wong-aws (approved):

baxeaz (approved):

PR #85, opened 2024-11-08
feat: Add InstallBuilder submitter installer XML.

karthikbekalp:

For installing Cinema 4D submitter, customers needed to follow the manual instructions on Github. 

This adds the XML file for providing installation for submitter through DeadlineCloudSubmitter installer. 

Easier for customers to use Cinema 4D when they can use installer instead of manually setting the variables and copying files. 

- Created a fresh instance. 
- Installed Cinema 4D 2025 on it. 
- Installed the submitter through a locally built installer. 
- Confirmed that the plugins directory was created and `C4DPYTHONPATH311` environment variable was set. 

Screenshot 2024-11-08 at 10 49 38 AM
Screenshot 2024-11-08 at 10 49 51 AM
Screenshot 2024-11-08 at 10 50 25 AM
Screenshot 2024-11-08 at 10 54 44 AM
Screenshot 2024-11-08 at 10 57 13 AM
Screenshot 2024-11-08 at 10 57 02 AM
Screenshot 2024-11-08 at 10 57 23 AM

If no Pyside GUI components are installed, we would prompt customers to install it. 

Screenshot 2024-11-13 at 12 02 38 AM

If customers select yes, the GUI components that are required for `deadline[gui]` is installed. 

If customers select No. We show a friendly message. 

Screenshot 2024-11-13 at 12 02 50 AM

No

No
----


install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -0,0 +1,132 @@
...
+             <name>c4ddeps</name>
+             <platforms>all</platforms>
+             <distributionFileList>
+                 <distributionDirectory allowWildcards="1">
+                     <origin>components/deadline-cloud-for-cinema-4d/dependency_bundle</origin>
+                 </distributionDirectory>
+             </distributionFileList>
+        </folder>
+        <folder>
+            <description>Program Files</description>
```
joel-wong-aws: Is there a different name we can use for this? When I read `Program Files` , I immediately thought of Window's Program Files, which I don't think it what you're referring to

install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -0,0 +1,132 @@
...
+        </stringParameter>
+        <directoryParameter>
+            <name>cinema_4d_installdir</name>
+            <title>Cinema 4D</title>
+            <description>Cinema 4D Installation Directory</description>
+            <explanation>Path to installation directory in the Cinema4D resources directory. For easiest installation, Cinema 4D should be installed before installing Deadline Cloud for Cinema 4D.</explanation>
+            <allowEmptyValue>0</allowEmptyValue>
+            <ask>yes</ask>
+            <cliOptionName>cinema4d-plugins-directory</cliOptionName>
+            <cliOptionText>Path to plugins directory in the Cinema4D resources directory. </cliOptionText>
```
joel-wong-aws: The `name`, `description`, `explanation`, and `cliOptiion*` fields all seem to describe different directory - the main C4D directory, the resources directory, and the `plugins` directory. Can we update the text to all point to the same place? I would propose pointing them all at the main install directory to be more clear to the user during installation

install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -0,0 +1,132 @@
...
+        </showInfo>
+		<fnAddPathEnvironmentVariable>
+			<progressText>Setting C4DPYTHONPATH311</progressText>
+			<name>C4DPYTHONPATH311</name>
+			<value>${cinema_4d_submitter_installdir}</value>
+			<scope>${installscope}</scope>
+			<insertAt>end</insertAt>
+		</fnAddPathEnvironmentVariable>
+        <showInfo>
+            <text>Set C4DPYTHONPATH311 to ${cinema_4d_submitter_installdir}</text>
```
joel-wong-aws: Would `Add C4DPYTHONPATH311 to ${cinema_4d_submitter_installdir}` be more accurate?

joel-wong-aws (changes requested):

install_builder/deadline-cloud-for-cinema-4d.xml
karthikbekalp: Makes sense, I was using modules from Nuke for this pattern. https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/mainline/install_builder/deadline-cloud-for-nuke.xml#L10-L23 

Maybe a better name would be `Cinema4D submitter installation directory`?

install_builder/deadline-cloud-for-cinema-4d.xml
karthikbekalp: Sure or maybe something like `Adding ${cinema_4d_submitter_installdir} to C4DPYTHONPATH311`?

install_builder/deadline-cloud-for-cinema-4d.xml
karthikbekalp: Discussed offline, updated.

joel-wong-aws (dismissed):

crowecawcaw:
Do the `ensurepip` commands need to be manually run? If so, I don't think we should add this to the installer until we figure out a better fix. The main benefit for adding the submitter to the installer is to avoid command line steps like this.

karthikbekalp:
> Do the `ensurepip` commands need to be manually run? If so, I don't think we should add this to the installer until we figure out a better fix. The main benefit for adding the submitter to the installer is to avoid command line steps like this.

So as I understand when we install the submitter using the DeadlineCloudSubmitter we also bundle its dependencies such that it should work out of the box once the installation is complete. I believe it uses https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/depsBundle.py script to bundle those dependencies.

install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -0,0 +1,132 @@
+<component>
+    <name>deadline_cloud_for_cinema_4d</name>
+    <description>Deadline Cloud for Cinema 4D 2024-2025</description>
+    <detailedDescription>Cinema 4D plugin for submitting jobs to AWS Deadline Cloud. Compatible with Cinema 4D 2024/2025</detailedDescription>
+    <canBeEdited>1</canBeEdited>
+    <selected>0</selected>
+    <show>1</show>
+    <folderList>
+        <folder>
+            <description>Cinema4D Dealine Cloud Extension</description>
```
lucaseck: nit: `Dealine` -> `Deadline`

install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -0,0 +1,132 @@
...
+    </parameterList>
+    <postInstallationActionList>
+        <createDirectory>
+            <path>${cinema_4d_installdir}/plugins</path>
+            <ruleList>
+                <fileTest condition="not_exists" path="${cinema_4d_installdir}/plugins"/>
+            </ruleList>
+        </createDirectory>
+        <copyFile>
+            <destination>${cinema_4d_installdir}/plugins/DeadlineCloud.pyp</destination>
```
lucaseck, line 101: Does this get removed on uninstall? In the past I've seen files moved by `copyFile` not get removed by default on uninstall.

install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -0,0 +1,132 @@
...
+            <origin>${installdir}/tmp/c4d_deps/DeadlineCloud.pyp</origin>
+        </copyFile>
+        <showInfo>
+            <text>Moved DeadlineCloud.pyp to ${cinema_4d_installdir}/plugins/</text>
```
lucaseck, line 106: Was this tested with a command line installation to see how it affects the flow there? I don't think it will block if running in unattended mode, but worth checking if you haven't already. Personally I'm not sure that this is needed as to me it just interrupts the installation flow with a popup that must be acknowledged I think?

karthikbekalp: Yup, I tried just now with the CLI and it worked. 

```
> DeadlineCloudSubmitter-windows-x64-installer.exe --mode unattended --enable-components deadline_cloud_for_cinema_4d  --cinema4d-plugins-directory "C:\Program Files\Maxon Cinema 4D 2025"
```

And the logs suggests that everything was installed as expected. 

```
...
Moved DeadlineCloud.pyp to C:\Program Files\Maxon Cinema 4D 2025/plugins/
Setting C4DPYTHONPATH311
Added C4DPYTHONPATH311 environment variable with value D:\Cinema4D in scope user
Broadcasting environment changes
...
Creating Uninstaller
Creating uninstaller 25%
Creating uninstaller 50%
Creating uninstaller 75%
Creating uninstaller 100%
Uninstaller icon changed
Installation completed
Log finished 11/12/2024 at 15:04:51
Exiting with code 0

```

install_builder/deadline-cloud-for-cinema-4d.xml
karthikbekalp: Updated.

install_builder/deadline-cloud-for-cinema-4d.xml
karthikbekalp: It does not, thanks I've updated it now, I'll test now.

joel-wong-aws (dismissed):

lucaseck (approved):

crowecawcaw (approved):

PR #91, opened 2024-11-13
fix: install pip if required before installing GUI dependencies

joel-wong-aws:
When installing the Deadline Cloud Cinema4D submitter on my workstation, I could not see the Deadline Cloud plugin in the C4D extensions menu. 

Upon investigation, I found errors in the C4D Python console with module not found errors. The message was actually referring to `pip` being missing. To remedy this, added `python -m ensurepip` before installing the GUI dependencies.

Also fixed a few typos, re-organized imports, and updated plugin display name.

Also fixed a typo and re-organized imports

Smoother installation for most users.

Opened Cinema4D and confirmed that the the Deadline Cloud extension was installed.

Yep!

No, not required. While this makes the installation easier, any of the existing troubleshooting docs should be kept in case of errors.

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/__init__.py
+++ b/src/deadline/cinema4d_submitter/__init__.py
@@ -1,10 +1,11 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+import logging as _logging
 import sys
 import subprocess
 from pathlib import Path
+
 import c4d
 
-import logging as _logging
 
 _logger = _logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


karthikbekalp (dismissed):
Thanks for fixing this issue.

src/deadline/cinema4d_submitter/__init__.py
```diff
@@ -23,42 +24,52 @@ def has_gui_deps():
...
         "-m",
         "pip",
         "install",
         f"deadline[gui]=={deadline.client.version}",
     ]
+
     # module_directory assumes relative install location of:
     #   * [installdir]/Submitters/Cinema4D/deadline/cinema4d_submitter/cinema4d_render_submitter.py
     module_directory = Path(__file__).parent.parent.parent
+    print(module_directory)
```
crowecawcaw: Is this from debugging?

crowecawcaw (approved):

src/deadline/cinema4d_submitter/__init__.py
joel-wong-aws: Whoops! Nice spotting, removed

karthikbekalp (approved):

Issue #95, opened 2024-11-19
Feature request: sticky settings for all fields


joel-wong-aws:
### Describe the problem

The following fields do not currently support sticky settings (settings that are saved and reused between submissions):
* priority
* initial state
* maximum failed tasks count
* maximum retries per task
* Queue environment fields, such as Conda packages and Conda channels
* Output path
* Multi-pass path
* takes
* require all input paths exist
* host requirements

### Proposed Solution

The AWS Deadline Cloud KeyShot integration had a [custom implementation for sticky settings](https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/93), which may be applicable to the Cinema 4D integration. Alternatively, we could look into adding sticky settings for this to the underlying [Deadline Cloud bundle submission flow](https://github.com/aws-deadline/deadline-cloud)

### Example Use Cases

I modified all of the above fields when submitting a Cinema 4D job. The settings were not saved between submissions.

karthikbekalp:
Tried to start working on a fix for this. Unfortunately, even though the settings can be stored, it gets overwritten at the [deadline cloud](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/ui/widgets/shared_job_settings_tab.py#L297-L300) dialog side. 
I'll get back to this once we have fixed it on the deadline cloud side.

karthikbekalp:
Closing this as we have implemented sticky settings for most of the fields.

PR #96, opened 2024-11-19
feat!: install Conda packages by default

joel-wong-aws:
By default, the list of Conda packages to use for a C4D submission is empty. This means that users will have to re-enter the Conda packages every time they want to submit a job if they are using Conda.

We believe most users will want to use `cinema4d=<major version>.* cinema4d-openjd=0.*` Conda packages by default, so we've added those. 

DefaultCondaPackages

Customers who wish to use the a different configuration can modify their `deadline/cinema4d_submitter/cinema4d_render_submitter.py` file manually.

Sticky settings are out of scope for this change. I've added a GitHub issue to track that [here](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/95)

Users submitting jobs from Cinema 4D should be able to onboard easier and faster!

Opened the submitter and confirmed that `cinema4d=2025.* cinema4d-openjd=0.*` were set as the default Conda packages. Submitted a Cinema 4D job and confirmed that it rendered successfully

Yep!

Yes! All tests output a cube, as expected.

No, not required

Yes, this is a breaking change. Customers that used AWS Deadline Cloud customer managed fleets that had Cinema 4D pre-installed on their workers and do not use Conda packages will need to update their `deadline/cinema4d_submitter/cinema4d_render_submitter.py` file to remove the default Conda packages.

----

BREAKING CHANGE

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -19,6 +19,7 @@ from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # pylint
 from deadline.client.exceptions import DeadlineOperationError
 from qtpy.QtCore import Qt  # type: ignore[attr-defined]
 
+from ._version import version_tuple as adaptor_version_tuple
 from .data_classes import (
     RenderSubmitterUISettings,
 )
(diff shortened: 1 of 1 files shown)


crowecawcaw (dismissed):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -426,7 +426,7 @@ def on_create_job_bundle_callback(
         job_setup_widget_type=SceneSettingsWidget,
         initial_job_settings=render_settings,
         initial_shared_parameter_values={
-            "CondaPackages": "",
+            "CondaPackages": f"cinema4d={str(c4d.GetC4DVersion())[:4]}.* cinema4d-openjd=0.*",
```
lucaseck: Is there a reason for the adaptor being 0.* vs 0.MINOR.* like most of the other integrations do? e.g. [houdini](https://github.com/aws-deadline/deadline-cloud-for-houdini/blob/244bc7130999f38e5d408b0e845939411b09d2a4/src/deadline/houdini_submitter/python/deadline_cloud_for_houdini/queue_parameters.py#L150)

lucaseck (dismissed):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
mwiebe: I think we should make it `0.MINOR.*` so we can do semantic versioning of the adaptor.

mwiebe (changes requested):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
joel-wong-aws: Updated to match the format of other packages and follow semantic versioning!

lucaseck (approved):

mwiebe (approved):

PR #108, opened 2024-11-21
chore: add length contraints to parameters

crowecawcaw:
There are no length constraints on the job parameters. Someone could put excessive amounts of text in the parameter fields which might cause issues with the submitter.

Add length constraints on the inputs. The constraints are high (100k characters) so reasonable inputs will not be affected. 

Better handling of edge cases and devious user inputs.

Not tested. I don't have C4D setup.

Does not need documentation.

No

----


crowecawcaw:
Correct place to add these is in Qt. Better PR: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/108

PR #110, opened 2024-11-21
fix: install C4D plugin into user directory to avoid needing elevated permissions

joel-wong-aws:
When attempting to do a user install, the Cinema 4D install directory was being set to a system directory, which resulted in the user receiving a permissions error.

Instead of attempting to do a user install to a system directory, install to the user directory and then use the `g_additionalModulePath` to register the submitter with Cinema 4D.

Smoother installation experience and one less click to install :)

Rebuilt the installer and used it to install Cinema 4D. Ran the job bundle output tests, all succeeded.

* Have you run the unit tests?
No, N/A

Yes

No, N/A

No

----

### DIFF
--- a/install_builder/deadline-cloud-for-cinema-4d.xml
+++ b/install_builder/deadline-cloud-for-cinema-4d.xml
@@ -73,42 +73,28 @@
     <parameterList>
         <stringParameter name="deadline_cloud_for_cinema_4d_summary" ask="0" cliOptionShow="0">
             <value>Deadline Cloud for Cinema 4D
-- Register the extension with Cinema 4D by moving the DeadlineCloud.pyp file to the Cinema 4D plugins folder. 
-- Set the C4DPYTHONPATH311 to retrieve deadline modules. 
+- Compatible with Cinema 4D 2024/2025
+- Installs the integrated Cinema 4D submitter to the installation directory
+- Registers the extension with Cinema 4D by setting the g_additionalModulePath
+- Sets the C4DPYTHONPATH311 to retrieve deadline modules
             </value>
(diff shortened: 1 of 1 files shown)


install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -73,52 +73,42 @@
...
-                <fileTest condition="not_exists" path="${cinema_4d_installdir}/plugins"/>
+                <fileTest condition="not_exists" path="${cinema_4d_submitter_installdir}/cinema_4d_plugins"/>
             </ruleList>
         </createDirectory>
         <copyFile>
-            <destination>${cinema_4d_installdir}/plugins/DeadlineCloud.pyp</destination>
+            <destination>${cinema_4d_submitter_installdir}/cinema_4d_plugins/DeadlineCloud.pyp</destination>
             <origin>${installdir}/tmp/c4d_deps/DeadlineCloud.pyp</origin>
         </copyFile>
-        <addFilesToUninstaller files="${cinema_4d_installdir}/plugins/DeadlineCloud.pyp" />
```
lucaseck, line 104: This is still needed to have the uninstaller remove the file no?

joel-wong-aws: Since the `DeadlineCloud.pyp` file is located inside the existing `cinema_4d_submitter_installdir`, which is already removed by default during uninstall, there is no need for a special command to remove this as part of the uninstaller!

lucaseck (approved):

ttblanchard (approved):

PR #123, opened 2024-12-09
chore: add tests verifying adaptor validates commands

crowecawcaw:
We want to verify the adapter is validating incoming data using the JSON schemas.

Add a unit tests.

More confidence data validation is working as expected.

Unit tests pass successfully.

No

n/a

No

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -392,11 +392,12 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
         performs a busy wait until the render completes.
         """
 
+        self.validators.run_data.validate(run_data)
+
         if not self._cinema4d_is_running:
             raise Cinema4DNotRunningError("Cannot render because Cinema4D is not running.")
 
         run_data["frame"] = int(run_data["frame"])
-        self.validators.run_data.validate(run_data)
         self._is_rendering = True
 
         for name in _CINEMA4D_RUN_KEYS:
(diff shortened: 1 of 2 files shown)


karthikbekalp (dismissed):
Looks good to me, ship it.

test/deadline_adaptor_for_cinema4d/unit/Cinema4DAdaptor/test_adaptor.py
```diff
@@ -58,3 +59,15 @@ def test_handle_errors_on_error_stdout(
             assert str(adaptor._exc_info) == f"Cinema4D Encountered an Error: {stdout}"
         else:
             assert match is None
+
+
+def test_adapter_rejects_malformed_init_data():
```
joel-wong-aws: nit: `adapter` -> `adaptor`

Same with test below

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -392,11 +392,12 @@ def on_run(self, run_data: dict) -> None:
         performs a busy wait until the render completes.
         """
 
+        self.validators.run_data.validate(run_data)
+
         if not self._cinema4d_is_running:
             raise Cinema4DNotRunningError("Cannot render because Cinema4D is not running.")
 
         run_data["frame"] = int(run_data["frame"])
-        self.validators.run_data.validate(run_data)
```
joel-wong-aws, line 399: Is there a reason we moved this validation up? The line directly above this modifies the run_data, so I would expect it to be validated too?

joel-wong-aws (changes requested):

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
crowecawcaw: The line above accesses the frame key before validating that it exists. It doesn't really matter in practice - it raises a KeyError or TypeError instead of a ValidationException for malformed data - but logically the data should be validated before it's used.

joel-wong-aws (approved):

karthikbekalp (approved):

Issue #118, opened 2024-11-29
Bug: Path mapping between a Windows scene and a Linux worker does not work as expected


Ahuge:
### Describe the bug

The `c4d.RDATA_PATH` (output path) appear as a relative path on Linux when it is actually a Windows path.

For example:

*On Windows*: `P:\ahughes\_DeadlineCloudExamples\simple_c4d_naming_output\image.png`
*On Linux*: `./P:/ahughes/_DeadlineCloudExamples/simple_c4d_naming_output/image.png`

The Linux path appears to be relative and therefore the `map_path` call returns a relative Linux path and the entire folder structure gets recreated relative to the scene file.

This causes additional issues because certain characters cannot be represented on Windows filesystems in paths (For example, the above path that the image.png gets written out to will be as follows `P:\ahughes\_DeadlineCloudExamples\P_\ahughes\_DeadlineCloudExamples\simple_c4d_naming_output\image.png` because Windows turns the `P:` into `P_`)

### Expected Behaviour

I would expect Cinema4D to give us a non-relative path, but I guess we have to have the adaptor work around this and detect that the path is a relative Windows path and treat it as absolute?

### Current Behaviour

A relative path is returned and converted to Linux, so the entire filesystem is recreated from the scene_file
`P:\ahughes\_DeadlineCloudExamples\P_\ahughes\_DeadlineCloudExamples\simple_c4d_naming_output\image.png` 

### Reproduction Steps

Create a scene in Windows Cinema4D and set the output path to be some absolute Windows path (I believe it converts it to absolute regardless) and then render the scene on a Linux Commandline worker.

Here is my modified plugin that you can use to just test the pathing:
```python
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
import os
import sys
import traceback

print('c4d python version: %s' % sys.version)
print('system paths:')
for n in sys.path:
    print(n)

import c4d

def setTake(take_name):
    doc = c4d.documents.GetActiveDocument()
    take_data = doc.GetTakeData()
    if not take_data:
        return

    def get_child_takes(take):
        child_takes = take.GetChildren()
        all_takes = child_takes
        if child_takes:
            for child_take in child_takes:
                all_takes.extend(get_child_takes(child_take))
        return all_takes

    main_take = take_data.GetCurrentTake()
    all_takes = [main_take] + get_child_takes(main_take)

    take = None
    for take in all_takes:
        if take.GetName() == take_name:
            break
[...]

PR #126, opened 2024-12-12
feat: Implement asset path mapping for Cinema 4D scene files

### LINKED ISSUE #125: Bug: Texture assets don't get path mapped
### Describe the bug

When running a Windows scene on a Linux worker, path mapping converts the Scene file and the output paths to Linux paths but do not convert paths to any of the assets

### Expected Behaviour

All paths within the scene should be path mapped.

Especially since the submitter knows about these paths already

### Current Behaviour

Paths are left as original and the Linux render fails

### Reproduction Steps

Create a Windows C4D scene file and make a material that links to a texture shader on disk.

When you submit the scene and render it on the Linux worker you will see errors like this:
```
c4dtoa] 00:00:01  4121MB ERROR   |   [texturesys] /Arnold_skydome_light_color: Could not open file "P:/ALL_NETWORKS/SHOWS/_RENDER_BENCHMARK/C4D/GLASS_GLOBE/GLASS_GLOBE/tex/10Shiodome_Stairs_bw3k.hdr"
2024/12/12 09:54:19-08:00 ADAPTOR_OUTPUT: STDOUT: Could not open file "P:/ALL_NETWORKS/SHOWS/_RENDER_BENCHMARK/C4D/GLASS_GLOBE/GLASS_GLOBE/tex/10Shiodome_Stairs_bw3k.hdr"
```
Obviously with the specific paths you had instead of my paths.

### Environment

At minimum:
1. Operating system: Scene created in Windows Server 2022, Worker running Rocky Linux 9.3
2. Version of Cinema 4D: Scene created in R26 / 26_0.CL390593.49591, worker rendering in 2024.4.1 / 2024_4_1_c6c50ba2c0c8_1301281096
3: Version of this package: 0.5.4

Ahuge:
Fixes: #125 

When using a Linux worker to render a scene created on Windows, assets used in the scene would not get path mapped
Implement path mapping on the asset texture files
Path mapping now happens on the C4D scene and Windows paths are converted to Linux ones
Testing In Progress on a client farm. -- IN PROGRESS

No as it implements functionality that was expected

Yes I believe, because we are changing functionality?

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -65,7 +65,7 @@ class Cinema4DHandler:
         self.doc = c4d.documents.GetActiveDocument()
         self.render_data = self.doc.GetActiveRenderData()
         self.render_data[c4d.RDATA_FRAMESEQUENCE] = c4d.RDATA_FRAMESEQUENCE_MANUAL
-        frame = int(self.render_kwargs["frame"])
+        frame = int(self.render_kwargs.get("frame", data.get("frame")))
         fps = self.doc.GetFps()
         self.render_data[c4d.RDATA_FRAMEFROM] = c4d.BaseTime(frame, fps)
         self.render_data[c4d.RDATA_FRAMETO] = c4d.BaseTime(frame, fps)
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -109,14 +109,14 @@ class Cinema4DHandler:
         if output_path:
             doc = c4d.documents.GetActiveDocument()
             render_data = doc.GetActiveRenderData()
-            render_data[c4d.RDATA_PATH] = output_path
+            render_data[c4d.RDATA_PATH] = self.map_path(output_path)
 
     def multi_pass_path(self, data: dict) -> None:
         multi_pass_path = data.get("multi_pass_path", "")
         if multi_pass_path:
             doc = c4d.documents.GetActiveDocument()
             render_data = doc.GetActiveRenderData()
-            render_data[c4d.RDATA_MULTIPASS_FILENAME] = multi_pass_path
+            render_data[c4d.RDATA_MULTIPASS_FILENAME] = self.map_path(multi_pass_path)
 
     def set_take(self, data: dict) -> None:
         """
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -50,6 +50,17 @@ class Cinema4DHandler:
         self.take = "Main"
         self.map_path = map_path
 
+    def _remap_assets(self) -> None:
+        asset_list: list[Dict[str, Any]] = []
+        c4d.documents.GetAllAssetsNew(
+            self.doc, allowDialogs=False, lastPath="", assetList=asset_list
+        )
+        for asset in asset_list:
+            asset_owner = asset.get("owner")
+            asset_param_id = asset.get("paramId")
+            if asset_owner and asset_param_id:
+                asset_owner[asset_param_id] = self.map_path(asset.get("filename", ""))
+
     def start_render(self, data: dict) -> None:
         self.doc = c4d.documents.GetActiveDocument()
         self.render_data = self.doc.GetActiveRenderData()
@@ -69,6 +80,8 @@ class Cinema4DHandler:
             self.render_data[c4d.RDATA_MULTIPASS_FILENAME] = self.map_path(
                 self.render_data[c4d.RDATA_MULTIPASS_FILENAME]
             )
+
+        self._remap_assets()
 
         bm = bitmaps.MultipassBitmap(
             int(self.render_data[c4d.RDATA_XRES]),


src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -50,6 +50,13 @@ def __init__(self, map_path: Callable[[str], str]) -> None:
         self.take = "Main"
         self.map_path = map_path
 
+    def _remap_assets(self):
+        asset_list = []
+        c4d.documents.GetAllAssetsNew(self.doc, allowDialogs=False, lastPath="", assetList=asset_list)
+        for asset in asset_list:
+            asset_owner = asset.get("owner")
```
joel-wong-aws, line 59: I gave this code a try but got `None` as the return value for the `.c4d` file's asset owner, leading to an error trying to index into the `None` object on the line below. 

Have you also encountered this error? This could be as simple as adding an `if asset_owner` condition.

If I only apply this owner to `non-None` asset owners, I still appear to get a segmentation fault from https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/118, but it seems like that is out of scope of his change?
```
2024/12/12 15:52:50-06:00 ADAPTOR_OUTPUT: STDERR: /usr/local/lib/python3.9/site-packages/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.sh: line 55: 185665 Segmentation fault      (core dumped) $C4DEXE ${ARGS[*]}
2024/12/12 15:52:50-06:00 openjd_fail: Error encountered while running adaptor: Cinema4D exited early and did not render successfully, please check render logs. Exit code 139
2024/12/12 15:52:50-06:00 ERROR: Entrypoint failed: STDERR: /usr/local/lib/python3.9/site-packages/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.sh: line 55: 185665 Segmentation fault      (core dumped) $C4DEXE ${ARGS[*]}
2024/12/12 15:52:50-06:00 openjd_fail: Error encountered while running adaptor: Cinema4D exited early and did not render successfully, please check render logs. Exit code 139
```

joel-wong-aws:
Thanks for the contribution! 

Does this work for any files that require path mapping or only textures?

What cases does this cover that the existing path mapping does not?
```
if self.render_data[c4d.RDATA_PATH]:
    self.render_data[c4d.RDATA_PATH] = self.map_path(self.render_data[c4d.RDATA_PATH])
```

Ahuge:
> Thanks for the contribution!
> 
> Does this work for any files that require path mapping or only textures?
> 
> What cases does this cover that the existing path mapping does not?
> 
> ```
> if self.render_data[c4d.RDATA_PATH]:
>     self.render_data[c4d.RDATA_PATH] = self.map_path(self.render_data[c4d.RDATA_PATH])
> ```

So the RDATA_PATH is just the output render path.
This case works for assets (texture files) used in the scene

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
Ahuge: Checking into that, thanks!

Ahuge: Updated with a guard and now using `paramId` instead of the hardcoded attribute

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -50,6 +50,15 @@ def __init__(self, map_path: Callable[[str], str]) -> None:
...
+            asset_param_id = asset.get("paramId")
+            if asset_owner and asset_param_id:
+                asset_owner[asset.get("paramId")] = self.map_path(asset.get("filename"))
```
karthikbekalp: Nit: We could re-use `asset_param_id` instead of calling `asset.get("paramId")`?

karthikbekalp: I'm trying to better understand this code. How does setting the `asset_owner` variable to a path with applied path mapping rules fix the issue? I don't see `asset_owner` referenced elsewhere else in the code.

karthikbekalp (commented):
Thanks for your contribution! 

Although I've a question about the fix. Feel free to respond whenever you get time. :)

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
karthikbekalp: Hmm, nevermind looks like the object from `asset.get("owner")` has its `paramId` modified and that fixes the issue?

Ahuge: Good catch, I'll update it

Ahuge: So the `asset` in this case is a dictionary containing all the information about the asset as it is used in the scene.
What we care most about are the `owner` (what node/shader/material/etc is using this asset), the `paramId` (which property in the owner that contains this asset), and the `filename` (The filename path on disk).

So `asset.get("paramId")` returns the attribute on the owner we need to modify the path of.
In most cases it seems like it will be `c4d.BITMAPSHADER_FILENAME` (for image files) but their documentation seems to want us to use the paramId, probably because there are cases where our asset is not an image file, or if a different node references the asset and doesn't store it in `c4d.BITMAPSHADER_FILENAME`.

karthikbekalp: Thanks for the detailed explanation.

karthikbekalp (changes requested):
The changes look good, Alex. Can you sign off your commit using `git commit --amend -s` so that the GitHub's DCO check passes?

Ahuge:
Updated with signoffs.


Ran tests with the [`cube.py`](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/tree/mainline/job_bundle_output_tests/redshift_textured/scene) scene as well as a client scene.

Both now do output path mapping as well as asset file mapping.
The key to the output path mapping appears to be mapping the output path *before* we set it in the scene.
Cinema4D on Linux appears to look at a Windows path that we are trying to set and realize that is not a valid Linux path and assume it must be relative.

If we path map before we set it in the scene, we end up passing a valid Linux path to Cinema 4D and everything works as expected

karthikbekalp (changes requested):
Thanks so much Alex. Unfortunately, there are linter errors now. https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/actions/runs/12319762560/job/34387883976?pr=126 

To solve this, you can run `hatch run fmt` and double check with `hatch run lint`. Sorry for the churn.

Ahuge:
> Thanks so much Alex. Unfortunately, there are linter errors now. https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/actions/runs/12319762560/job/34387883976?pr=126
> 
> To solve this, you can run `hatch run fmt` and double check with `hatch run lint`. Sorry for the churn.

No problem, thank you for pointing that out.
Done!

karthikbekalp (dismissed):
Thanks! The changes look good although I see that we have an empty file inserted `adaptor.sh` in this PR. Is this after the formatting and running the linter?

Ahuge:
> Thanks! The changes look good although I see that we have an empty file inserted `adaptor.sh` in this PR. Is this after the formatting and running the linter?

That's very peculiar. I change the executable bit on that file but it shouldn't be empty.
And in my codebase it still shows as existing.

Ahuge:
Dropping the executable bit commit to make it clearer, sorry

karthikbekalp (approved):
Thanks for removing the `adaptor.sh` change. Now the changes look good to me. :)

joel-wong-aws:
Thanks for all your work on this, Alex! I've tested this (along with https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/127) and it works great :) 

Thank you also for explaining what `c4d.RDATA_PATH` does to me.

I wouldn't consider this a breaking change, since this fixes an existing bug and shouldn't break any existing workflows that aren't already broken, so I've removed the `!` from the title.

Ship it!

joel-wong-aws (approved):

PR #127, opened 2024-12-13
fix: add path mapping rules to convert Windows paths to C4D's Linux path format

### LINKED ISSUE #118: Bug: Path mapping between a Windows scene and a Linux worker does not work as expected

joel-wong-aws:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/118

Jobs with assets that use absolute paths will fail due to missing assets

This adds additional path mapping rules for Windows -> Linux in the form of `./C:\*` to allow paths to be mapped in the intended way. In conjunction with https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/126 this applies path mapping to `RDATA_PATH`, `RDATA_MULTIPASS_FILENAME`, and all assets. 

Absolute path references for assets now work on Linux!

Submitted a physical render job (cube) with a texture that was mapped with an absolute path. It mapped the path correctly and succeeded the render!

Yes

I ran the physical job bundle output test and a test for a physical cube with a texture. On my test instance, Redshift is not working for any scene.

I added some in-code docstrings. We're still not confident enough to have official support for Linux, but both Physical and Redshift renders appear to be working on RedHat Linux types without GPUs. With GPUs, only physical renders have been working.

No, this was an existing broken path before, so it would not have been part of a customer flow.

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import os
 import re
+import sys
 import threading
 import time
 from functools import wraps
@@ -72,6 +73,32 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
     # Will be optionally changed after the scene is set.
     _expected_outputs: int = 1  # Total number of renders to perform.
     _produced_outputs: int = 0  # Counter for tracking number of complete renders.
+
+    def __init__(self, *args, **kwargs):
+        if sys.platform == "linux" and "path_mapping_data" in kwargs:
+            # on Linux, Cinema4D interprets Windows absolute paths as relative paths
+            # e.g. `C:\Users\test-user\Documents\file.bmp` becomes
+            # `./C:/Users/test-user/Documents/file.bmp`
+            # To map these paths with job attachments, we duplicate any existing Windows path mapping
+            # and then add a `./` prefix to it so that it converts correctly on Linux
+            path_mapping_data = kwargs["path_mapping_data"] or {}
+            path_mapping_rules = path_mapping_data.get("path_mapping_rules", [])
+            for rule in path_mapping_rules.copy():
+                source_path_format = rule.get("source_path_format", "")
+                # if there is no destination_os, the rule applies
+                destination_os = rule.get("destination_os", "linux")
+                if source_path_format.lower().startswith(
+                    "win"
+                ) and destination_os.lower().startswith("linux"):
+                    prefixed_rule = {
+                        "source_path": f"./{rule['source_path']}",
+                        "source_path_format": source_path_format,
+                        "destination_path": rule["destination_path"],
+                    }
+                    if "destination_os" in rule:
+                        prefixed_rule["destination_os"] = rule["destination_os"]
+                    path_mapping_rules.append(prefixed_rule)
+        super().__init__(*args, **kwargs)
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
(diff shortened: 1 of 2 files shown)


Ahuge:
Nice work! 

Looking good!

crowecawcaw (approved):
Is there a way to unit test this?

karthikbekalp (approved):
Thanks for the fix!

joel-wong-aws:
> Is there a way to unit test this?

Yes, I already have an action item to add more unit tests for the Cinema 4D integration. I'll add some tests for this as part of that.

PR #128, opened 2024-12-14
test: add sample job bundle for physical render with texture

joel-wong-aws:
When testing https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/127, I needed a physical render test case with assets. I created this job bundle test for that and I think it would be useful to others!

This test is mostly a copy of [redshift_textured](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/tree/mainline/job_bundle_output_tests/redshift_textured) but using a different texture, cube size, cube position, and asset name.

While there is some repeated code in this PR from `redshift_textured.py`, the Cinema4D user scripts run `cube.py` as `__main__`, meaning that I can't use relative imports to parent directories (making it hard to share code). To make this a standalone test (without using more complicated approaches like `exec`), I've elected to leave the duplication in this PR.

Helps to debug farm issues and also avoid regressions.

Ran `cube.py` as a user script. It generated a `.c4d` and `.bmp` file that were automatically opened. I submitted the job and ran it on a farm. The job passed and output a cube with a checkerboard texture.

No, N/A

No, N/A

No

----

### DIFF
--- a/job_bundle_output_tests/physical_textured/expected_job_bundle/asset_references.yaml
+++ b/job_bundle_output_tests/physical_textured/expected_job_bundle/asset_references.yaml
@@ -1,0 +1,10 @@
+assetReferences:
+  inputs:
+    directories: []
+    filenames:
+    - C:\normalized\job\bundle\dir\physical_textured.c4d
+    - C:\normalized\job\bundle\dir\tex\checkerboard.bmp
+  outputs:
+    directories:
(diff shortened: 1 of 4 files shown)


karthikbekalp (approved):
Thanks for adding a new test!

leon-li-inspire (approved):

PR #129, opened 2024-12-16
chore: reformat code

joel-wong-aws:
Reformatted the code using `hatch fmt`. This is the built in [hatch formatter](https://hatch.pypa.io/1.9/config/static-analysis/) (`ruff`) which is not run as part of `hatch run fmt` (which [formats using black](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/hatch.toml#L14-L17)). The `black` formatter is currently only applying a subset of the rules from `hatch fmt`.

Also updated a job bundle output test to increase test coverage and avoid a SonarCloud duplicated code error.

I ran the `physical` job bundle output test, since it was modified.

No, N/A. I decided not to include `hatch format` in `DEVELOPMENT.md` yet because it outputs a number of formatting errors e.g. `test\test_copyright_headers.py:72:5: S101 Use of 'assert' detected` which are false positives.

No

----

### DIFF
--- a/job_bundle_output_tests/physical/scene/cube.py
+++ b/job_bundle_output_tests/physical/scene/cube.py
@@ -1,4 +1,5 @@
 import os
+
 import c4d
 
 
@@ -6,8 +7,8 @@ def main():
     doc = c4d.documents.GetActiveDocument()
     doc.Flush()
     cube = c4d.BaseObject(c4d.Ocube)
-    cube[c4d.PRIM_CUBE_LEN] = c4d.Vector(200, 200, 200)
-    cube.SetAbsPos(c4d.Vector(0, 170, -170))
+    cube[c4d.PRIM_CUBE_LEN] = c4d.Vector(300, 300, 300)
+    cube.SetAbsPos(c4d.Vector(0, 170, 170))
     doc.InsertObject(cube)
     render_data = doc.GetActiveRenderData()
     render_data[c4d.RDATA_PATH] = "renders/$prj"
--- a/depsBundle.py
+++ b/depsBundle.py
@@ -4,7 +4,6 @@ import re
 import shutil
 import subprocess
 import sys
-
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any
--- a/hatch_custom_hook.py
+++ b/hatch_custom_hook.py
@@ -1,9 +1,9 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 import os
 import shutil
+from typing import Any
 
 from hatchling.builders.hooks.plugin.interface import BuildHookInterface
-from typing import Any
 
 
 class HatchCustomBuildHook(BuildHookInterface):
--- a/job_bundle_output_tests/redshift/scene/cube.py
+++ b/job_bundle_output_tests/redshift/scene/cube.py
@@ -1,4 +1,5 @@
 import os
+
 import c4d
 
 
--- a/job_bundle_output_tests/redshift_multipass/scene/cube.py
+++ b/job_bundle_output_tests/redshift_multipass/scene/cube.py
@@ -1,4 +1,5 @@
 import os
+
 import c4d
 
 
--- a/job_bundle_output_tests/redshift_takes/scene/cube.py
+++ b/job_bundle_output_tests/redshift_takes/scene/cube.py
@@ -1,4 +1,5 @@
 import os
+
 import c4d
 
 
--- a/job_bundle_output_tests/redshift_textured/scene/cube.py
+++ b/job_bundle_output_tests/redshift_textured/scene/cube.py
@@ -1,5 +1,6 @@
 import os
 import struct
+
 import c4d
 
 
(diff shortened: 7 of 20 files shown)


job_bundle_output_tests/physical/scene/cube.py
```diff
@@ -1,13 +1,14 @@
...
 import c4d
 
 
 def main():
     doc = c4d.documents.GetActiveDocument()
     doc.Flush()
     cube = c4d.BaseObject(c4d.Ocube)
-    cube[c4d.PRIM_CUBE_LEN] = c4d.Vector(200, 200, 200)
-    cube.SetAbsPos(c4d.Vector(0, 170, -170))
+    cube[c4d.PRIM_CUBE_LEN] = c4d.Vector(300, 300, 300)
```
karthikbekalp, line 10: Hmm, any reason the dimensions and positions were changed?

karthikbekalp (commented):
Thanks for reformatting the code. 

>  decided not to include hatch format in DEVELOPMENT.md yet because it outputs a number of formatting errors e.g. test\test_copyright_headers.py:72:5: S101 Use of 'assert' detected which are false positives.

If this tool generally gives false positives, should we remove it? I say this because of 2 reasons: 

- This can reduce the amount of dependencies (i.e. ruff) that we may need to update.
- As I understand, we may need to manually reformat this again later to ensure the formatting is correct? So we need to reformat the code periodically every few days? 
How about other repositories? How do they manage this?

job_bundle_output_tests/physical/scene/cube.py
joel-wong-aws: I updated this test to increase test coverage and avoid a SonarCloud duplicated code error!

karthikbekalp: Cool, thanks!

joel-wong-aws:
> If this tool generally gives false positives, should we remove it? I say this because of 2 reasons:

I think the issue is that based on our current (lack of) `ruff` configuration, the tool doesn't have enough information to determine what to flag. If we were to add some `ruff` configuration, that would remove the false positives

> * This can reduce the amount of dependencies (i.e. ruff) that we may need to update.

[hatch fmt](https://hatch.pypa.io/1.13/config/internal/static-analysis/) (including `ruff`) is built into hatch, and we use hatch for various reasons, so we can't remove it altogether

> * As I understand, we may need to manually reformat this again later to ensure the formatting is correct? So we need to reformat the code periodically every few days?

Correct, the tooling right now doesn't run automatically in our CI workflows. We would need to add some configuration to get this to run, including to remove those false positives. There are also some legitimate code style issues flagged by `hatch fmt` that can't be automatically fixed.

> How about other repositories? How do they manage this?

As far as I am aware, they only use `hatch run fmt` instead of `hatch fmt`. `hatch run fmt` has a subset of rules. Ideally, we should improve the linting of all repositories by running `hatch fmt` on them, then adding `hatch fmt --check` to the [CI runner definition]( https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_python_build.yml), but that will require a lot of refactoring to all codebases, so we can start by simply applying the rules here with `hatch fmt`.

karthikbekalp:
> > If this tool generally gives false positives, should we remove it? I say this because of 2 reasons:
> 
> I think the issue is that based on our current (lack of) `ruff` configuration, the tool doesn't have enough information to determine what to flag. If we were to add some `ruff` configuration, that would remove the false positives
> 
> > * This can reduce the amount of dependencies (i.e. ruff) that we may need to update.
> 
> [hatch fmt](https://hatch.pypa.io/1.13/config/internal/static-analysis/) (including `ruff`) is built into hatch, and we use hatch for various reasons, so we can't remove it altogether
> 
> > * As I understand, we may need to manually reformat this again later to ensure the formatting is correct? So we need to reformat the code periodically every few days?
> 
> Correct, the tooling right now doesn't run automatically in our CI workflows. We would need to add some configuration to get this to run, including to remove those false positives. There are also some legitimate code style issues flagged by `hatch fmt` that can't be automatically fixed.
> 
> > How about other repositories? How do they manage this?
> 
> As far as I am aware, they only use `hatch run fmt` instead of `hatch fmt`. `hatch run fmt` has a subset of rules. Ideally, we should improve the linting of all repositories by running `hatch fmt` on them, then adding `hatch fmt --check` to the [CI runner definition](https://github.com/aws-deadline/.github/blob/mainline/.github/workflows/reusable_python_build.yml), but that will require a lot of refactoring to all codebases, so we can start by simply applying the rules here with `hatch fmt`.

Cool, that makes sense. Let's start with Cinema 4D then. Approving the PR.

karthikbekalp (approved):
Discussed about the issues in the thread in this PR. Looks good to me.

leon-li-inspire (approved):

job_bundle_output_tests/physical/scene/cube.py
```diff
@@ -1,13 +1,14 @@
 import os
+
```
baxeaz, line 2: Huh, there's a specific rule that says the c4d import comes at the end after a newline?

joel-wong-aws: Yeah, according to [the PEP8 Python style guide](https://peps.python.org/pep-0008/#imports):

> Imports should be grouped in the following order:
> 1 .Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.

PR #130, opened 2024-12-17
draft: fix: Ensure filenames are encoded in UTF-8.

karthikbekalp:

Cinema 4D submissions fail to run if filenames contain non-ASCII values. For example, a scene file named `TestSceneñ.c4d` would not render.

```
openjd_fail: Error encountered while starting adaptor: 
Cinema4D Encountered an Error: FileNotFoundError: The scene file 'C:\some_path\TestñSceneñ.c4d' does not exist
```

In our handler, [we check if a file exists](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py#L178-L179) before loading it into Cinema 4D.

However, Windows systems use cp1252 encoding by default, not UTF8. This causes incorrect filename encoding and failed file existence checks, preventing renders. Encoding to UTF8 fixes most renders.

While renders now work for characters like 
- Symbols like for Bitcoin ₿
- Latin character ę 
- Latin character ñ
- Greek character β
- Cyrillic Б  
They still fail for emojis and some mathematical operators. [Cinema 4D's load document function returns None](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py#L180) without details on why it fails.

Regardless, this is a partial fix that should help most of our customers.
I will also investigate further if its possible to have general solution that can fix it for all DCCs that have this issue. 

Customers with scene files containing non-ASCII characters in filenames can now use them. This excludes filenames with emojis or mathematical symbols.

Yes

Yes, I also added a new test to check for this condition specifically. 

No

No

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import os
 from typing import Any, Callable, Dict
+import locale
 
 try:
     import c4d  # type: ignore
@@ -38,6 +39,7 @@ class Cinema4DHandler:
         """
         Constructor for the c4dpy handler. Initializes action_dict and render variables
         """
+
         self.action_dict = {
             "scene_file": self.set_scene_file,
             "take": self.set_take,
@@ -164,6 +166,16 @@ class Cinema4DHandler:
         """
         self.render_kwargs["frame"] = int(data.get("frame", ""))
 
+    def _convert_filename_to_utf8_encoding(self, scene_file: str) -> str:
+        try:
+            return scene_file.encode(locale.getpreferredencoding()).decode("utf-8")
+        except UnicodeError as ue:
+            print(f"Error: Encoding to UTF-8 failed due to unicode error. Exception: {ue}")
+        except Exception as e:
+            print(f"Error: Encoding to UTF-8 failed due to unexpected error. Exception: {e}")
+        print("Returning scene file string as is as encoding/decoding failed.")
+        return scene_file
+
     def set_scene_file(self, data: dict) -> None:
         """
         Opens the scene file in Cinema4D.
@@ -175,6 +187,9 @@ class Cinema4DHandler:
             FileNotFoundError: If path to the scene file does not yield a file
         """
         scene_file = data.get("scene_file", "")
+
+        scene_file = self._convert_filename_to_utf8_encoding(scene_file)
+
         if not os.path.isfile(scene_file):
             raise FileNotFoundError(f"The scene file '{scene_file}' does not exist")
         doc = c4d.documents.LoadDocument(
--- a/job_bundle_output_tests/redshift_textured_with_nonascii_characters/expected_job_bundle/asset_references.yaml
+++ b/job_bundle_output_tests/redshift_textured_with_nonascii_characters/expected_job_bundle/asset_references.yaml
@@ -1,0 +1,12 @@
+assetReferences:
+  inputs:
+    directories: []
+    filenames:
+    - C:\normalized\job\bundle\dir\redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411\
+      _\u062A.c4d
+    - C:\normalized\job\bundle\dir\tex\checkerboard-_\u20BF_\u0119_\xF1_\u03B2_\u0411\
+      _\u062A.bmp
+  outputs:
+    directories:
+    - C:\normalized\job\bundle\dir\renders
+  referencedPaths: []
--- a/job_bundle_output_tests/redshift_textured_with_nonascii_characters/expected_job_bundle/parameter_values.yaml
+++ b/job_bundle_output_tests/redshift_textured_with_nonascii_characters/expected_job_bundle/parameter_values.yaml
@@ -1,0 +1,18 @@
+parameterValues:
+- name: Cinema4DFile
+  value: C:\normalized\job\bundle\dir\redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411\
+      _\u062A.c4d
+- name: OutputPath
+  value: C:\normalized\job\bundle\dir\renders\redshift_textured
+- name: MultiPassPath
+  value: ''
+- name: Frames
+  value: '1'
+- name: deadline:targetTaskRunStatus
+  value: READY
+- name: deadline:maxFailedTasksCount
+  value: 20
+- name: deadline:maxRetriesPerTask
+  value: 5
+- name: deadline:priority
+  value: 50
--- a/job_bundle_output_tests/redshift_textured_with_nonascii_characters/expected_job_bundle/template.yaml
+++ b/job_bundle_output_tests/redshift_textured_with_nonascii_characters/expected_job_bundle/template.yaml
@@ -1,0 +1,103 @@
+specificationVersion: jobtemplate-2023-09
+name: physical-_₿_ę_ñ_β_Б_ت.c4d
+parameterDefinitions:
+- name: Cinema4DFile
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  userInterface:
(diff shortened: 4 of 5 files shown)


src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -175,6 +187,10 @@ def set_scene_file(self, data: dict) -> None:
             FileNotFoundError: If path to the scene file does not yield a file
         """
         scene_file = data.get("scene_file", "")
+
+        if locale.getpreferredencoding().lower() not in ("utf8", "utf-8"):
```
crowecawcaw: Could we always encode and decode and avoid the branch here?

crowecawcaw (dismissed):

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
karthikbekalp: I was considering not modifying filenames for UTF-8 default encoding file systems, as it usually wouldn't change anything. Is there any reason to do it other than avoiding the branch?

But I do see your point that as most of our rendering is in Windows, we could always encode/decode. Rev 2 coming up.

karthikbekalp:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud-for-cinema-4d&pullRequest=130) **Quality Gate failed**
> 
> Failed conditions  [26.8% Duplication on New Code](https://sonarcloud.io/component_measures?id=aws-deadline_deadline-cloud-for-cinema-4d&pullRequest=130&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=aws-deadline_deadline-cloud-for-cinema-4d&pullRequest=130)

We can override merge even with this check failing. We should plan on refactoring all the tests in this folder so that we re-use the code amongst the tests instead of rewriting. But it is out of scope for this PR.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -164,6 +166,16 @@ def set_frame(self, data: dict) -> None:
         """
         self.render_kwargs["frame"] = int(data.get("frame", ""))
 
+    def _convert_filename_to_utf8_encoding(self, scene_file: str) -> str:
+        try:
+            return scene_file.encode(locale.getpreferredencoding()).decode("utf-8")
```
mwiebe, line 171: This doesn't look right to me, because the Python 3 `str` type is already unicode. If this fixes something, it means that there was an earlier processing step that was wrong. We should track down that earlier step instead of patching it after it's already wrong.

karthikbekalp: > because the Python 3 str type is already unicode

Yup, that tripped me as well when I was trying to root cause this issue. Turns out python uses Unicode but it does not use UTF-8 encoding by default. The default encoding is platform dependent and in Windows it is CP-1252. This is what I found out while investigating this issue in the official docs. (Its for the [`open()` library function](https://docs.python.org/3/library/functions.html#open) but I think it uses the same philosophy here) 

> We should track down that earlier step instead of patching it after it's already wrong.

I traced the issue back to adaptor.py while investigating, which showed the problem even before Cinema 4D started running.

```
...
2024/12/18 09:52:41-06:00 ADAPTOR_OUTPUT: INFO: Loading 'init_data' schema from C:\Program Files\Python312\Lib\site-packages\deadline\cinema4d_adaptor\Cinema4DAdaptor\schemas\init_data.schema.json
2024/12/18 09:52:41-06:00 ADAPTOR_OUTPUT: INFO: Loading 'run_data' schema from C:\Program Files\Python312\Lib\site-packages\deadline\cinema4d_adaptor\Cinema4DAdaptor\schemas\run_data.schema.json
2024/12/18 09:52:42-06:00 ADAPTOR_OUTPUT: INFO: In Adaptor: Scene file is C:\ProgramData\Amazon\OpenJD\session-xxx\assetroot-yyyy\scene\redshift_textured_β.c4d
2024/12/18 09:52:42-06:00 ADAPTOR_OUTPUT: INFO: In adaptor: Scene file is file? False
...
```

Patching it in the adaptor unfortunately causes the adaptor to freeze while running. It does not happen often but only at times and I tried root causing the issue but couldn't find it. I patched it in handler as it was the most straightforward fix and quick fix to unblock customers. 

```
2024/12/18 10:14:37-06:00 ADAPTOR_OUTPUT: INFO: In Adaptor: Scene file is C:\ProgramData\Amazon\OpenJD\session-xxx\assetroot-yyyy\scene\redshift_textured-_₿_ę_ñ_β_Б_ت.c4d
2024/12/18 10:14:37-06:00 ADAPTOR_OUTPUT: INFO: In adaptor: Scene file is file? False

...
It was stuck and I had to fail the job manually. 
...
2024/12/18 10:18:44-06:00 Canceling subprocess 3868 via notify then terminate method at 2024-12-18T16:15:44Z.
```

Initially, this seemed to be a Cinema 4D-specific issue, not occurring in other DCCs like Unreal and Keyshot on Windows. Hence, I ruled out checking the dependencies of Cinema 4D. 
However, while reviewing logs yesterday evening, I discovered that Keyshot jobs, despite being successful, had a similar error in the adaptor logs. This suggests the issue could affect all DCCs that run in Windows.

```
2024/12/17 12:49:39-06:00 ADAPTOR_OUTPUT: STDOUT:     self._perform_action(action)
2024/12/17 12:49:39-06:00 ADAPTOR_OUTPUT: STDOUT:   File "C:\Users\job-user\.conda\envs\Lib\openjd\adaptor_runtime_client\base_client_interface.py", line 233, in _perform_action
2024/12/17 12:49:39-06:00 ADAPTOR_OUTPUT: STDOUT:     action_func(a.args)
2024/12/17 12:49:39-06:00 ADAPTOR_OUTPUT: STDOUT:   File "C:\Users\job-user\.conda\envs\hashname_xxx\Lib\deadline\keyshot_adaptor\KeyShotClient\keyshot_handler.py", line 110, in set_scene_file
2024/12/17 12:49:39-06:00 ADAPTOR_OUTPUT: STDOUT:     raise FileNotFoundError(f"The scene file '{scene_file}' does not exist")
2024/12/17 12:49:39-06:00 ADAPTOR_OUTPUT: STDOUT: FileNotFoundError: The scene file 'C:\ProgramData\Amazon\OpenJD\session-xxx\assetroot-yyy\AppData\Local\Temp\tmpd3n517x9\unpack\Keyframe-Robot-Scene Sharing Version  ñot Angry.bip' does not exist
...
```

Since Cinema 4D customers were currently impacted and are blocked for working, I prioritized fixing it for Cinema 4D first. After that, I plan to investigate the issue further.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -3,6 +3,7 @@
 
 import os
 from typing import Any, Callable, Dict
+import locale
```
joel-wong-aws, line 6: nit: can you run `hatch fmt` on this? I would expect the imports to be in a different order.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
mwiebe: Have you looked into creating a minimal reproducer example as a small job template? That could help isolate the issue from the environment and provide faster iteration to debug it.

joel-wong-aws (approved):

karthikbekalp:
Closing this PR in favor for https://github.com/OpenJobDescription/openjd-adaptor-runtime-for-python/pull/171

PR #131, opened 2024-12-18
chore: remove unnecessary code from create_adaptor_packaging_artifact.sh

joel-wong-aws:
When running the adaptor packaging artifact script, I learned that some of the code was extraneous.

Removed it.

No, N/A

No, N/A

No, N/A

----


karthikbekalp (approved):
Nice job, Joel!

crowecawcaw (approved):

PR #144, opened 2025-01-10
fix: initialize connection to Maxon assets DB on Windows to fix confusing error message

joel-wong-aws:
When running Cinema 4D on a Windows EC2 instance for the first time, the following message is output when launching Cinema 4D:
```
sslConnection.DoHandShake failed while connecting to https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat:
            SSL_do_handshake returned error: SSL_ERROR_SSL (certificate verify failed Certificate Information: not available)
```
After investigation, it was determined that this is NOT a security issue. Because the SSL handshake fails, the connection is NOT made at all.

However, the asset database may not be updated on the launch. It's unclear if this has a functional impact. To work around the issue, we initialize the connection to the Maxon assets database when launching Cinema 4D by running `curl https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat` on adaptor startup.

Prevents confusing error messages and ensures the assets database is updated.

I created a Windows SMF fleet on Deadline Cloud. Upon submitting a job to a queue associated with the fleet, it autoscaled to add a new worker.

Before the code was merged, the new worker received the above error on the first Cinema 4D launch.

When I added this code, I instead got the following on the first Cinema 4D launch: `Newer version of asset database found: https://assets.maxon.net/assets/MaxonAssets.db (2025-01-03 16:48:20 / ---)`

#### Have you run the unit tests?
Yes, they all passed

#### Did you run the job bundle output tests
I ran the physical cube job bundle output test.

No, N/A

No

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -182,6 +182,33 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
         self._server = AdaptorServer(self._action_queue, self)
         self._server.serve_forever()
 
+    def _initialize_maxon_assets_db_connection(self) -> None:
+        """
+        When starting `Commandline.exe` the first time on Windows, Cinema 4D sometimes has the following error:
+        ```
+        sslConnection.DoHandShake failed while connecting to https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat:
+            SSL_do_handshake returned error: SSL_ERROR_SSL (certificate verify failed Certificate Information: not available)
+        ```
+        This indicates that a secure connection was not able to be made to the Maxon assets DB, so Cinema 4D
+        did not connect to it at all. (i.e. it only connects if the connection is secure)
+
+        However, Cinema 4D appears to quit prematurely the first time because on subsequent retries, the secure
+        handshake will succeed.
+
+        To prevent this from occurring, we pre-emptively initialize the connection.
+
+        The Maxon assets DB is used to pull in assets from Maxon and if outdated, recently released assets may not be
+        available.
+        """
+        if sys.platform in ["win32", "cygwin"]:
+            _logger.info("Initializing the Maxon assets DB connection")
+            curl_subprocess = LoggingSubprocess(
+                args=["curl", "https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat"]
+            )
+            # Note that if curl has an error, this will not re-raise the error. This is intended because updating the
+            # asset database does not affect rendering in most cases and there is NO security risk if it has an error.
+            curl_subprocess.wait()
+
     def _start_cinema4d_server_thread(self) -> None:
         """
         Starts the cinema4d adaptor server in a thread.
@@ -415,6 +442,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
         self.validators.init_data.validate(self.init_data)
 
         self.update_status(progress=0, status_message="Initializing Cinema4D")
+        self._initialize_maxon_assets_db_connection()
         self._start_cinema4d_server_thread()
         self._populate_action_queue()
         self._start_cinema4d_client()


karthikbekalp (dismissed):
Discussed offline, looks good to me.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -182,6 +182,30 @@ def _start_cinema4d_server(self) -> None:
...
+        However, Cinema 4D appears to quit prematurely the first time because on subsequent retries, the secure
+        handshake will succeed.
+
+        To prevent this from occurring, we pre-emptively initialize the connection.
+        """
+        if sys.platform in ["win32", "cygwin"]:
+            _logger.info("Initializing the Maxon assets DB connection")
+            curl_subprocess = LoggingSubprocess(
+                args=["curl", "https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat"]
+            )
```
crowecawcaw, line 207: Does it work equally well if you use Python to make a request instead of curl? This introduces a dependency on curl.

crowecawcaw (commented):
This seems like an unintuitive solution to the error. Are you pretty confident that this fix works? Like you can replicate the issue and then see it fixed with this trick? If it works, it works!

joel-wong-aws:
> This seems like an unintuitive solution to the error. Are you pretty confident that this fix works? Like you can replicate the issue and then see it fixed with this trick? If it works, it works!

Yeah, I did a lot of testing to verify that this actually works. Specifically, I created a Windows SMF fleet on Deadline Cloud. Upon submitting a job to a queue associated with the fleet, it autoscaled to add a new worker.

Before the code was merged, the new worker received the error mentioned in the description on the first Cinema 4D launch.

When I added this code, I instead got the following on the first Cinema 4D launch: `Newer version of asset database found: https://assets.maxon.net/assets/MaxonAssets.db (2025-01-03 16:48:20 / ---)`

I added the above details to the PR description. I have tried this multiple times and got the same behaviour everytime.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
joel-wong-aws: I'll take a look at Python's standard libraries to see if any of them work!

joel-wong-aws: I tried this with the built-in `urllib` and got a 403 error from `https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat`:
```
>>> import urllib.request
>>> urllib.request.urlopen("https://assets.maxon.net/assets/MaxonAssets.db/_index/modified.dat")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Program Files\Python312\Lib\urllib\request.py", line 215, in urlopen
    return opener.open(url, data, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\urllib\request.py", line 521, in open
    response = meth(req, response)
               ^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\urllib\request.py", line 630, in http_response
    response = self.parent.error(
               ^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\urllib\request.py", line 559, in error
    return self._call_chain(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\urllib\request.py", line 492, in _call_chain
    result = func(*args)
             ^^^^^^^^^^^
  File "C:\Program Files\Python312\Lib\urllib\request.py", line 639, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 403: Forbidden
```

I then tried this with `http.client.HTTPSConnection` ([code](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/144/commits/22731c91c76c1a9353b718bef10e361942148136)) and got:
`2025/01/10 17:09:31-06:00 ADAPTOR_OUTPUT: INFO: Could not initialize the Maxon assets DB connection. This does not affect rendering in most case. Error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1006)`

Neither of these fix the issue and I don't see any other simple ways to use python to do this. Yes, this adds a dependency on `curl`, but it's the only approach I've found which consistently worked. It's [been on Windows since 2018](https://devblogs.microsoft.com/commandline/tar-and-curl-come-to-windows/), so it should have pretty wide availability. [Maxon requires Windows 10 20H2 as a minimum](https://www.maxon.net/en/requirements/cinema-4d-2024-requirements), and that was released in 2020, so it will have `curl` on it.


We can always convert to a different approach later as well.

karthikbekalp (approved):

crowecawcaw (approved):

Issue #84, opened 2024-11-07
Bug: LongPathsEnabled on Windows seems to be ignored, and results in read/write failure when paths > 256 chars.


andymanaws:
### Expected Behaviour

By enabling LongPathsEnabled on Windows you can utilise assets, and render outputs, that have >256 chars.

But renders via Deadline Cloud seem to throw errors regard file lengths exceeding 256 chars.

### Current Behaviour

Job attachments assets will result in an error when syncing:

"your file path is longer than what Windows allow."

Output images fail to write in job attachments:

"Found 0 files (Output directory C:\ProgramData\Amazon\OpenJD\session-0fac1829f3cf4e848c2c18f474f8987e8oe..."

### Reproduction Steps

Ensure Windows is configured correctly:
https://www.autodesk.com/support/technical/article/caas/sfdcarticles/sfdcarticles/The-Windows-10-default-path-length-limitation-MAX-PATH-is-256-characters.html

Use a texture that has a lot of characters - e.g. include 

bobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbobbob

Set the output location to a folder which has >256 chars.

Render from the app - it works.
Render from Deadline - it fails.

### Code Snippet

N/A

karthikbekalp:
Hmm, I believe there was a fix explicitly to notify customers when long paths are not enabled in https://github.com/aws-deadline/deadline-cloud/pull/403/files. @AlexTranAmz do you have any ideas why this could happen even if an instance is configured properly?

karthikbekalp:
Discussed with @AlexTranAmz, looks like we need to enable the long paths in the registry as well as add UNC paths (add \\?\ to the start of the path instead of C:\ or D:\).

joel-wong-aws:
I did an initial investigation of this issue. To my understanding, Cinema 4D itself does not support long paths. I have submitted a ticket to Maxon about this.

That said, our Deadline Cloud integration code exacerbates the problem in a couple ways:

[...]

PR #148, opened 2025-01-16
fix: improve error handling for sticky settings having a long path

joel-wong-aws:
Relates to https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/84

Cinema 4D does not support long paths. Because of this, there are various errors when attempting to submit jobs with long path files. One of the errors is:

olderror

Handle the error where the sticky settings file cannot be created by raising a more useful error message. I chose to raise an error for the long path case rather than simply ignoring the failure because there will be [further issues later in submission anyway](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/84#issuecomment-2596345886) if the .c4d file path is not shortened.

If the issue is not a long path issue, then only a warning is logged and the submission is allowed.

Easier debugging for long path issues.

Created a `.c4d` file with a long path where the sticky settings file would make the file exceeds Cinema 4D's max path length.

updatederror

Yes

No, N/A

No, N/A

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/data_classes.py
+++ b/src/deadline/cinema4d_submitter/data_classes.py
@@ -3,12 +3,16 @@ from __future__ import annotations
 
 import dataclasses
 import json
+import logging
+import traceback
 from dataclasses import dataclass, field
 from pathlib import Path
 
 from .takes import TakeSelection  # type: ignore
 
 RENDER_SUBMITTER_SETTINGS_FILE_EXT = ".deadline_render_settings.json"
+
+_logger = logging.getLogger(__name__)
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -58,21 +62,32 @@ def load_sticky_settings(self, scene_filename: str):
...
+            with open(sticky_settings_filename, "w", encoding="utf8") as fh:
+                obj = {
+                    field.name: getattr(self, field.name)
+                    for field in dataclasses.fields(self)
+                    if field.metadata.get("sticky")
+                }
+                json.dump(obj, fh, indent=1)
+        except OSError as e:
+            traceback.print_exc()
+            if len(sticky_settings_path) >= 256:
```
karthikbekalp, line 86: Should we shorten the file path further (to 200) to avoid the same issue on the worker side when other prefixes are added?

joel-wong-aws: I think we should address that separately. If textures are long, they will have the same issue. I think we need a more comprehensive solution for all files that considers the input file structure against the worker file structure. However, this issue with a long sticky settings file will occur by itself will likely still occur even with a more comprehensive solution.

karthikbekalp: Ok. Thanks for your clarification.

karthikbekalp (approved):

crowecawcaw:
Are long paths something we could reasonably support in the future? If so, maybe we could make an issue with an Enhancement label for it.

crowecawcaw (approved):

joel-wong-aws:
> Are long paths something we could reasonably support in the future? If so, maybe we could make an issue with an Enhancement label for it.

Unfortunately, some parts of Cinema 4D itself do not support long paths, so we can't add long path support until it is supported by C4D. Until then, we will simply improve our error handling logic. Note that once full long path support is fully added to C4D, this code path will no longer error and the sticky settings will save correctly.

PR #152, opened 2025-01-21
fix!: handle custom frame ranges

joel-wong-aws:
Unit test coverage is lacking for the `deadline-cloud-for-cinema-4d integration`. This increases risks of changes mistakenly causing regressions and errors in implementation.

Added some unit tests for `cinema4d_submitter/scene.py`. In the process, noticed that the handling of custom frame ranges was incorrect. Added correct handling of custom frame ranges and unit tests.

Custom frame ranges will now work!

1. Updated the C4D submitter `scene.py` file on my workstation. 
2. Selected Render > Render Settings > Custom Frames
3. Input `1,4-5,11-20:2`
4. Go to Extensions > AWS Deadline Cloud Submitter > Submit
a) Before change: Frame 0 was sent to Deadline Cloud
b) After change: Frames 1, 4, 5, 11, 13, 15, 17, and 19 were sent to Deadline Cloud

No, N/A

Yes, this removes the unused `Scene` `__iter__` method and also requires the `data` argument in `Animation`'s method for `current_frame`, `start_frame`, `end_frame`, and `frame_step`. It also changes the return type of `Animation`s `frame_list` function to a `str` and removes the `extension_padding` function. While we don't know of any use cases where these functions would be called externally, this is technically a breaking change.

----

### DIFF
--- a/test/deadline_submitter_for_cinema4d/unit/test_scene.py
+++ b/test/deadline_submitter_for_cinema4d/unit/test_scene.py
@@ -167,7 +167,7 @@ class TestFrameRange:
                 "1-100:7",
                 id="start=1, stop=100, step=7",
             ),
-            pytest.param(4, 10, None, "4-10", id="start=1, stop=100, step=None"),
+            pytest.param(4, 10, None, "4-10", id="start=4, stop=10, step=None"),
             pytest.param(6, 14, 2, "6-14:2", id="start=6, stop=14, step=2"),
             pytest.param(1, None, 7, "1", id="start=1, stop=None, step=7"),
             pytest.param(10, 10, 10, "10", id="start=10, stop=10, step=10"),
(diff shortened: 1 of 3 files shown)


crowecawcaw (dismissed):

test/deadline_submitter_for_cinema4d/unit/test_scene.py
```diff
@@ -29,44 +29,161 @@ def test_get_output_directores():
...
+                id="start=2, stop=13, step=3",
+            ),
+            pytest.param(
+                1,
+                100,
+                7,
+                "1-100:7",
+                id="start=1, stop=100, step=7",
+            ),
+            pytest.param(4, 10, None, "4-10", id="start=1, stop=100, step=None"),
```
baxeaz: Should this be start=4 stop=10?

joel-wong-aws: Yes! Great spotting! Updated!

crowecawcaw (approved):

baxeaz (approved):

PR #163, opened 2025-02-04
fix: Use RDATA_FRAMESEQUENCE_CUSTOM only if it exists.

karthikbekalp:

There was an issue in https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/152/files as the `RDATA_FRAMESEQUENCE_CUSTOM` is a new variable and does not exist on Cinema 4D 2024 or older releases of Cinema 4D 2025. 

We first check if the attribute exists before doing a `c4d.RDATA_FRAMESEQUENCE_CUSTOM` and this should avoid the errors. 

Customers can work on Cinema4D 2024 versions as well. 

@joel-wong-aws helped me run the tests manually as my UI elements were broken. 

Yes

No

----


joel-wong-aws (approved):

joel-wong-aws:
I have tested this on my machine with Cinema 4D 2024 and it works as expected (no errors). I retested with Cinema 4D 2025 and it also still worked with custom frames.

viknith (approved):

PR #166, opened 2025-02-05
chore: Refactor the Cinema 4D submitter for tests.

karthikbekalp:
### What was the problem/requirement? (What/Why)

We are working on adding integration tests that can automate our release process. 

Unfortunately, our submitter was tightly coupled and hence made testing difficult. 

### What was the solution? (How)

This improves on our submitter code by refactoring it into smaller functions that can be relatively easily tested. 

For more information on how our tests would look, take a look here: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/compare/mainline...karthikbekalp:deadline-cloud-for-cinema-4d:integp1 

### What is the impact of this change?

Improves our code so that it can be tested more efficiently. This is a refactor so there is no customer changes. 

### How was this change tested?

I ran our Job bundle output tests. Additionally, I also compared the outputs of the physical render without my changes to ensure there are no changes between the `template.yaml`, `parameter_values.yaml` or the `asset_references.yaml`. 

### Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.

Yes. 

Screenshot 2025-02-05 at 10 38 05 AM

### Was this change documented?
No, this is just a refactor. 

### Is this a breaking change?
No

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #170, opened 2025-02-07
test: Add submitter automated test setup.

karthikbekalp:

We are working on adding integration tests that can automate our release process. 

This CR is one in a series of tests that will help us gain confidence in our code. 

- CR to refactor the submitter code: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/166
- This CR to add the submitter tests: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/170
- Upcoming CRs: (to add adaptor tests)

Adds integration tests that devs can run locally. 

I setup the `C4D_PYTHON` env location to point to c4dpy for running the scripts. 

No, this is just for tests. This should have no customer impact. 

No

----


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,15 +18,6 @@ See [DEVELOPMENT.md](https://github.com/aws-deadline/deadline-cloud-for-cinema-4
 - Have you run the unit tests?
 *delete text ending here*
 
-### Did you run the "Job Bundle Output Tests"? If not, why not? If so, paste the test results here.
```
karthikbekalp, line 21: Removed this step and I'll replace it with instructions to run the integration tests once all the job bundle tests are moved to the integration tests.

src/deadline/cinema4d_submitter/integ_test_helpers.py
```diff
@@ -0,0 +1,104 @@
...
+    This is an example of host_requirements for a submission.
+    """
+    return {
+        "attributes": [
+            {"name": "attr.worker.os.family", "anyOf": ["windows"]},
+        ]
+    }
+
+
+def internal_create_job_bundle(job_bundle_dir: str):
```
joel-wong-aws, line 81: I think this is great progress and coverage!

Is there a way we can expand coverage by adding optional `parent` and `test` arguments to `show_submitter` i.e. `def show_submitter(parent: QWidget=None, test=False)` and then actually open the window (using the `test` argument to avoid blocking the thread) and programmatically clicking on "export" (using the `parent` reference) and comparing the exported job bundle to the expected result?

This would also make it super easy to convert to run on actual workers - just programmatically click "submit" instead of "export"!

Also, a nit: would a better name for this function be `def simulate_submitter`?

test/integ/test_cinema4d.py
```diff
@@ -0,0 +1,151 @@
...
+
+def test_submitter(
+    cinema4d_location: Path,
+    test_scenes_folder_location: Path,
+) -> None:
+    """
+    Test the submitter code.
+    """
+    test_scene_folder_location = (
+        test_scenes_folder_location / "redshift_textured_with_nonascii_characters"
```
joel-wong-aws, line 134: It would be great to do this in a for loop for all sample scenes

karthikbekalp: Yup, that's the intention. Once we are happy with the structure, I'll just parametrize the tests so that we can run all the tests one by one.

src/deadline/cinema4d_submitter/integ_test_helpers.py
karthikbekalp: > Is there a way we can expand coverage by adding optional parent and test arguments to show_submitter i.e. def show_submitter(parent: QWidget=None, test=False) and then actually open the window

Hmm, I was not sure if that was possible. I'll try to see if we can manipulate the Qt widgets as you suggested.

DEVELOPMENT.md
```diff
@@ -88,16 +88,14 @@ To run the unit tests, simply use hatch:
...
-5. Submit or Export Bundle and confirm that the scene is rendered successfully. (The output should have a cube floating)
-6. Repeat the steps for all the test scenes in the folder.
+##### Integration Tests (Currently WIP)
+
+We run the integration tests by running it in Cinema 4D's python i.e. c4dpy. You would have to setup Cinema 4D licensing before you run the tests.
+
+1. Set the environment variable `C4D_PYTHON` to the location of the Cinema 4D python.
+   1. `set C4D_PYTHON=<c4dpy location>` on Windows Command or `$env:C4D_PYTHON = <c4dpy location>` on Windows Powershell.
+      1. The default location for `c4dpy` on Windows is `C:\Program Files\Maxon Cinema 4D 2025\c4dpy`
+3. Run `hatch run integ:test`
```
crowecawcaw, line 130: Optional idea: can we embed these instructions in the `hatch run integ:test` command somehow? Maybe it prints out instructions for how to set the path if it isn't set already, or maybe it asks for the path and provides the default.

DEVELOPMENT.md
```diff
@@ -88,16 +88,14 @@ To run the unit tests, simply use hatch:
...
-1. Launch Cinema 4D
-2. Go to Extensions tab > User Scripts > Run Script ...
-3. Select script file "./job_bundle_output_tests/{test}/scene/cube.py" within the GitHub repo.
-    The script builds the scene and saves it to "./job_bundle_output_tests/{test}/scene/{test}.c4d"
-4. Select Extensions tab > Deadline Cloud Submitter
-5. Submit or Export Bundle and confirm that the scene is rendered successfully. (The output should have a cube floating)
-6. Repeat the steps for all the test scenes in the folder.
+##### Integration Tests (Currently WIP)
+
+We run the integration tests by running it in Cinema 4D's python i.e. c4dpy. You would have to setup Cinema 4D licensing before you run the tests.
```
crowecawcaw: Wording nit: "We run integration tests in Cinema 4D's python."

Also could you write a little more about the integ architecture? It looks like pytest runs in C4D's python interpreter and that gives it access to C4D's API. And it looks like we're loading a real scene, running submitter functions on it, writing out a job bundle, and comparing the job bundle against a reference.

DEVELOPMENT.md
karthikbekalp: That's a great idea. I'll try to update the message so that it is clearer when there are failures.

src/deadline/cinema4d_submitter/integ_test_helpers.py
karthikbekalp: I was able to suppress the dialogs that are opened by Cinema 4D submitter code but unfortunately we would get message boxes that are opened which are blocking UI components from deadline-cloud [over here](https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/ui/dialogs/submit_job_to_deadline_dialog.py#L380-L388). 

This would cause a message box that has to be manually closed and other UI side effects. :(
I think refactoring this is not trivial and out of scope for this PR. Hence, I suggest that we keep the current scope of code coverage.

DEVELOPMENT.md
karthikbekalp: Makes sense, updated!

test/integ/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.fixture
+def cinema4d_location() -> Path:
+    if "C4D_PYTHON" not in os.environ:
+        raise EnvironmentError(
+            "Environment variable 'C4D_PYTHON' not found. "
+            "Please set the C4D_PYTHON environment variable to the location of the Cinema 4D python. "
+            r"The default location for `c4dpy` on Windows is 'C:\Program Files\Maxon Cinema 4D 2025\c4dpy'"
+        )
```
crowecawcaw, line 16: Improvement idea: we could check the default location and try running a script there. If it works, we continue on. If it fails, we throw this error message. Than if someone has a default install, they don't need to set the env variable and we can remove the instructions about it.

crowecawcaw (approved):

test/integ/conftest.py
karthikbekalp: I can do that. I've an upcoming PR where I can make this improvement if you don't mind. :)

test/integ/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+
+import pytest
+import os
+
+from pathlib import Path
+
+
+@pytest.fixture
+def cinema4d_location() -> Path:
+    if "C4D_PYTHON" not in os.environ:
```
joel-wong-aws, line 11: supernit: how about `C4D_PYTHON_EXECUTABLE` to match `C4D_COMMANDLINE_EXECUTABLE`?

src/deadline/cinema4d_submitter/integ_test_helpers.py
joel-wong-aws: Idea: instead of manually removing certain functions from here, can we mock them out instead? That way, if anything else is added to _show_submitter`, it will automatically be included here.

test/integ/test_cinema4d.py
```diff
@@ -0,0 +1,151 @@
...
+        "deadline-cloud-for-cinema-4d"
+    )[0]
+
+    # Get list of files in both directories
+    expected_job_bundle_files = set(
+        f.name for f in expected_job_bundle_dir_path.glob("*") if f.is_file()
+    )
+    generated_job_bundle_files = set(
+        f.name for f in generated_job_bundle_dir_path.glob("*") if f.is_file()
+    )
```
joel-wong-aws, line 84: Should we assert that the size of each of this sets is the same?

joel-wong-aws (approved):

test/integ/conftest.py
karthikbekalp: Sure, I'll make the changes in my upcoming PR.

test/integ/test_cinema4d.py
karthikbekalp: The generated job bundle directory currently also has the scene files that we use to generate the job bundles so the number of files in expected job bundle and generated job bundle will not be equal. 
I can probably put the scene files and attachments in a separate directory. I'll do this in my upcoming PR.

Issue #175, opened 2025-02-12
Bug: g5.xlarge and g6.xlarge instances cannot run Redshift


joel-wong-aws:
### Describe the bug

When rendering on Deadline Cloud Windows SMF using the `cinema4d` Conda package with g5.xlarge or g6.xlarge instances, the following error occurs:
```
2025/02/11 18:03:10-06:00 ADAPTOR_OUTPUT: STDOUT: Rendering frame 150 at <Wed Feb 12 00:03:09 2025>
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: Redshift Error: ======================================================================================================
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: ASSERT FAILED
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: 
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: File C:\builds\maxondev\dev\redshift\core\redshiftcore\SourceCode\BaseLib\Memory\Memory.cpp
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: Line 64
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: 
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: Failed to allocate mem( 1045504 bytes)
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: ======================================================================================================
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: 
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: 
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: Redshift Error: Writing Redshift crash dump to C:\Users\job-user\AppData\Local\Temp\Redshift\20250212-000319-4436-4196.dmp
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: 
2025/02/11 18:03:19-06:00 ADAPTOR_OUTPUT: STDOUT: CRITICAL: Redshift: Core Critical Error Detected [RenderContext.cpp(2515)]
[...]

PR #179, opened 2025-02-25
docs: Add recommendation to use 2x memory for GPU to avoid instability.

karthikbekalp:

We occasionally see Cinema 4D application crash or hang when the memory of the instance is less than 2 times the size of GPU VRAM. 

We have other solutions that we are working on as well. But for now, we update the documentation to notify customers to use larger memory instances. 

I believe 2x the memory size should be sufficient for most of the renders. I've submitted > 8K resolution jobs and confirmed that it worked as required. 

Doc update to just inform customers. 

No tests run as this is a doc update. 

Yes

No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This library requires:
 1. Python 3.9 or higher; but Python 3.11 is recommended as this is the version Cinema 4D uses natively.
 1. Windows or macOS operating system for job submission and Windows operating system for job rendering. There are some instructions below on how to setup the adaptors for job rendering on Linux but they are experimental.
 
-Important: Workers hosts with GPUs that run Cinema 4D adaptor must have sufficient RAM (at least 2x the amount of VRAM) to run Cinema 4D correctly. For example, if your GPU has 16GB VRAM, your system should have at least 32GB RAM. Insufficient memory can lead to unstable rendering behavior.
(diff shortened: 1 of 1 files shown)


crowecawcaw:
Do you have a reference about the relationship between VRAM and RAM requirements? Also is there a specific error message or behavior you noticed that we could mention in this doc update?

karthikbekalp:
> Do you have a reference about the relationship between VRAM and RAM requirements? Also is there a specific error message or behavior you noticed that we could mention in this doc update?

Yup. It is related to this issue. https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/175 
But as part of that issue we will also be detecting the issue and surfacing it to our customers in a better way.

crowecawcaw (dismissed):

README.md
```diff
@@ -29,6 +29,8 @@ This library requires:
 1. Python 3.9 or higher; but Python 3.11 is recommended as this is the version Cinema 4D uses natively.
 1. Windows or macOS operating system for job submission and Windows operating system for job rendering. There are some instructions below on how to setup the adaptors for job rendering on Linux but they are experimental.
 
+Important: Worker hosts that run Cinema 4D adaptor must have sufficient RAM (at least 2x the amount of VRAM) to run Cinema 4D correctly. For example, if your GPU has 16GB VRAM, your system should have at least 32GB RAM. Insufficient memory can lead to unstable rendering behavior.
```
joel-wong-aws: `Worker hosts that run` -> `Workers hosts with GPUs that run the`

Also, can we link to the official C4D docs for this?

karthikbekalp: Makes sense. Updated.

joel-wong-aws (approved):

crowecawcaw (approved):

PR #183, opened 2025-02-28
fix: Replace C4D tokens with values in paths.

### LINKED ISSUE #176: Feature request: Support C4D tokens in output path
### Describe the problem

Currently AWS Deadline Cloud Cinema 4D integration does not support C4D tokens such as `$camera`, etc. in output paths.

### Proposed Solution

Add support for parsing and expanding C4D tokens in the output paths in Cinema 4D adaptor. 

### Example Use Cases

This would allow users to use token-based path structures they use in their local Cinema 4D setups when submitting jobs to the cloud.

karthikbekalp:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/176

Cinema 4D submitter did not replace the C4D tokens in the override output paths and the asset references before submitting. 

This was only observed in override output paths and not during regular submissions. 

This meant that if customers added c4d tokens like $prj within a path for example: C:/my/scene/location/$prj/my_scene then the substitution of the value of "prj" never takes place. 

I investigated how the CInema 4D tokens worked and it looks like the paths were updated to resolve the tokens when receiving from Cinema 4D but the substitution does not occur when submitted using the Override output path. 

I also investigated if we can solve the issue for customers with old submitters as well but considering this would require non-trivial changes to job attachments, we will only be updating for submitters for now. 

The override and the asset references are updated to refer to the path after cinema 4d token substitutions. 

I tested it manually by submitting the jobs.

Here's the parameter values before my change:

```
parameterValues:
...
- name: OutputPath
  value: D:\Users\karthikbekalp\Desktop\BigFreakingScene\$prj_$res_$take\my_scene
- name: MultiPassPath
  value: D:\Users\karthikbekalp\Desktop\BigFreakingScene\$prj_$res_$take
...
```

Here's the parameter values after my change:

```
parameterValues:
...
- name: OutputPath
  value: D:\Users\karthikbekalp\Desktop\BigFreakingScene\test_1280x720_New_Take\my_scene
- name: MultiPassPath
  value: D:\Users\karthikbekalp\Desktop\BigFreakingScene\test_1280x720_New_Take
...
```

Yes

No

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -412,10 +412,12 @@ def create_job_bundle(
     elif settings.take_selection == TakeSelection.CURRENT:
         submit_takes = takes["current_data_list"]
 
-    # Add overrides to asset references
+    # Add overrides to asset references and update the paths with C4D render path tokens.
     if settings.output_path:
+        settings.output_path = Scene.replace_render_path_tokens(settings.output_path)
         asset_references.output_directories.add(os.path.dirname(settings.output_path))
     if settings.multi_pass_path:
+        settings.multi_pass_path = Scene.replace_render_path_tokens(settings.multi_pass_path)
         asset_references.output_directories.add(os.path.dirname(settings.multi_pass_path))
 
     # # Check if there are multiple frame ranges across the takes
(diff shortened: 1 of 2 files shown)


crowecawcaw (approved):
Is there a convenient way to cover this with a test?

karthikbekalp:
> Is there a convenient way to cover this with a test?

Currently no. But I guess I can convert one of the submitter tests to have the C4D tokens in the override paths and check if that fixes it. I'm working on tests right now and I'll keep that in mind while implementing it. :)

joel-wong-aws (approved):

PR #186, opened 2025-03-13
test: Add adaptor integration tests.

karthikbekalp:

We are working on adding integration tests that can automate our release process. 

This CR is one in a series of tests that will help us gain confidence in our code. 

- CR to refactor the submitter code: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/166
- CR to add the submitter tests: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/170
- This CR to add tests that verify submitter + adaptor: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/186
- Upcoming CR: Automate all the tests in job bundle output tests. 

Adds integration tests that devs can run locally. 

I ran `hatch run integ:test` which gives me the following output. 

```
scheduling tests via LoadScheduling

test/integ/test_cinema4d.py::test_integ
[gw0] [100%] PASSED test/integ/test_cinema4d.py::test_integ 

============================================== slowest 5 durations =============================================== 
142.09s call     test/integ/test_cinema4d.py::test_integ
0.00s setup    test/integ/test_cinema4d.py::test_integ
0.00s teardown test/integ/test_cinema4d.py::test_integ
========================================= 1 passed in 143.68s (0:02:23) ==========================================
```

I've updated the instructions required to get the tests working on dev machines as well. 

No, this is just for tests. This should have no customer impact. 

No

----


test/integ/utils.py
```diff
@@ -0,0 +1,191 @@
...
+                    "openjd",
+                    "run",
+                    str(template_path),
+                    "--step",
+                    step["name"],
+                    "--job-param",
+                    json.dumps(job_params),
+                ]
+            )
+
```
karthikbekalp: This is the code that was added in this PR. Other functions were just moved from another file.

DEVELOPMENT.md
```diff
@@ -226,3 +184,59 @@ To run the unit tests, simply use hatch:
...
+         test_cinema4d.py       # Runs all the tests scenes in a parametrized fashion.
+</pre>
+
+You would have to setup Cinema 4D licensing before you run the tests.
+
+1. Set the environment variable `C4D_PYTHON` to the installation folder of Cinema 4D.
+   1. `set C4D_PYTHON=<Cinema 4D location>` on Windows Command or `$env:C4D_PYTHON = <Cinema 4D location>` on Windows Powershell.
+      1. The default location for `Cinema 4D` on Windows is `C:\Program Files\Maxon Cinema 4D 2025\`. This location would be automatically used if the directory exists.
+2. For running the adaptor tests, we would need to install `pywin32` to the installation paths as its an adaptor dependency.
+   2.1 Run `pip install pywin32==308 -t <your Python site-packages location>`.
```
joel-wong-aws, line 239: Can we just include this in the `requirements-testing.txt`?

test/integ/conftest.py
```diff
@@ -8,14 +8,22 @@
 
 @pytest.fixture
 def cinema4d_location() -> Path:
-    if "C4D_PYTHON" not in os.environ:
+    if "C4D_LOCATION" not in os.environ:
```
joel-wong-aws: nit: can we use guard clauses here instead of nested if statements? It makes the logic easier to understand

```
if "C4D_LOCATION" in os.environ:
    return Path(os.environ["C4D_LOCATION"])

print("No C4D_LOCATION set as environment variable, checking in default Cinema 4D location.")
default_path = Path(r"C:\Program Files\Maxon Cinema 4D 2025")
if default_path.exists():
    print("Detected default installation directory for Cinema 4D. Using it.")
    return default_path

raise EnvironmentError(...)
```

test/integ/utils.py
```diff
@@ -0,0 +1,191 @@
...
+    with patch.dict(os.environ, test_env):
+        with open(template_path, encoding="utf-8") as f:
+            template = safe_load(f)
+
+        with open(params_path, encoding="utf-8") as f:
+            parameter_values = safe_load(f)["parameterValues"]
+            job_params = {
+                item["name"]: item["value"]
+                for item in parameter_values
+                if item["name"] in ["Cinema4DFile", "Frames", "MultiPassPath", "OutputPath"]
```
joel-wong-aws: Why just these four job parameters? (if this was intentional, a comment could help)

test/integ/utils.py
```diff
@@ -0,0 +1,191 @@
...
+    run_command(
+        [str(cinema4d_location), str(scene_script_location), str(path_to_create_job_bundle)]
+    )
+
+
+def is_openjd_run_with_cinema4d_successful(
+    cinema4d_location: Path,
+    template_path: Path,
+    params_path: Path,
+) -> bool:
```
joel-wong-aws: nit: instead of returning a bool here which is always True (or raising an `AssertionError` otherwise), can we name this `assert_openjd_run_with_cinema4d_successful` and return `None` instead? Or name it something like `openjd_run_job_and_assert_successful`?

DEVELOPMENT.md
karthikbekalp: Oh interesting idea, I'll test this. But I think it required it to be added into the path where Cinema 4D runs.

test/integ/conftest.py
karthikbekalp: Surely, I'll refactor.

test/integ/utils.py
karthikbekalp: Yup, other than these parameter values others are either queue environments or deadline parameters. I'll add a comment.

test/integ/utils.py
karthikbekalp: Makes sense, I'll refactor.

DEVELOPMENT.md
karthikbekalp: I just tried it out, unfortunately it does not work as it requires that the package be within reach from Cinema 4D's path and it does not seem to consider the hatch's environment.

karthikbekalp: I do have an upcoming PR, I can timebox this and try out if there's a way. :)

test/integ/utils.py
```diff
@@ -0,0 +1,194 @@
...
+
+    with patch.dict(os.environ, test_env):
+        with open(template_path, encoding="utf-8") as f:
+            template = safe_load(f)
+
+        with open(params_path, encoding="utf-8") as f:
+            parameter_values = safe_load(f)["parameterValues"]
+            job_params = {item["name"]: item["value"] for item in parameter_values}
+
+            # Remove the queue Env Parameters
```
baxeaz, line 77: Do we just remove these because they're extra noise?  They're not hurting anything are they?

test/integ/conftest.py
```diff
@@ -8,14 +8,20 @@
...
-            "Please set the C4D_PYTHON environment variable to the location of the Cinema 4D python. "
-            r"The default location for `c4dpy` on Windows is 'C:\Program Files\Maxon Cinema 4D 2025\c4dpy'"
-        )
-
-    return Path(os.environ["C4D_PYTHON"])
+    if "C4D_LOCATION" in os.environ:
+        return Path(os.environ["C4D_LOCATION"])
+
+    print("No C4D_LOCATION set as environment variable, checking in default Cinema 4D location.")
+    default_path = Path(r"C:\Program Files\Maxon Cinema 4D 2025")
```
baxeaz, line 15: Do these essentially follow a pattern year after year?  If it's only once a year maybe it's not a huge burden, but wondering if it might be better to wrap this in some logic so we aren't coming back and updating this all the time.  Or if we'll need to update our images or some other parts of the tests between years anyway maybe it doesn't matter :)

baxeaz (approved):

joel-wong-aws (approved):

test/integ/utils.py
karthikbekalp: Nope, its just the stuff that we add when creating a job to suggest Deadline to download the conda packages, set the max retry count etc.

test/integ/conftest.py
karthikbekalp: > Do these essentially follow a pattern year after year? 

Cinema 4D have followed this starting from 2023 onward IIRC.  

>  if it might be better to wrap this in some logic so we aren't coming back 

That makes sense. Although I'd punt on this for now. :)

PR #189, opened 2025-03-21
chore(installer): Refactor to use individual installers.

karthikbekalp:

We wanted to move fast and hence we are moving forward with building Cinema 4D specific installers. 

We are following the approach similar to other teams like : https://github.com/aws-deadline/deadline-cloud-for-blender/pull/208 and making changes similar to them. 

Faster releases for submitters. 

I ran which build the Windows installers for me that worked well. 

```
hatch run installer:build-installer --install-builder-path 'C:\Program Files\InstallBuilder Enterprise 25.3.0' --platform windows --local-dev
```

Similarly built the installers for MacOS. 

Ran the test installers and confirmed everything passed. 

```
==================================================================================== slowest 5 durations ==================================================================================== 
76.46s setup    test/installer/test_installer.py::TestUserInstall::test_install
66.00s setup    test/installer/test_installer.py::TestWindows::test_user_permissions
62.34s setup    test/installer/test_installer.py::TestUserInstall::test_uninstall
19.37s call     test/installer/test_installer.py::TestUserInstall::test_uninstall
5.19s call     test/installer/test_installer.py::test_default_location
========================================================================= 4 passed, 9 skipped in 162.50s (0:02:42) ==========================================================================
```

No

----


.github/workflows/build_installers.yml
```diff
@@ -0,0 +1,29 @@
...
+    if: (github.repository == 'aws-deadline/deadline-cloud-for-cinema-4d')
+    name: Build Installers
+    uses: aws-deadline/.github/.github/workflows/reusable_build_installers.yml@mainline
+    secrets: inherit
+    with:
+      environment: "mainline"
+      ref_type: ${{inputs.ref_type}}
+      ref: ${{inputs.ref}}
+      oses: "['Windows', 'MacOS']"
+      project_name: "deadline-cloud-for-cinema-4d"
```
joel-wong-aws: supernit: missing newline at the end of this file and various other files in this PR.

Adding newlines at the ends of files helps reduce merge conflicts.

joel-wong-aws (dismissed):

test/installer/test_installer.py
```diff
@@ -0,0 +1,106 @@
...
+    assert python_dir.name in top_level_dir
+    assert "installer_version.txt" in top_level_dir
+    assert uninstaller in top_level_dir
+
+    # Just check that we have dependencies in this folder
+    module_dir = [f.name for f in (python_dir / "modules").iterdir()]
+    assert "deadline" in module_dir
+    assert "qtpy" in module_dir
+
+    # Check the Cinema_4d module is here and there's a version file
```
karthikbekalp: Just found an issue with test_installers. Fixing it now.

.github/workflows/build_installers.yml
karthikbekalp: Makes sense, updated!

test/installer/test_installer.py
karthikbekalp: Fixed!

pipeline/build_installer.sh
viknith, line 1: Noticed you don't have a powershell script for build_installer?

.github/workflows/build_installers.yml
```diff
@@ -0,0 +1,29 @@
+name: "Build Installers"
```
viknith, line 1: Looks different from the yaml reference file here: https://github.com/aws-deadline/deadline-cloud-for-blender/blob/mainline/.github/workflows/build_installers.yml
May want to re-evaluate any changes to make?

.github/workflows/build_installers.yml
```diff
@@ -0,0 +1,29 @@
...
+  BuildInstallers:
+    if: (github.repository == 'aws-deadline/deadline-cloud-for-cinema-4d')
+    name: Build Installers
+    uses: aws-deadline/.github/.github/workflows/reusable_build_installers.yml@mainline
+    secrets: inherit
+    with:
+      environment: "mainline"
+      ref_type: ${{inputs.ref_type}}
+      ref: ${{inputs.ref}}
+      oses: "['Windows', 'MacOS']"
```
viknith: Noob question, isn't there C4D on Linux? https://www.maxon.net/en/downloads/cinema-4d-2024-downloads

viknith: Unless these are specifying the OS that builds the installer, not running it?

karthikbekalp: There is C4D for Linux but its a non-GUI version for worker nodes. So customers can render using Linux but they don't have the ability to open C4D scene files to edit in Linux which means we don't need a submitter for Linux.

.github/workflows/build_installers.yml
karthikbekalp: Hmm, I think the yml file is identical to the one in blender no? Its just that the environment and OSes are directly defined in the job instead of in the inputs... 

I was following Keyshot's example here: https://github.com/aws-deadline/deadline-cloud-for-keyshot/pull/190/files as it is very similar to C4D i.e. both are only for Windows and MacOS.

pipeline/build_installer.sh
karthikbekalp: My bad, will add. Thanks!

.github/workflows/build_installers.yml
karthikbekalp: Anyway, updated the file based on the new yml file from Blender.

pipeline/build_installer.sh
karthikbekalp: Added in the most recent commit in this PR.

install_builder/deadline-cloud-for-cinema-4d.xml
```diff
@@ -68,49 +68,114 @@
...
+                    <origin>${installdir}/tmp/c4d_deps/DeadlineCloud.pyp</origin>
+                </copyFile>
+                <fnAddPathEnvironmentVariable>
+                    <progressText>Setting C4DPYTHONPATH311</progressText>
+                    <name>C4DPYTHONPATH311</name>
+                    <value>${cinema_4d_submitter_installdir}</value>
+                    <scope>${installscope}</scope>
+                    <insertAt>end</insertAt>
+                </fnAddPathEnvironmentVariable>
+                <fnAddPathEnvironmentVariable>
```
viknith, line 146: nit: Should the level that this path variable get set in Windows depend on whether it's a System install or a User install? IIRC there are two path variables, one for User and one for System. Just something to check, maybe it doesn't matter

pipeline/build_installer.ps1
```diff
@@ -0,0 +1,3 @@
+$ErrorActionPreference = "Stop"
```
viknith, line 2: Still waiting on confirmation for whether this is missing copyright headers here. I included them just in case though.

viknith (dismissed):

install_builder/deadline-cloud-for-cinema-4d.xml
karthikbekalp: Yes, that's intentional. When a customer uses system installation, the environment variables are set in the system variables. This is useful when an administrator wants to grant multiple users within a system access to the application.

pipeline/build_installer.ps1
karthikbekalp: Nice catch, I'll update it. right away.

viknith (approved):

joel-wong-aws (approved):

PR #199, opened 2025-04-09
chore(dev): fix dev installer scripts, and determine version based off tag

epmog:

Builds were failing:

```
Usage: build_installer_cli.py [OPTIONS]
Try 'build_installer_cli.py --help' for help.
Error: Invalid value for '--install-builder-license-path': Must specify --install_builder_license_path when none of --dev, --local-dev are specified
```
Was missing the change to return when we have the value (and to pretty up the error message from snake to kebab)

pull the changes from Blender, also included the new generation of the installed version file to be the tag rather than the date. I also removed the quotes around the variable expansion, since that's what Blender has and that works. But I don't believe that's necessary

built locally, and ran the installer locally. Confirmed version is the tag plus hash of additional commits.

N/A
N/A

----

### DIFF
--- a/pipeline/build_installer.sh
+++ b/pipeline/build_installer.sh
@@ -4,4 +4,4 @@
 # Set the -e option
 set -e
 
-hatch run installer:build-installer "$@"
+hatch run installer:build-installer $@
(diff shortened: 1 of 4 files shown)


karthikbekalp (approved):
Yup, I had the exact same changes!

joel-wong-aws (approved):

PR #207, opened 2025-04-30
fix: Partial fix for failures with RS RenderView when opening Cinema 4D submitter

karthikbekalp:

The AWS Deadline Cloud submitter for Cinema 4D caused crashes while opening RS RenderView. 

Screenshot 2025-04-23 at 4 15 40 PM

Screenshot 2025-04-23 at 4 14 41 PM

We believe the issue was that Redshift's RenderView was using the submitter Qt modules which caused failures. 

This issue only happens if the PyQt modules are loaded at the beginning. Hence, this solution moves the loading of the modules to after the user clicks the submitter. 

Huge kudos to @ryyakobe  and @epmog  for helping out here! 

This is a partial fix because if the submitter is called before the RS render view this would still be an issue. We are currently in contact with Maxon support for the cleaner solution to this problem.

Redshift's RenderView will start working with our submitter installed. 

Tested this on mac and confirmed that it works! 

Also tested it on Windows and confirmed it does not cause any side effects. 

Yes

Yes. 

Yes, but this is in the plugin which does not change our tests. 

No. 

No. 

----

### DIFF
--- a/deadline_cloud_extension/DeadlineCloud.pyp
+++ b/deadline_cloud_extension/DeadlineCloud.pyp
@@ -23,18 +23,9 @@ if sys.platform in ["win32", "cygwin"]:
     # Required due to a longstanding bug in C4D to not include the python dll library path required by many compiled libs
     #   Read more here: https://github.com/danbradham/wheels/issues/4#issuecomment-1772721170
     os.add_dll_directory(dll_path)
-elif sys.platform == "darwin": # If MacOS
-    # This path will be replaced when the installer is running
-    sys.path.append("C4D_Submitter_Installation_Dir_To_Replace")
 
 root = os.path.dirname(__file__)
 sys.path.insert(0, os.path.join(root, 'modules'))
-
-if 'deadline.cinema4d_submitter' not in sys.modules.keys():
-    import deadline.cinema4d_submitter
-else:
-    import importlib
-    importlib.reload(deadline.cinema4d_submitter)
 
 import c4d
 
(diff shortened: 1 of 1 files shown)


deadline_cloud_extension/DeadlineCloud.pyp
```diff
@@ -44,8 +35,24 @@ PLUGIN_ID = 1064358
 
 class DeadlineCloudRenderCommand(c4d.plugins.CommandData):
 
-    def Execute(self, doc):
+    def _import_and_show_submitter(self):
+        import deadline.cinema4d_submitter
```
joel-wong-aws, line 39: With this change, the logic
```
if 'deadline.cinema4d_submitter' not in sys.modules.keys():
else:
    import importlib
    importlib.reload(deadline.cinema4d_submitter)
```

is changed to simply 
```
```

Was the `else` condition previously used? Where did it come from?

karthikbekalp: Nice eye. 

I don't think the `importlib.reload` was ever being used before. Just to verify I added print statements and tried opening and reopening the submitter multiple times and nothing got printed. 

The docs mention to [use reload the modules](https://developers.maxon.net/docs/py/2024_0_0a/misc/pluginstructure.html#python-plugins-reloading) if the function PluginMessage was being used and we are listening to specifically C4DPL_RELOADPYTHONPLUGIN messages from Cinema 4D to reload some resources like sockets. 

Right now, we don't use that function so I believe we can remove it.

joel-wong-aws (approved):

deadline_cloud_extension/DeadlineCloud.pyp
```diff
@@ -44,8 +35,24 @@ PLUGIN_ID = 1064358
...
         deadline.cinema4d_submitter.show_submitter()
+    
+    def _execute_for_mac(self):
+        # This variable will be replaced to a path when the installer is running
+        C4D_MAC_INSTALLATION_PATH = "C4D_Submitter_Installation_Dir_To_Replace"
```
crowecawcaw, line 44: It feels brittle to hardcoded this directory into the script, but maybe it's unavoidable. Do the manual install steps still work?

crowecawcaw (approved):

deadline_cloud_extension/DeadlineCloud.pyp
karthikbekalp: >  Do the manual install steps still work?

Yes. If the path actually does not exist, it is then skipped so manual steps should still work.

PR #228, opened 2025-06-05
ci: update release workflow

moorec-aws:
There are changes to the [re-usable workflows](https://github.com/aws-deadline/.github/pull/39).

1. Updated `code_quality.yml` to accept tags as inputs, and moved the Unit Tests from the `release_bump.yml` to the `release_publish.yml` workflow.
2. The `release_publish.yml` workflow automatically starts after a ChangeLog PR has been merged. We used the new `reusable_tag_release.yml` workflow to create a new tag. We validate the tag has not been released and the author of the associated commit. The workflow can be re-run, specifying an existing tag.
3. Building and Staging the installer has been moved to its own workflow.

More modular workflows, added automation.

Workflows were tested using Blender Repository:https://github.com/moorec-aws/deadline-cloud-for-blender/actions/runs/15218951005

No

No

----

### DIFF
--- a/.github/workflows/release_bump.yml
+++ b/.github/workflows/release_bump.yml
@@ -17,15 +17,8 @@ concurrency:
   group: release
 
 jobs:
-  UnitTests:
-    name: Unit Tests
-    uses: ./.github/workflows/code_quality.yml
-    with:
-      branch: mainline
-
   Bump:
     name: Version Bump
-    needs: UnitTests
     uses: aws-deadline/.github/.github/workflows/reusable_bump.yml@mainline
     secrets: inherit
     with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release_bump.yml
```diff
@@ -17,15 +17,8 @@ concurrency:
   group: release
 
 jobs:
-  UnitTests:
```
karthikbekalp, line 20: Hmm, any particular reason why we removed the unit tests job? Is it because we now do it in release_publish instead?

moorec-aws: Yes. I moved it because it is a bit redundant. When the ChangeLog PR is pushed, it will run the tests as part of the PRs validation checks. Then we now have it in the release_publish because we are testing the build from the tag checkout.

karthikbekalp: Cool. Makes sense.

karthikbekalp (approved):

joel-wong-aws (approved):

PR #231, opened 2025-06-11
feat: Remember last used output paths and take selection for job submissions

gianelli-lagos-aws:
Fixes: *< (https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/95) >*

The fields output_path, multi_pass_path, and take_selection in the submitter's job specific settings do not have sticky settings which can be inconvenient. The fields override_output_path and override_multi_pass_path do not exist and they would be useful in order to save checkbox checked states. 

Add sticky settings to output_path, multi_pass_output, and take_selection. 
Create two new boolean fields, override_output_path and override_multi_pass_path, with sticky metadata to track checkbox states. For override_output_path and override_multi_pass_path, if the checkbox is unchecked the default render settings paths will be used for the job. 

Customers will have settings that are saved and reused between submissions.

* Built the package with the changes and verified that settings persist after closing and reopening Cinema 4D.
* Tested with existing settings files to ensure backward compatibility is maintained.
* Examined the JSON settings file to confirm that the new fields are being properly saved.
* Tested edge cases such as enabling/disabling checkboxes and changing paths multiple times.
* Ran unit tests:
  * 0.02s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[Project not found-True]
============================================================ 53 passed in 4.58s ============================================================ 
* Ran integration tests:
  * =========================================================== slowest 5 durations ============================================================ 
109.25s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
82.20s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
78.92s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
75.83s call     test/integ/test_cinema4d.py::test_integ[redshift]
54.39s call     test/integ/test_cinema4d.py::test_integ[physical]
====================================================== 6 passed in 456.20s (0:07:36) =======================================================

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -415,12 +415,24 @@ def create_job_bundle(
         submit_takes = takes["current_data_list"]
 
     # Add overrides to asset references and update the paths with C4D render path tokens.
-    if settings.output_path:
-        settings.output_path = Scene.replace_render_path_tokens(settings.output_path)
-        asset_references.output_directories.add(os.path.dirname(settings.output_path))
-    if settings.multi_pass_path:
-        settings.multi_pass_path = Scene.replace_render_path_tokens(settings.multi_pass_path)
-        asset_references.output_directories.add(os.path.dirname(settings.multi_pass_path))
+    scene_output_path, scene_multi_pass_path = Scene.get_output_paths()
+    if settings.override_output_path:
+        if settings.output_path:
+            settings.output_path = Scene.replace_render_path_tokens(settings.output_path)
+            asset_references.output_directories.add(os.path.dirname(settings.output_path))
+    else:
+        if scene_output_path:
+            settings.output_path = Scene.replace_render_path_tokens(scene_output_path)
+            asset_references.output_directories.add(os.path.dirname(scene_output_path))
+
+    if settings.override_multi_pass_path:
+        if settings.multi_pass_path:
+            settings.multi_pass_path = Scene.replace_render_path_tokens(settings.multi_pass_path)
+            asset_references.output_directories.add(os.path.dirname(settings.multi_pass_path))
+    else:
+        if scene_multi_pass_path:
+            settings.multi_pass_path = Scene.replace_render_path_tokens(scene_multi_pass_path)
+            asset_references.output_directories.add(os.path.dirname(scene_multi_pass_path))
 
     # # Check if there are multiple frame ranges across the takes
     first_frame_range = submit_takes[0].frame_range
--- a/src/deadline/cinema4d_submitter/data_classes.py
+++ b/src/deadline/cinema4d_submitter/data_classes.py
@@ -113,7 +113,12 @@ class RenderSubmitterUISettings:
                 obj = {
                     field.name: getattr(self, field.name)
                     for field in dataclasses.fields(self)
-                    if field.metadata.get("sticky") and field.name != "timeouts"
+                    if field.metadata.get("sticky")
+                    and field.name != "timeouts"
+                    and not (
+                        (field.name == "output_path" and not self.override_output_path)
+                        or (field.name == "multi_pass_path" and not self.override_multi_pass_path)
+                    )
                 }
                 obj["timeouts"] = self.timeouts.to_sticky_settings_dict()
                 json.dump(obj, fh, indent=1)
--- a/src/deadline/cinema4d_submitter/data_classes.py
+++ b/src/deadline/cinema4d_submitter/data_classes.py
@@ -51,15 +51,17 @@ class RenderSubmitterUISettings:
     description: str = field(default="", metadata={"sticky": True})
 
     override_frame_range: bool = field(default=False, metadata={"sticky": True})
+    override_output_path: bool = field(default=False, metadata={"sticky": True})
+    override_multi_pass_path: bool = field(default=False, metadata={"sticky": True})
     frame_list: str = field(default="", metadata={"sticky": True})
-    output_path: str = field(default="")
-    multi_pass_path: str = field(default="")
+    output_path: str = field(default="", metadata={"sticky": True})
+    multi_pass_path: str = field(default="", metadata={"sticky": True})
 
     input_filenames: list[str] = field(default_factory=list, metadata={"sticky": True})
     input_directories: list[str] = field(default_factory=list, metadata={"sticky": True})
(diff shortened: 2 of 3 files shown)


src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -87,6 +89,8 @@ def load_sticky_settings(self, scene_filename: str):
                         if name in sticky_fields:
                             if name == "timeouts":
                                 self.timeouts.update_from_sticky_settings(value)
+                            elif name == "take_selection":
```
karthikbekalp, line 93: Nit: A comment explaining why "take_selection" requires a separate check would be helpful.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -415,12 +415,20 @@ def create_job_bundle(
         submit_takes = takes["current_data_list"]
 
     # Add overrides to asset references and update the paths with C4D render path tokens.
-    if settings.output_path:
+    default_output_path, default_multi_pass_path = Scene.get_output_paths()
```
karthikbekalp: Nit: Instead of `default_output_path` can we name it as `scene_output_path` to emphasize the source of the path?

karthikbekalp (commented):
Thanks Gianelli. The PR looks great albeit some nit comments. 

Can you also check with different render tokens and if those work as expected too? https://help.maxon.net/c4d/en-us/#html/54506.html

karthikbekalp:
I think the commit message could be improved.

Instead of:
`feat: add sticky settings to job specific fields for cinema4d submitter`

We could use:
`feat: Remember last used output paths and take selection for job submissions`

This would be more helpful for our customers when they read the release notes to decide whether to upgrade to a newer release.

karthikbekalp (dismissed):
LGTM. Great job!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Yes, I think that `scene_output_path` is better. Thank you!

src/deadline/cinema4d_submitter/data_classes.py
gianelli-lagos-aws: I agree. Thank you!

gianelli-lagos-aws: I agree. Thank you!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Yes, i think `scene_output_path` is better. Thank you!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -415,12 +415,24 @@ def create_job_bundle(
...
 
     # Add overrides to asset references and update the paths with C4D render path tokens.
-    if settings.output_path:
-        settings.output_path = Scene.replace_render_path_tokens(settings.output_path)
-        asset_references.output_directories.add(os.path.dirname(settings.output_path))
-    if settings.multi_pass_path:
-        settings.multi_pass_path = Scene.replace_render_path_tokens(settings.multi_pass_path)
-        asset_references.output_directories.add(os.path.dirname(settings.multi_pass_path))
+    scene_output_path, scene_multi_pass_path = Scene.get_output_paths()
+    if settings.override_output_path:
```
joel-wong-aws, line 419: I think there's one case that this approach misses.

Imagine I set my output path via Cinema 4D (Render > Render Settings > Save) to C:\A\Path. I then open the submitter, check that my job settings show an output path of `C:\A\Path` (no path override) and submit my job.

The second time I submit, I change my output path to `C:\B\Path` in Cinema 4D (Render > Render Settings > Save). When I open the submitter the second time, my job settings will still show an output path of `C:\A\Path` with no path override, even though it will actually submit with an output path of `C:\B\Path`.

I think we need to handle this case to avoid customer confusion. There are two ways we can do this:
(1) If override output path is not set, then do not load the output path from the sticky settings file
(2) If override output path is not set, then do not save the output path to the sticky settings file

I think either approach could be okay. (1) saves the correct output path to the sticky settings file but doesn't use it, while (2) leaves an empty/missing output path

gianelli-lagos-aws: Hello, I can see how this approach misses! I think I will go for the second option since if override output path is false then it best for the output path to not be in the json file. Thank you!

src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -113,7 +113,12 @@ def save_sticky_settings(self, scene_filename: str):
                 obj = {
                     field.name: getattr(self, field.name)
                     for field in dataclasses.fields(self)
-                    if field.metadata.get("sticky") and field.name != "timeouts"
+                    if field.metadata.get("sticky")
+                    and field.name != "timeouts"
+                    and not (
+                        (field.name == "output_path" and not self.override_output_path)
+                        or (field.name == "multi_pass_path" and not self.override_multi_pass_path)
+                    )
```
karthikbekalp: nit: I think we can improve the readability of the code here:

```
             with open(sticky_settings_filename, "w", encoding="utf8") as fh:
                obj = {
                    field.name: getattr(self, field.name)
                    for field in dataclasses.fields(self)
                    if field.metadata.get("sticky")
                    and field.name not in ["timeouts", "output_path", "multi_pass_path"]
                }
                obj["timeouts"] = self.timeouts.to_sticky_settings_dict()
                if self.override_output_path:
                    obj["output_path"] = self.output_path
                if self.override_multi_pass_path:
                    obj["multi_pass_path"] = self.multi_pass_path
                json.dump(obj, fh, indent=1)
```

karthikbekalp (commented):
Added one comment to improve readability, but other than that. LGTM.

src/deadline/cinema4d_submitter/data_classes.py
gianelli-lagos-aws: I agree, this is much clearer! I will work on that, thank you.

karthikbekalp (approved):

joel-wong-aws (approved):

PR #180, opened 2025-02-25
feat: Add configurable GUI timeouts for jobs in submitter.

karthikbekalp:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/174

### What was the problem/requirement? (What/Why)
Cinema 4D randomly gets stuck while running and this can lead to customer overspend. 

### What was the solution? (How)
We are adding timeouts to Deadline Cloud library with https://github.com/aws-deadline/deadline-cloud/pull/611 

We will be using that add timeouts to Cinema 4D submitter. 

### What is the impact of this change?
Customers can now configure timeouts in the submitter. 

### How was this change tested?

- Have you run the unit tests?
Yes

### Was this change documented?
Yes

### Is this a breaking change?

No

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #235, opened 2025-06-19
feat: Remember last used shared job settings for job submissions

gianelli-lagos-aws:
Fixes: *< https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/95 >*

The fields priority, initial_state , max_failed_tasks_count, max_retries_per_task, and max_worker_count do not have sticky settings which can be inconvenient for the customer. 

* Update the refresh_ui() method to use getattr() with fallback values for backward compatibility.
* Modify the method to load saved values for all four fields when they exist in the settings object.
* Update the update_settings() method to save all four field values when they exist in the settings object.

Customers will have settings that are saved and reused between submissions.

* Since there were also changes made to the deadline-cloud-for-cinema-4d repo, both packages were built with the changes and tested together to ensure proper integration.
* Verified that all settings persist after closing and reopening Cinema 4D.
* Tested with both new scenes and existing scenes to ensure backward compatibility.
* Validated that the JSON settings file contains all the expected fields with correct values.

Yes. 
0.05s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[Project not found-True]
0.04s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[CRITICAL: Stop [ge_file.cpp(1172)]-False]
============================================================================== 53 passed in 3.90s ============================================================

Yes. 
94.44s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
77.32s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
74.20s call     test/integ/test_cinema4d.py::test_integ[redshift]
73.27s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
53.66s call     test/integ/test_cinema4d.py::test_integ[physical_textured]
======================================================================== 6 passed in 427.19s (0:07:07) ============================

----


src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -50,6 +50,13 @@ class RenderSubmitterUISettings:
     name: str = field(default="", metadata={"sticky": True})
     description: str = field(default="", metadata={"sticky": True})
 
+    priority: int = field(default=50, metadata={"sticky": True})
+    initial_state: str = field(default="READY", metadata={"sticky": True})
+    max_failed_tasks_count: int = field(default=20, metadata={"sticky": True})
+    max_retries_per_task: int = field(default=5, metadata={"sticky": True})
+    limited_max_worker_count: bool = field(default=False, metadata={"sticky": True})
+    max_worker_count: int = field(default=1, metadata={"sticky": True})
```
karthikbekalp: I believe here the default should be -1? Also can we add a comment suggesting that it is -1 when there is no max worker count limit?

gianelli-lagos-aws: Yes! I updated it and added a comment. Thank you!

src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -50,6 +50,15 @@ class RenderSubmitterUISettings:
     name: str = field(default="", metadata={"sticky": True})
     description: str = field(default="", metadata={"sticky": True})
 
+    priority: int = field(default=50, metadata={"sticky": True})
+    initial_state: str = field(default="READY", metadata={"sticky": True})
+    max_failed_tasks_count: int = field(default=20, metadata={"sticky": True})
+    max_retries_per_task: int = field(default=5, metadata={"sticky": True})
+    limited_max_worker_count: bool = field(default=False, metadata={"sticky": True})
```
karthikbekalp: Hmm, do we use this field anywhere now? Considering we only check the max_worker_count in https://github.com/aws-deadline/deadline-cloud/pull/711/files?

karthikbekalp: Thanks for removing the field.

src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -50,6 +50,14 @@ class RenderSubmitterUISettings:
     name: str = field(default="", metadata={"sticky": True})
     description: str = field(default="", metadata={"sticky": True})
 
+    priority: int = field(default=50, metadata={"sticky": True})
+    initial_state: str = field(default="READY", metadata={"sticky": True})
+    max_failed_tasks_count: int = field(default=20, metadata={"sticky": True})
+    max_retries_per_task: int = field(default=5, metadata={"sticky": True})
+    max_worker_count: int = field(
+        default=-1, metadata={"sticky": True}
+    )  # -1 indicates no limited max worker count
```
joel-wong-aws: supernit: `no limited` -> `unlimited`

src/deadline/cinema4d_submitter/data_classes.py
```diff
@@ -50,6 +50,14 @@ class RenderSubmitterUISettings:
     name: str = field(default="", metadata={"sticky": True})
     description: str = field(default="", metadata={"sticky": True})
 
+    priority: int = field(default=50, metadata={"sticky": True})
```
joel-wong-aws, line 54: Out of scope for this PR, but nice to have: these fields are shared with all submitters. Can we add a data class and shared functions in `deadline-cloud` so that this can be shared with other integrations?

joel-wong-aws (dismissed):

karthikbekalp (changes requested):
The changes look good to me. 

This PR requires https://github.com/aws-deadline/deadline-cloud/pull/711/files from deadline cloud to work. Although, adding this should not break the code it would be better if we could merge this once the deadline cloud changes are merged and released. 

For example, something like this: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/180/files where the `pyproject.toml` needed some updates so that the feature could work.

src/deadline/cinema4d_submitter/data_classes.py
karthikbekalp: I think its out of scope for this PR and the effort may be non-trivial. But I'll let @gianelli-lagos-aws  respond. :)

karthikbekalp (approved):
As the changes are now in as part of: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/266 we should be able to merge this PR in.

joel-wong-aws (approved):

PR #237, opened 2025-06-25
fix: support all takes flow for multi-take/render setting/frame range jobs

joel-wong-aws:
Jobs submitted with all takes that had multiple takes + render settings + frame ranges were failing to submit.

There were a couple issues with this flow:
- The "frame" parameter was not being handled correctly (it's unused in the multi-take flow)
- Long take names >64 characters were not being handled correctly
- Take names with special characters were not being handled correctly

Unblocks customer to use more complicated multi-take flows.

Tested manually with a multiple takes + render settings + frame range scene, as well as added an integration test for it.

Yes

I ran all the physical based render tests

Yes

Yes, via in-code docstrings/comments where applicable.

No

----


leon-li-inspire (dismissed):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -465,6 +481,57 @@ def create_job_bundle(
...
+        parameter_name = removed_job_parameter_chars.sub("", take_data.display_name)[
+            : 64 - len("Frames")
+        ]
+        # ensure the first character is allowed or prefix with an _
+        if not allowed_first_job_parameter_chars.match(parameter_name):
+            parameter_name = f"_{parameter_name}"[: 64 - len("Frames")]
+        # ensure all parameter names are unique
+        if parameter_name in parameter_names:
+            # example: NewTake_00001
+            parameter_name = f"{parameter_name[:64 - len('Frames') - 6]}_{take_number:05}"
```
crowecawcaw, line 528: Do we need to check if the new parameter name is unique?

joel-wong-aws: Added simple error handling for this case with message `f"Unable to generate unique parameter name for take '{take_name}', please change the take name."`! I don't think it's worth focusing too much on this case, this will only occur if someone names has at least 3 takes named like:
`take_1`
`take_1_00001_00003`
`take_1_00001`
which seems unlikely, and the error message seems clear enough now for them to resolve the issue.

crowecawcaw (approved):

leon-li-inspire (approved):

PR #240, opened 2025-06-30
docs: Updated the requirements to use NVIDIA drivers > 551.78 or later.

karthikbekalp:
We found that Redshift couldn't be rendered on Nvidia GPUs on driver version of 535. 

Turns out Redshift requires driver version 551.78 or later. This PR just updates the documentation for this. 

Better documentation for customers. 

Only doc change. 

Yes

No, only a doc change. 
----

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,9 @@ This library requires:
 
 1. Cinema 4D 2024 - 2025
    * Redshift is supported but not required
-1. Python 3.9 or higher; but Python 3.11 is recommended as this is the version Cinema 4D uses natively.
-1. Windows or macOS operating system for job submission and Windows operating system for job rendering. There are some instructions below on how to setup the adaptors for job rendering on Linux but they are experimental.
+2. Python 3.9 or higher; but Python 3.11 is recommended as this is the version Cinema 4D uses natively.
+3. Windows or macOS operating system for job submission and Windows operating system for job rendering. There are some instructions below on how to setup the adaptors for job rendering on Linux but they are experimental.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,8 +26,9 @@ This library requires:
 
 1. Cinema 4D 2024 - 2025
    * Redshift is supported but not required
-1. Python 3.9 or higher; but Python 3.11 is recommended as this is the version Cinema 4D uses natively.
-1. Windows or macOS operating system for job submission and Windows operating system for job rendering. There are some instructions below on how to setup the adaptors for job rendering on Linux but they are experimental.
+2. Python 3.9 or higher; but Python 3.11 is recommended as this is the version Cinema 4D uses natively.
+3. Windows or macOS operating system for job submission and Windows operating system for job rendering. There are some instructions below on how to setup the adaptors for job rendering on Linux but they are experimental.
+4. When rendering using Redshift with Cinema 4D on NVIDIA GPUs, driver version 551.78 or later is required. Using older drivers can result in job failures.
```
crowecawcaw: Do we need to specify what type of NVIDA drivers this is? Is it GRID drivers?

crowecawcaw (changes requested):

README.md
karthikbekalp: That's a good call. Updated the readme.

crowecawcaw (approved):

crowecawcaw (approved):

joel-wong-aws (approved):

PR #244, opened 2025-07-07
feat!: Allow deactivating automatic error checking in Cinema4D jobs

gianelli-lagos-aws:
Fixes: *<https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/193>*

This PR is the 1st of a set of PRs. The other 2 PRs are the following: 
2nd: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/245
3rd: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/246

Key changes this PR brings:
* Add a checkbox labeled "Activate automatic error checking" in the job specific settings of the submitter.
* Set the default value to True (error checking activated) for backward compatibility.
* Update the _configure_settings() method to initialize the checkbox state from settings.
* Update the update_settings() method to save the checkbox state to settings.

### What was the problem/requirement? (What/Why)
Implement a solution to deactivate automatic error checking during rendering to provide users with more control over the rendering process.

### What was the solution? (How)
Add a checkbox that allows user to activate and deactivate error checking.
 
The checkbox is shown in the image below highlighted by the red box!
Activate_Error_Checking_checkbox

### What is the impact of this change?
Customers will be able to have more control over the rendering process. 

### How was this change tested?
* Tested rendering workflows with error checking activated (default behavior).
* Tested rendering workflows with error checking deactivated.
* Verified that error messages are properly caught when error checking is activated.
* Verified that error messages are ignored when error checking is deactivate.
* Tested backward compatibility with existing jobs that don't have this setting.

- Have you run the unit tests?
Yes.
0.03s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::test_adaptor_rejects_malformed_init_data
0.03s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[Project not found-True]
======================================================================= 56 passed in 4.08s ======================================

- Have you run the integration tests?
Yes.
432.28s call     test/integ/test_cinema4d.py::test_integ[physical_multi_takes]
88.41s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
77.87s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
75.99s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
[...]

Issue #193, opened 2025-03-28
Feature request: Add an option to deactivate automatic error checking in logs.


karthikbekalp:
### Describe the problem

In the [adaptor code](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/f881c982f7cd9dec8538da6e95e6a34cec739349/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py#L247-L268), we check for multiple regex patterns and fail the job if the worker logs contain any of these patterns.

However, this error checking may not always be necessary, and customers should have the option to disable it if they wish. Please provide a way to implement this functionality.

### Proposed Solution

There is a similar feature in Houdini. However, unlike Houdini, Cinema 4D can produce "critical" in their logs that may or may not cause job failures. Therefore, we should exercise caution when implementing this feature.

### Example Use Cases

This [related PR](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/192/files) addresses the issue where customers' jobs failed because we included "Critical Stop" as one of the error regex patterns.

If customers had the ability to disable this check, they would have been able to submit their jobs successfully.

joel-wong-aws:
This feature is available in release [0.8.0](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/releases/tag/0.8.0)!

PR #245, opened 2025-07-07
feat!: Allow deactivating error checking in the Cinema4D adaptor

gianelli-lagos-aws:
Fixes: *<https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/193>*

This PR is the 2nd of a set of PRs. The other 2 PRs are the following: 
1st: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/244
3rd: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/246

Key changes this PR brings:
Since a checkbox that allows user to deactivate error checking was added in the first PR, we also need to make changes to the adaptor:
* Add a class variable _activate_error_checking to store the setting.
* Update the on_start() method to read this setting from init_data.
* Modify the _get_regex_callbacks() method to conditionally include error regexes based on this setting.
* Add a unit test

Implement a solution to deactivate automatic error checking during rendering to provide users with more control over the rendering process.

Add a checkbox that allows user to activate and deactivate error checking.
 
The checkbox is shown in the image below highlighted by the red box!
Activate_Error_Checking_checkbox

Customers will be able to have more control over the rendering process. 

* Tested rendering workflows with error checking activated (default behavior).
* Tested rendering workflows with error checking deactivated.
* Verified that error messages are properly caught when error checking is activated.
* Verified that error messages are ignored when error checking is deactivate.
* Tested backward compatibility with existing jobs that don't have this setting.

Yes.
0.03s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::test_adaptor_rejects_malformed_init_data
0.03s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[Project not found-True]
======================================================================= 56 passed in 4.08s ======================================

Yes.
432.28s call     test/integ/test_cinema4d.py::test_integ[physical_multi_takes]
88.41s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
77.87s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
75.99s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
74.46s call     test/integ/test_cinema4d.py::test_integ[redshift]
================================================================== 7 passed in 857.94s (0:14:17) ================================

Yes, I added a checkbox that allows customers to deactivate automatic error checking.

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -73,6 +73,9 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
     # Will be optionally changed after the scene is set.
     _expected_outputs: int = 1  # Total number of renders to perform.
     _produced_outputs: int = 0  # Counter for tracking number of complete renders.
+    _deactivate_error_checking: int = (
+        0  # 0=activate, 1=deactivate - controls whether error regex callbacks are added
+    )
 
     def _print_adaptor_version(self) -> None:
         """Prints the adaptor version information."""
@@ -107,7 +110,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
-        return SemanticVersion(major=0, minor=1)
+        return SemanticVersion(major=0, minor=2)
 
     @staticmethod
     def _get_timer(timeout: int | float) -> Callable[[], bool]:
@@ -275,7 +278,13 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
                 re.compile(r".*Rendering was internally aborted.*", re.IGNORECASE),
                 re.compile(r'.*Cannot find procedure "rsPreference".*', re.IGNORECASE),
             ]
-            callback_list.append(RegexCallback(error_regexes, self._handle_error))
+
+            # Only add error regexes if error checking is not deactivated
+            if not self._deactivate_error_checking:
+                _logger.warning("Adding error regexes to callback list")
+                callback_list.append(RegexCallback(error_regexes, self._handle_error))
+            else:
+                _logger.warning("NOT adding error regexes to callback list")
 
             insufficient_ram_regexes = re.compile(r".*Failed to allocate mem.*", re.IGNORECASE)
             callback_list.append(
@@ -455,6 +464,8 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
         """
         self.validators.init_data.validate(self.init_data)
 
+        self._deactivate_error_checking = int(self.init_data.get("deactivate_error_checking", "0"))
+
         self.update_status(progress=0, status_message="Initializing Cinema4D")
         self._initialize_maxon_assets_db_connection()
         self._start_cinema4d_server_thread()
@@ -481,7 +492,6 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
         This starts a render in Cinema4D for the given frame, scene and layer(s) and
         performs a busy wait until the render completes.
         """
-
         self.validators.run_data.validate(run_data)
 
         if not self._cinema4d_is_running:
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -279,7 +279,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
                 re.compile(r'.*Cannot find procedure "rsPreference".*', re.IGNORECASE),
             ]
 
-            # Only add error regexes if error checking is not deactivated
+            # Only add error regexes if error checking is activated
             if self._activate_error_checking:
                 _logger.warning("Adding error regexes to callback list")
                 callback_list.append(RegexCallback(error_regexes, self._handle_error))
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -281,7 +281,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
 
             # Only add error regexes if error checking is activated
             if self._activate_error_checking:
-                _logger.warning("Adding error regexes to callback list")
+                _logger.info("Adding error regexes to callback list")
                 callback_list.append(RegexCallback(error_regexes, self._handle_error))
             else:
                 _logger.warning("NOT adding error regexes to callback list")
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -73,8 +73,8 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
     # Will be optionally changed after the scene is set.
     _expected_outputs: int = 1  # Total number of renders to perform.
     _produced_outputs: int = 0  # Counter for tracking number of complete renders.
-    _deactivate_error_checking: int = (
-        0  # 0=activate, 1=deactivate - controls whether error regex callbacks are added
+    _activate_error_checking: int = (
+        1  # 0=deactivate, 1=activate - controls whether error regex callbacks are added
     )
 
     def _print_adaptor_version(self) -> None:
@@ -280,7 +280,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
             ]
 
             # Only add error regexes if error checking is not deactivated
-            if not self._deactivate_error_checking:
+            if self._activate_error_checking:
                 _logger.warning("Adding error regexes to callback list")
                 callback_list.append(RegexCallback(error_regexes, self._handle_error))
             else:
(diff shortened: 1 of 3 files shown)


src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -275,7 +278,13 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
                 re.compile(r".*Rendering was internally aborted.*", re.IGNORECASE),
                 re.compile(r'.*Cannot find procedure "rsPreference".*', re.IGNORECASE),
             ]
-            callback_list.append(RegexCallback(error_regexes, self._handle_error))
+
+            # Only add error regexes if error checking is not deactivated
```
karthikbekalp: There is a double negative here. Can we rename this variable to `_activate_error_checking` to simplify?

karthikbekalp (commented):
Only 1 comment on the naming of the variable to simplify the logic. Let me know your thoughts.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
gianelli-lagos-aws: Yes, having a double negative can be quite confusing and not good. I have renamed it. Renaming it led to changing the other PRs that go along with this one since using activate_error_checking is clearer than deactivate_error_checking in general.

karthikbekalp: Discussed this offline and updated the name to "Activate Error Checking" everywhere.

test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py
```diff
@@ -9,6 +11,30 @@
...
+    re.compile(r".*Files cannot be written.*", re.IGNORECASE),
+    re.compile(r".*Enter Registration Data.*", re.IGNORECASE),
+    re.compile(r".*Unable to write file.*", re.IGNORECASE),
+    re.compile(r".*\[rlm\] abort_on_license_fail enabled.*", re.IGNORECASE),
+    re.compile(r".*RenderDocument failed with return code.*", re.IGNORECASE),
+    re.compile(r".*Frame rendering aborted.*", re.IGNORECASE),
+    re.compile(r".*Rendering was internally aborted.*", re.IGNORECASE),
+    re.compile(r'.*Cannot find procedure "rsPreference".*', re.IGNORECASE),
+]
+
```
karthikbekalp: Hmm, I wonder if we want to duplicate this? We do have checks in https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/523dadc435e1b585892c52ac232dcff27bbc479a/test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py#L79-L106 to ensure that we fail the jobs when a substring is detected. 

The test can be just to check whether the length of the error regex is greater than 0 no?

karthikbekalp (changes requested):
Just one comment to simplify the tests and avoid duplication.

test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py
gianelli-lagos-aws: Yes, I was wondering that as well. I think that if we were to check whether the length is greater than 0 or not, then that would just test anything that is added to the callback_list and not the specific error_regexes. In adaptor.py, there are other regexes added to the callback_list that are not in error_regexes. Not sure if checking the length would be the best here. I was wondering if I should create a separate error_regexes file which both test_adaptor.py and adaptor.py can import. Let me know what you think!

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
karthikbekalp: nit: The comment is slightly out of date now.

karthikbekalp (dismissed):
Just 1 nit. But other than that, everything looks good!

test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py
karthikbekalp: Discussed offline about this. We decided to check whether any regex calls the handle_error function or not.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
gianelli-lagos-aws: True! I will update it.

joel-wong-aws:
nit: the PR title (and commit message) doesn't need `with unit tests`, since that doesn't help the user understand the change

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/schemas/init_data.schema.json
```diff
@@ -5,7 +5,8 @@
         "scene_file": { "type": "string" },
         "take": { "type": "string" },
         "output_path": { "type": "string" },
-        "multi_pass_path": { "type": "string" }
+        "multi_pass_path": { "type": "string" },
+        "activate_error_checking": { "type": "string" }
```
joel-wong-aws, line 9: nit: to make this more generalizable, we could make this an enum with string values like `all`, `none`, which would open it to more options in the future.

That said, since making this an enum would not change the data type, it's backward compatible with this. This is good to go as-is!

test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py
```diff
@@ -197,3 +201,37 @@ def test_adaptor_prints_version_on_init(init_data, capfd):
...
+    )
+
+    if activate_error_checking == 1:
+        assert (
+            has_handle_error_callback
+        ), "Error checking should be activated when activate_error_checking=1"
+    else:
+        assert (
+            not has_handle_error_callback
+        ), "Error checking should be deactivated when activate_error_checking=0"
```
joel-wong-aws, line 237: Thanks for adding the unit tests!

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -275,7 +278,13 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
                 re.compile(r".*Rendering was internally aborted.*", re.IGNORECASE),
                 re.compile(r'.*Cannot find procedure "rsPreference".*', re.IGNORECASE),
             ]
-            callback_list.append(RegexCallback(error_regexes, self._handle_error))
+
+            # Only add error regexes if error checking is activated
+            if self._activate_error_checking:
+                _logger.warning("Adding error regexes to callback list")
```
joel-wong-aws: nit: this doesn't need to be a warning - there's nothing wrong/actionable with this flow. Same with the next `_logger.warning`. Both can be `_logger.info`

joel-wong-aws (dismissed):

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
gianelli-lagos-aws: Hello! I agree that the first log should be info since it would be stating the usual settings as a warning. However, I think the 2nd log should be a warning since we are changing what is usually expected by not adding error regexes. Let me know what you think. Thank you!

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/schemas/init_data.schema.json
gianelli-lagos-aws: I think that with string values like all and none, it would make the logic a bit more complex for now. I think it is best to keep it simple with "1" and "0".  Thank you! That is a good point for having more options in the future especially if we use other options other then numbers.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -275,7 +278,13 @@ def _get_regex_callbacks(self) -> list[RegexCallback]:
...
                 re.compile(r'.*Cannot find procedure "rsPreference".*', re.IGNORECASE),
             ]
-            callback_list.append(RegexCallback(error_regexes, self._handle_error))
+
+            # Only add error regexes if error checking is activated
+            if self._activate_error_checking:
+                _logger.info("Adding error regexes to callback list")
+                callback_list.append(RegexCallback(error_regexes, self._handle_error))
+            else:
+                _logger.warning("NOT adding error regexes to callback list")
```
joel-wong-aws, line 287: nit: this doesn't need to be `warning` level either: if the deactivate fail on errors option is selected, then this is intended and there isn't anything actionable here either. It can be `info` too

joel-wong-aws (approved):

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
gianelli-lagos-aws: Hello! I think it is best to leave it as warning since it can be easier for users to see that the error checking is deactivated. Warning makes it stand out instead of getting lost in the logs. Although selecting deactivate error checking is intentional, it still makes a significant change to the adaptor and that should stand out. If a job were to fail unexpectedly, warning level would help users understand why that happened faster.

joel-wong-aws: Makes sense!

karthikbekalp (approved):

Issue #210, opened 2025-05-01
Feature request: Tile rendering in Cinema 4D


karthikbekalp:
### Describe the problem

It would be useful to be able to use tile rendering so that we can speed up the rendering time. 

### Proposed Solution

We have a blog post on how one can create tile rendering in Maya: https://aws.amazon.com/blogs/media/create-a-tile-rendering-job-with-modifications-for-aws-deadline-cloud/
 
Cinema 4d has "Render Tiles" in its asset browser that can be used for tile rendering. 
Then, it can be combined using FFmpeg over here: https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/job_bundles/tile_render_maya_ffmpeg_for_blogpost/template.yaml#L278-L318

### Example Use Cases

This can allow us to create multiple jobs that can be rendered in parallel and then stitched together.

karthikbekalp:
Turns out Cinema 4D supports tile rendering natively. Added instructions for it: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/docs/tile_rendering/tile_rendering.md 

Customers would still have to stitch the files themselves though.

PR #256, opened 2025-07-17
feat: Add tile rendering support in Cinema 4D submitter

gianelli-lagos-aws:
Fixes: *<https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/210>*

Cinema 4D users need tile rendering support to efficiently render large images by breaking them into smaller tiles that can be processed in parallel, reducing memory usage and improving performance.

*Added tile rendering configuration options to the Cinema 4D submitter UI by implementing a checkbox to enable/disable tile rendering.
*Updated job template generation to include tile rendering parameters
*Integrated ffmpeg to automatically stitch rendered tiles back into a single output image
*Implement solution for the submitter to retrieve the tile rendering data needed such as tilers per axis from the camera object's user data. Once retrieving tile per axis, that will be used to automatically set the right frame range needed for tile rendering.
*Create a doc with tile rendering instructions *<https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/260>*
*Automatically added the needed CondaPackage (ffmpeg) and CondaChannel (conda-forge) for tile rendering to work if the user checks the use tile rendering checkbox.
tile_rendering

*Users can now render large, high-resolution images more efficiently by splitting them into smaller tiles
*Improves performance by enabling parallel processing of image sections
*Maintains output quality with seamless stitching of tiles into a single final image
*This works for windows and linux

*Verified correct job bundle generation with tile rendering parameters
*Tested tile rendering and stitching with various tile counts
*Validated that the final stitched output matches expected quality standards.
*Validated that the OutputPath and MultiPassPath are handled correctly since both can be selected with the same or different path leading to different asset-roots. If both OutputPath and MultiPassPath are selected then the outputs from both will be stitched. This leads to 2 stitched output images, one for the regular output and one for multi-pass outputs. This can be seen in the image below. 
Screenshot 2025-07-17 at 5 08 28 PM

Yes.
0.07s call     test/unit/deadline_submitter_for_cinema4d/test_submitter.py::test_get_job_template
0.02s call     test/unit/deadline_submitter_for_cinema4d/test_scene.py::test_get_output_directores
0.02s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[Project not found-True]
0.02s call     test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py::TestCinema4DAdaptor_on_cleanup::test_handle_errors_on_error_stdout[Asset missing-True]
====================================================================================== 56 passed in 4.06s =====================

Yes.
424.53s call     test/integ/test_cinema4d.py::test_integ[physical_multi_takes]
93.59s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
77.01s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
74.80s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
73.91s call     test/integ/test_cinema4d.py::test_integ[redshift]
================================================================================ 7 passed in 849.37s (0:14:09) ===========================

Yes. I added a checkbox. The checkbox allows the job template to have the tile rendering support and set other tile rendering requirements. 

----


src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -484,6 +496,12 @@ def create_job_bundle(
         for step in job_template["steps"]:
             step["hostRequirements"] = host_requirements
 
+    if settings.use_tile_rendering:
+        for step in job_template["steps"]:
+            step["hostRequirements"] = {
+                "attributes": [{"name": "attr.worker.os.family", "anyOf": ["windows"]}]
```
karthikbekalp: Hmm, `ffmpeg` should be able to be run on Linux as well right, is there a reason why we want to limit it to just Windows?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -204,6 +207,15 @@ def _get_job_template(
                 % take_data.name
             )
 
+    # For tile rendering, preserve tile assembly step
+    if settings.use_tile_rendering and len(job_template["steps"]) >= 1:
+        tile_rendering_step_copy = deepcopy(tile_rendering_step)
+        if "dependencies" in tile_rendering_step_copy:
+            for dependency in tile_rendering_step_copy["dependencies"]:
+                if "dependsOn" in dependency:
+                    dependency["dependsOn"] = take_data.display_name
```
karthikbekalp: Hmm, the tile rendering step should have dependency on all other steps (i.e. takes) not just one of the steps correct?

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +131,187 @@ steps:
...
+              # Format required by ffmpeg: "file '/path/to/file.ext'" (one per line)
+              > "$concat_file"  # Initialize empty file
+              for file in "${selected[@]}"; do
+                echo "file '$file'" >> "$concat_file"
+              done
+              
+              # Execute ffmpeg to stitch tiles together
+              # Parameters:
+              # -y                : Overwrite output file without asking
+              # -f concat -safe 0 : Use concat demuxer with relaxed path safety
```
karthikbekalp: Hmm, can you explain why we need this flag? Maybe describe what "concat demuxer with relaxed path" safety means? 
I think these are 2 flags: 
- "-f concat" to specify format as "contact" i.e. to concatenate the images. 
- "-safe 0" to deactivate safety checks, any reason why we added it?

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +131,187 @@ steps:
...
         cancelation:
           mode: NOTIFY_THEN_TERMINATE
+- name: tile assembly
+  dependencies:
+  - dependsOn: Render
+  hostRequirements:
+    attributes:
+    - name: attr.worker.os.family
+      anyOf:
+      - windows
```
karthikbekalp: Is this Windows specific, can we modify the command so that it works on Linux and Windows?

karthikbekalp (changes requested):
Thanks for the PR. Added few comments for some clarifications. Feel free to add more description about the tests that you did as well as any UI changes.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Yes! I just tested it and will update this PR soon. Thank you!

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
gianelli-lagos-aws: Yes, I will add more to the comments.
-f concat: tells ffmpeg to use the concat demuxer to read multiple input files as a single stream. Using single stream is better because we are using `-filter_complex tile=Param.TilesPerAxisxParam.TilesPerAxis`. `tile` allows us to tile several successive frames together. Having separate streams, one for each image, and keeping track of them would make this complicated since you can have 2x2, 3x3, 4x4, or 5x5.
-safe 0: deactivates safety checks to allow absolute file paths in the concat file. The file paths put into the concat file are absolute paths.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: `ffmpeg` does run on linux as well! I will remove the Host Requirements. Thank you!

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
gianelli-lagos-aws: `ffmpeg` works on linux. I will update this! Thank you.

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -55,6 +55,16 @@ parameterDefinitions:
...
   - '1'
+- name: TilesPerAxis
+  type: INT
+  userInterface:
+    control: SPIN_BOX
+    label: Tiles per axis
+    groupLabel: Tile Rendering Settings
+  minValue: 2
+  maxValue: 5
+  default: 2
```
joel-wong-aws: Why a default of 2? That means 4 tiles per frame, right?

My understanding is that users typically want to start without tiling, and may add tiling later if their frames are taking too long to render

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -85,6 +85,7 @@ def _get_parameter_values(
     parameter_values.append(
         {"name": "ActivateErrorChecking", "value": settings.activate_error_checking}
     )
+    parameter_values.append({"name": "TilesPerAxis", "value": settings.tiles_per_axis})
```
joel-wong-aws: If the `settings.use_tile_rendering` option is not selected, we shouldn't send the `TilesPerAxis` job parameter (and it should be removed from our job template), otherwise it will confusingly show in our job parameters:
image

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -204,6 +207,16 @@ def _get_job_template(
                 % take_data.name
             )
 
+        # For tile rendering, preserve tile assembly step
+        if settings.use_tile_rendering and len(job_template["steps"]) >= 1:
```
joel-wong-aws, line 335: When I submit a multi-take job with different frame ranges for each take and tile rendering, the job fails to create with the following error:
```
Error occurred: Could not decode job template. Error: '17 validation errors for JobTemplate
steps[11] -> name:
	String should have at most 64 characters
steps[13] -> name:
	String should have at most 64 characters
steps[1] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[3] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[5] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[7] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[9] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[11] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[13] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[15] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[17] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[19] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[21] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[23] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[25] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[27] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames
steps[29] -> script -> embeddedFiles[0] -> data:
	Variable Param.Frames does not exist at this location. Did you mean one of: Param.AFrames, Param.BFrames'
job-f816c92ea43840b7a0c7c1762377806f
```

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +131,189 @@ steps:
         - file://{{ Task.File.runData }}
         cancelation:
           mode: NOTIFY_THEN_TERMINATE
+- name: Tile Assembly
```
joel-wong-aws: I'm confused. The purpose of tile rendering is to divide rendering work among multiple machines (right?). From what I can see, this renders all the tiles on the same worker. Am I missing something?

joel-wong-aws (changes requested):

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
gianelli-lagos-aws: Hello! Users can select whether they want tiles rendering support or not. The default is 2 because that is the minimum number of tiles per axis you can select on Cinema 4D using a Render Tiles Camera. The maximum value of tiles per axis is 5 in Cinema 4D so I set 5 as the max here as well.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: I agree! I will work on that. Thank you!

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
karthikbekalp: This step uses `ffmpeg` to stitch together all the frames that have been rendered on different machines. So even though the stills are rendered on multiple workers, the output images are stitched together at the end. 

Cinema 4D supports a camera called "Render Tiles" which would divide the scene into parts. For example, if the tiles per axis is 2 then it divides 1 frame into 4 frames i.e frame "0-3" which then can be stitched together. 

We will also add documentation on how customers can set this up its easy to follow the workflow they need to do to get tile rendering working.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -156,6 +157,18 @@ def _build_ui(self, settings):
 
         lyt.addWidget(export_group_box, 7, 0, 1, 2)
 
+        tile_rendering_group_box = QGroupBox("Tile Rendering Settings", self)
```
karthikbekalp, line 159: We also need to add a note so that customers have followed the instructions to enable tile rendering because just activating this button will not automatically divide the scene file.

gianelli-lagos-aws: Yes! I will add that as well. Thank you!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Interesting. I will take a look at the cinema4d file you sent and set up the tile rendering. Thank you!

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
karthikbekalp: nit: I think we can remove `# -f concat -safe 0 : Use concat demuxer with relaxed path safety` now as the comment describes it? 

Also, let's have the same indent (same number of spaces before `-f`) as the others options.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -160,6 +162,13 @@ def _get_job_template(
         # ignore if description is missing from template
         job_template.pop("description", None)
 
+    if not settings.use_tile_rendering:
```
karthikbekalp: We discussed about this offline, instead of adding the parameter value in the template and removing it later we would only add it whenever tile rendering is enabled.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -179,6 +188,8 @@ def _get_job_template(
 
     # Replicate the default step, once per render take, and adjust its settings
     default_step = job_template["steps"][0]
+    if settings.use_tile_rendering:
+        tile_rendering_step = job_template["steps"][1]
```
karthikbekalp: As discussed offline, it would be nice to have a separate function which adds changes to the job template to increase the readability of the code.

karthikbekalp (changes requested):
Nice work @gianelli-lagos-aws ! Just few nit and few comments to increase the readability of the code overall.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -146,6 +148,18 @@ def _get_parameter_values(
...
+def _add_tile_assembly_step(
+    job_template: dict[str, Any], tile_rendering_step: dict[str, Any], take_data: TakeData
+) -> None:
+    tile_rendering_step_copy = deepcopy(tile_rendering_step)
+    tile_rendering_step_copy["name"] = "Tile Assembly - " + take_data.display_name
+    if "dependencies" in tile_rendering_step_copy:
+        for dependency in tile_rendering_step_copy["dependencies"]:
+            if "dependsOn" in dependency:
+                dependency["dependsOn"] = take_data.display_name
+    job_template["steps"].append(tile_rendering_step_copy)
```
joel-wong-aws, line 251: If we're adding the tile assembly template, can we assert that the frame range is in the form `X-Y` and `Y - X = tiles_per_axis * tiles_per_axis`?

The user could also accidentally input `something thing `0-18:2`, which means "every second frame starting from 0" and would cause the ffmpeg step to fail

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +121,188 @@ steps:
...
+        # Returns:
+        #   0 on success, 1 on failure
+        # =============================================================================
+        stitch_tiles() {
+          local asset_root="$1"    # Directory containing tile images
+          local output_name="$2"   # Base name for output file
+          
+          # Find all supported image files in the asset directory
+          # Supported formats: tif, png, jpg, jpeg, tga, bmp, hdr, dpx (Cinema 4D compatible)
+          local files=($(find "$asset_root" -type f \( -iname "*.tif" -o -iname "*.png" -o -iname "*.jpg" -o -iname "*.jpeg" -o -iname "*.tga" -o -iname "*.bmp" -o -iname "*.hdr" -o -iname "*.dpx" \) 2>/dev/null | sort))
```
joel-wong-aws: nit: `-iname "*.tif" -o -iname "*.png" -o -iname "*.jpg" -o -iname "*.jpeg" -o -iname "*.tga" -o -iname "*.bmp" -o -iname "*.hdr" -o -iname "*.dpx"` is quite repetititive, can we update the script so that it pulls the allowed file extensions from a list?

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +121,188 @@ steps:
...
+              
+              return 0
+            else
+              printf "ERROR: Expected %d tiles but found %d\n" "$EXPECTED_TILES" "${#selected[@]}"
+              return 1
+            fi
+          else
+            printf "No image files found in %s\n" "$asset_root"
+            return 1
+          fi
```
joel-wong-aws: nit: instead of using if-elses for these error conditions, can we use a guard clause?

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +121,188 @@ steps:
...
+              local skip_count=$EXPECTED_TILES
+              local current_count=0
+              
+              for file in "${files[@]}"; do
+                # Skip if file matches any of the output file names
+                if [[ "$file" == *"stitched_output"* || "$file" == *"stitched_multipass"* ]]; then
+                  continue
+                elif [ $current_count -lt $skip_count ]; then
+                  # Skip files already processed in main output
+                  current_count=$((current_count + 1))
```
joel-wong-aws: What happens if there are only multi-pass outputs and no regular outputs? Will this still work?

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -121,3 +121,188 @@ steps:
...
+              > "$concat_file"  # Initialize empty file
+              for file in "${selected[@]}"; do
+                echo "file '$file'" >> "$concat_file"
+              done
+              
+              # Execute ffmpeg to stitch tiles together
+              # Parameters:
+              # -y                : Overwrite output file without asking
+              # -f concat: Combines multiple image files into a single input stream for the tile filter
+              # -safe 0: Allows absolute file paths in the concat file
```
joel-wong-aws: It's unclear to me why we need these two arguments. We have a similar job bundle here and it doesn't need these arguments for tile rendering. `-safe 0` especially seems out of place:
https://github.com/aws-deadline/deadline-cloud-samples/blob/1369b2d0fe4c8c4236a988cfc69828149d685158/job_bundles/tile_render_maya_ffmpeg_for_blogpost/template.yaml#L315

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
gianelli-lagos-aws: If in Render Settings there are only multi-pass outputs being saved and not regular outputs, it will still work since the regular output path would be an empty string. If OutputPath is an empty string, it will not call the function and move onto MultiPassPath. When MultiPassPath is checked and calls the function, it will take the standard case which takes the first files and doesn't skip any. I only skip files when the OutputPath is equal to the MultiPassPath since to get to the multi-pass outputs I need to skip the regular outputs.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Yes! I can look into that. Thank you!

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
gianelli-lagos-aws: Yes, I agree. Thank you!

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
gianelli-lagos-aws: Hello! -safe 0 deactivates safety checks to allow absolute file paths in the concat file. Each file path put into the concat file is an absolute file path to the output. I am adding it because without it i get an Unsafe file name error.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,56 @@ def show_submitter():
...
+                continue
+            if tiles_per_axis_value and is_valid_tile_rendering_data(obj):
+                return tiles_per_axis_value
+            else:
+                raise ValueError(
+                    "Tile rendering is checked and Render Tiles camera object was found but the Render Tiles camera configuration was invalid.\nAdd a Reference camera and check `Use Tiling`"
+                )
+
+    raise ValueError(
+        "Tile rendering is checked but no Render Tiles camera object was found in the scene.\nRefer to the tile rendering instructions in the Cinema 4D GitHub repository: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d"
```
joel-wong-aws: Can we add a separate document specifically for setting up tile rendering? That way, it will be clear to users where to look for instructions.

Also, we need to write the documentation :)

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,56 @@ def show_submitter():
...
+                # Create a DescID for the Tiles per Axis parameter
+                tiles_per_axis_desc_id = c4d.DescID(
+                    c4d.DescLevel(700, 5, 0), c4d.DescLevel(3, 15, 0)
+                )
+                tiles_per_axis_value = obj[tiles_per_axis_desc_id]
+            except Exception:
+                continue
+            if tiles_per_axis_value and is_valid_tile_rendering_data(obj):
+                return tiles_per_axis_value
+            else:
```
joel-wong-aws: nit: the `else` is unnecessary here because the `if` condition always returns. We can simply do 

```
if tiles_per_axis_value and is_valid_tile_rendering_data(obj):
    return tiles_per_axis_value
raise ValueError(
    "Tile rendering is checked and Render Tiles camera object was found but the Render Tiles camera configuration was invalid.\nAdd a Reference camera and check `Use Tiling`"
)

File: src/deadline/cinema4d_submitter/cinema4d_render_submitter.py [hunk 19]
```diff
@@ -80,6 +81,56 @@ def show_submitter():
...
+
+    for obj in objects:
+        if isinstance(obj, c4d.CameraObject):
+            try:
+                # Create a DescID for the Tiles per Axis parameter
+                tiles_per_axis_desc_id = c4d.DescID(
+                    c4d.DescLevel(700, 5, 0), c4d.DescLevel(3, 15, 0)
+                )
+                tiles_per_axis_value = obj[tiles_per_axis_desc_id]
+            except Exception:
```
joel-wong-aws: What type of exception is raised and where? e.g. is it a `KeyError` on `tiles_per_axis_value = obj[tiles_per_axis_desc_id]`? Can we be more specific about what types of exceptions we capture? e.g. 

```
            # Create a DescID for the Tiles per Axis parameter
            tiles_per_axis_desc_id = c4d.DescID(
                c4d.DescLevel(700, 5, 0), c4d.DescLevel(3, 15, 0)
            )

           try:
                tiles_per_axis_value = obj[tiles_per_axis_desc_id]
            except KeyError:
                continue

File: src/deadline/cinema4d_submitter/cinema4d_render_submitter.py [hunk 20]
```diff
@@ -170,6 +235,18 @@ def _get_job_template(
...
 
+    if settings.use_tile_rendering:
+        job_template["parameterDefinitions"].append(
+            {
+                "name": "TilesPerAxis",
+                "type": "INT",
+                "description": "The number of tiles across the x and y axis.",
+                "minValue": 2,
+                "maxValue": 5,
+                "default": 2,
```
joel-wong-aws: Can we make the default the same as `get_tiles_per_axis_value()`?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -683,6 +764,21 @@ def on_create_job_bundle_callback(
         """
         Callback function for creating a job bundle when submitting the job.
         """
+
+        # If `use_tile_rendering` is checked, ffmpeg and conda-forge must be added to its corresponding parameter.
+        # This allows tile rendering support to work.
+        for param in queue_parameters:
+            if param["name"] == "CondaChannels" and settings.use_tile_rendering:
```
joel-wong-aws: I think the logic should be re-ordered here. This entire block should be wrapped in `if settings.use_tile_rendering` instead of the `for` loop. That makes it easier to understand and is more efficient

src/deadline/cinema4d_submitter/tile_assembly_step.py
```diff
@@ -0,0 +1,204 @@
...
+#   $2 (output_name): Base name for the output stitched image
+# Returns:
+#   0 on success, 1 on failure
+# =============================================================================
+stitch_tiles() {
+  local asset_root="$1"    # Directory containing tile images
+  local output_name="$2"   # Base name for output file
+  
+  # Find all supported image files in the asset directory
+  # Supported formats: tif, png, jpg, jpeg, tga, bmp, hdr, dpx (Cinema 4D compatible)
```
joel-wong-aws: supernit: this is now self-documented in the code, so there's no need for this duplicate comment :)

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -146,6 +199,18 @@ def _get_parameter_values(
     return parameter_values
 
 
+def _add_tile_assembly_step(
+    job_template: dict[str, Any], tile_rendering_step: dict[str, Any], take_data: TakeData
+) -> None:
+    tile_rendering_step_copy = deepcopy(tile_rendering_step)
+    tile_rendering_step_copy["name"] = "Tile Assembly - " + take_data.display_name
```
joel-wong-aws: Do we need to truncate this to 64 characters? `take_data.display_name` can already be 64 characters, and 64 characters is the max step name length, so this could exceed that.

joel-wong-aws (changes requested):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Yes, I am working on a PR for that!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: True! thank you.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Yes it does have to do with `tiles_per_axis_value = obj[tiles_per_axis_desc_id]`. I can update that for more clarity.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Ohh I see. Not sure but I can look into that.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: I agree. thank you!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: It is an AttributeError

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,67 @@ def show_submitter():
         traceback.print_exc()
 
 
+def is_valid_tile_rendering_data(camera_obj):
+    """
+    Check if the Tile Rendering camera has valid user data.
+    """
+    try:
+        # Use the correct DescID for the Reference camera
+        reference_camera_desc_id = c4d.DescID(c4d.DescLevel(700, 5, 0), c4d.DescLevel(4, 133, 0))
```
karthikbekalp, line 100: Did we get these values (i.e. 700,5,0 etc) from some public Cinema 4D documentation? 

The reason I ask this is to understand if these values could change in the future and hence break our workflow?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,67 @@ def show_submitter():
...
+        # Use the correct DescID for the "Use Tiling" checkbox
+        use_tiling_desc_id = c4d.DescID(c4d.DescLevel(700, 5, 0), c4d.DescLevel(1, 400006001, 0))
+        use_tiling_state = camera_obj[
+            use_tiling_desc_id
+        ]  # Value is either 1 for checked or 0 for unchecked
+
+        # Check if the camera has the valid tile rendering data
+        return use_tiling_state and isinstance(reference_camera_obj, c4d.CameraObject)
+    except Exception:
+        return False
```
karthikbekalp, line 115: Can we add a log line if we get exceptions?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,67 @@ def show_submitter():
         traceback.print_exc()
 
 
+def is_valid_tile_rendering_data(camera_obj):
+    """
+    Check if the Tile Rendering camera has valid user data.
```
karthikbekalp: Hmm, its not clear what is meant by "valid user data" here? Can you elaborate a bit?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,67 @@ def show_submitter():
...
+            # Create a DescID for the Tiles per Axis parameter
+            tiles_per_axis_desc_id = c4d.DescID(c4d.DescLevel(700, 5, 0), c4d.DescLevel(3, 15, 0))
+            try:
+                tiles_per_axis_value = obj[tiles_per_axis_desc_id]
+            except AttributeError:
+                continue
+            if tiles_per_axis_value and is_valid_tile_rendering_data(obj):
+                return tiles_per_axis_value
+            raise ValueError(
+                "Tile rendering is checked and Render Tiles camera object was found but the Render Tiles camera configuration was invalid.\nAdd a Reference camera and check `Use Tiling`"
```
karthikbekalp: nit: Use "non valid" instead.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: +1 to Joel's idea, if we do know what the doc URL would be, we could add that here and when the documentation PR gets, it will no longer be a broken link.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -80,6 +81,67 @@ def show_submitter():
...
+    raise ValueError(
+        "Tile rendering is checked but no Render Tiles camera object was found in the scene.\nRefer to the tile rendering instructions in the Cinema 4D GitHub repository: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d"
+    )
+
+
+def set_tile_rendering_frame_range():
+    """
+    Set the frame range for Tile Rendering.
+    """
+    tiles_per_axis = get_tiles_per_axis_value()
```
karthikbekalp: nit: Hmm, I believe `get_tiles_per_axis_value` also uses the doc variable? Can we re-order the line below so that we pass this value to that function as an argument to that function instead of calling `c4d.documents.GetActiveDocument()`?

src/deadline/cinema4d_submitter/tile_assembly_step.py
```diff
@@ -0,0 +1,208 @@
...
+      done
+    else
+      # Standard case - select tiles that match the output filename pattern
+      for file in "${files[@]}"; do
+        if [ $count -lt $EXPECTED_TILES ]; then
+          # Match files containing the base filename (without extension)
+          if [[ "$file" == *"$(basename "$output_file_path" | cut -d. -f1)"* ]]; then
+            selected+=("$file")
+            count=$((count + 1))
+          fi
```
karthikbekalp: Hmm, I think it would be best to select tiles that just match the output file pattern no?

src/deadline/cinema4d_submitter/tile_assembly_step.py
```diff
@@ -0,0 +1,208 @@
...
+      done
+      
+      # Execute ffmpeg to stitch tiles together
+      # Parameters:
+      # -y                : Overwrite output file without asking
+      # -f concat: Combines multiple image files into a single input stream for the tile filter
+      # -safe 0: Allows absolute file paths in the concat file
+      # -filter_complex   : Apply tile filter to arrange images in grid
+      ffmpeg -y -f concat -safe 0 -i "$concat_file" -filter_complex tile={{Param.TilesPerAxis}}x{{Param.TilesPerAxis}} "$output_file"
+
```
karthikbekalp: nit: Generally the readability of the code is improved when we divide the code into smaller functions. 

For example, the stitch_files() function could have been divided into smaller functions like 

1. Get the files for stitching the outputs. 

2. Do the stitching using ffmpeg

3. Clean up

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Thanks for adding !

karthikbekalp (changes requested):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: No, these values were retrieved by going through each camera object and getting it's properties: 
```
objects = doc.GetObjects()
for obj in objects:
    if isinstance(obj, c4d.CameraObject):
        for data, pid, _ in obj.GetDescription(c4d.DESCFLAGS_GET_NONE):
            print(data[c4d.DESC_NAME])
            print(data)
            print(pid)
            print(_)
```
There should be a section that looks like this in the output: 
```
User Data
<c4d.BaseContainer object at 0x0000020E3998C440>
(700, 1, 0)
(1000491, 0, 0)
Tiles per Axis
<c4d.BaseContainer object at 0x0000020E399A2340>
((700, 5, 0), (3, 15, 0))
(700, 1, 0)
Reference Camera
<c4d.BaseContainer object at 0x0000020E3999C740>
((700, 5, 0), (4, 133, 0))
(700, 1, 0)
Use Tiling
<c4d.BaseContainer object at 0x0000020E3999C8C0>
((700, 5, 0), (1, 400006001, 0))
(700, 1, 0)

File: src/deadline/cinema4d_submitter/tile_assembly_step.py (hunk 29 above)
Comment (gianelli-lagos-aws) [reply src/deadline/cinema4d_submitter/tile_assembly_step.py]: Yes, I need to look further into that since I was having issues with using the full output file pattern.

File: src/deadline/cinema4d_submitter/tile_assembly_step.py (hunk 30 above)
Comment (gianelli-lagos-aws) [reply src/deadline/cinema4d_submitter/tile_assembly_step.py]: I agree, I will look into it! Ty!

File: src/deadline/cinema4d_submitter/cinema4d_render_submitter.py [hunk 31]
```diff
@@ -80,6 +81,84 @@ def show_submitter():
...
+            # Create a DescID for the Tiles per Axis parameter
+            tiles_per_axis_desc_id = c4d.DescID(c4d.DescLevel(700, 5, 0), c4d.DescLevel(3, 15, 0))
+            try:
+                tiles_per_axis_value = obj[tiles_per_axis_desc_id]
+            except AttributeError:
+                continue
+            if tiles_per_axis_value and has_valid_tile_rendering_configuration(obj):
+                return tiles_per_axis_value
+            raise ValueError(
+                "Tile rendering is checked and Render Tiles camera object was found but the Render Tiles camera configuration was non valid.\nAdd a Reference camera and check `Use Tiling`"
```
joel-wong-aws, line 140: Please link to the docs here as well :)

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -146,6 +236,21 @@ def _get_parameter_values(
...
 
 
+def _add_tile_assembly_step(
+    job_template: dict[str, Any], tile_rendering_step: dict[str, Any], take_data: TakeData
+) -> None:
+    tile_rendering_step_copy = deepcopy(tile_rendering_step)
+    prefix = "Tile Assembly - "
+    max_name_length = 64 - len(prefix)
+    truncated_name = take_data.display_name[:max_name_length]
+    tile_rendering_step_copy["name"] = prefix + truncated_name
```
joel-wong-aws, line 246: nit: can we simplify this to simply 
`tile_rendering_step_copy["name"] = f"Tile Assembly -{take_data.display_name}"[:64]`

PR #263, opened 2025-07-28
docs: Cinema 4D Tile Rendering Set up

gianelli-lagos-aws:

Add a doc that shows how to set up Tile Rendering in Cinema 4D.

Added a doc that contains the instructions on how to set up tile rendering in Cinema 4D.

Customers will be able to use tile rendering.

----


joel-wong-aws (dismissed):

docs/tile_rendering/tile_rendering.md
```diff
@@ -0,0 +1,66 @@
...
+### Initial Setup
+
+#### Set Up Camera
+
+1. Add a default physical camera object to your scene
+2. Position the camera to properly frame the objects you want to render
+
+#### Add Render Tiles Camera
+
+1. Locate the "Render Tiles" camera in the Asset Browser (under "Model" section)
```
karthikbekalp, line 24: We discussed about this offline. We will add few more pictures to provide more guidance for customers who want to try out tile rendering

docs/tile_rendering/tile_rendering.md
```diff
@@ -0,0 +1,69 @@
+# Setting Up Tile Rendering in Cinema 4D for Deadline Cloud
+
+*These instructions include a 3x3 tile render as an example.*
+
+Tile rendering divides a single image into smaller sections (tiles) that are rendered separately, then stitched back together into the final image. This approach can significantly reduce render times for large or complex scenes by distributing the workload.
```
karthikbekalp: nit: Can we rephrase this so as to suggest that customers would have to themselves stitch the files together later? 

I think it is mentioned at the end of this doc but at that point it can be late for customers who don't want to stitch it themselves.

docs/tile_rendering/tile_rendering.md
karthikbekalp: Thanks for adding this. Now it should be much easier for customers to follow.

karthikbekalp (dismissed):
Just 1 small comment about stitching the images together. I think it would be best if we share to our customers in the beginning rather than towards the end that we will not stitch the files together for them. 
But not a blocker.

docs/tile_rendering/tile_rendering.md
gianelli-lagos-aws: Completely agree, thank you!

karthikbekalp (approved):
Great job Gianelli!

docs/tile_rendering/tile_rendering.md
```diff
@@ -0,0 +1,69 @@
...
+
+*These instructions include a 3x3 tile render as an example.*
+
+Tile rendering divides a single image into smaller sections (tiles) that are rendered separately. You will need to stitch these tiles back together yourself using professional image editing software or scripting tools. This approach can significantly reduce render times for large or complex scenes by distributing the workload.
+
+## Renderer Compatibility
+
+* You can use any renderer (Physical, Redshift, or Standard) for your scene
+* **Important:** The reference camera used by the Render Tiles camera must be a Physical camera, regardless of your scene's renderer
+
```
karthikbekalp, line 11: nit: Do we need this ? I believe we do have setup camera step where we ask customers to use "Physical"?

joel-wong-aws (approved):

docs/tile_rendering/tile_rendering.md
gianelli-lagos-aws: I think this is good to have since it makes it clear that other options such as the redshift camera are not going to work.

karthikbekalp: Cool, thanks for the clarification.

PR #269, opened 2025-08-04
docs: Publish user guide to GitHub pages

gianelli-lagos-aws:
Add a GitHub page that contains a user guide for Cinema 4D and AWS Deadline Cloud. 

Add a GitHub page that has a quick step by step instructions with good visuals for onboarding.
Here is a link to see what the page looks like: https://gianelli-lagos-aws.github.io/deadline-cloud-for-cinema-4d/
The gh-pages branch that is generated by the mkdocs.yaml file should be automatically created and updated by GitHub Actions when the PR gets merged. Then in settings the gh-page branch must be added to the Pages tab.

Customers will have a better onboarding experience.

----


docs/user_guide/index.md
```diff
@@ -0,0 +1,143 @@
...
+*The Deadline Cloud submitter integrates directly into Cinema 4D's Extensions menu.*
+
+## Why Use Deadline Cloud for Cinema 4D?
+
+<div class="benefit-tabs">
+  <div class="carousel-container">
+    <button class="carousel-arrow left" onclick="prevTab()">‹</button>
+    <div class="tab-content-wrapper">
+      <div class="tab-content active" id="tab-0">
+        <h3>⚡ Scale Your Renders ⚡</h3>
```
karthikbekalp: nit: I think the content in the carousel is almost as big as the section titles. I think it would be better if they were a little smaller.

docs/user_guide/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+function nextTab() {
+  const nextIndex = (currentTab + 1) % 4;
+  showTab(nextIndex);
+}
+
+function prevTab() {
+  const prevIndex = (currentTab - 1 + 4) % 4;
+  showTab(prevIndex);
+}
```
karthikbekalp: Hmm, what does this script do in particular? Is it for focusing on tabs?

mkdocs.yml
```diff
@@ -0,0 +1,41 @@
...
+  - Quick Start: getting-started.md
+
+markdown_extensions:
+  - toc:
+      permalink: true
+  - attr_list
+  - md_in_html
+  - pymdownx.emoji:
+      emoji_index: !!python/name:material.extensions.emoji.twemoji
+      emoji_generator: !!python/name:material.extensions.emoji.to_svg
```
karthikbekalp: Hmm, where are these extensions coming from? Are these from Github?

karthikbekalp (changes requested):
Nice job Gianelli! The site looks really cool. I have a suggestion about our approach moving forward.

I noticed we're starting to add scripts and HTML alongside the markdown. While this gives us more flexibility, I'm wondering if we should stick to using just the default markdown syntax for simplicity? 

My main concern is that adding scripts and custom HTML will create additional maintenance overhead - we'd need to maintain both our content AND our custom front-end code. Using plain markdown files would make it much easier for anyone to contribute content in the future without needing to understand the additional customization.

However, I haven't worked extensively with frontend development, so I might be missing some important considerations. @crowecawcaw, would you be able to share your thoughts on this? I'm particularly interested in whether keeping it simple with markdown would limit us too much in the long run.

docs/user_guide/index.md
gianelli-lagos-aws: I agree, it could be a little smaller. I will work on that. Thank you!

docs/user_guide/index.md
gianelli-lagos-aws: Yes, this is for the carousel. It allows the slides cycle by moving to the next tab and back to the start by itself. There are also the options to manually click for the next tab and the previous tab. If the user clicks that then the timer resets for the automatic cycle of slides.

docs/user_guide/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+## The Cinema 4D Submitter
+
+![Cinema 4D Submitter Interface](images/submitter-dialog.png)
+
+*The Deadline Cloud submitter integrates directly into Cinema 4D's Extensions menu.*
+
+## Why Use Deadline Cloud for Cinema 4D?
+
+<div class="benefit-tabs">
```
crowecawcaw: These docs look great!

Agree with @karthikbekalp 's suggestion to stick to plain Markdown for now. The custom tiles and carousel look great. In the near future though, I expect we'll have a standardized template for the user guides, or maybe build them all together into a single centralized site. It'll be easier to make that migration if there's less custom HTML and JS in the mix.

mkdocs.yml
gianelli-lagos-aws: They are python markdown extensions, I found documentation here https://pypi.org/project/pymdown-extensions/ and that includes this link https://facelessuser.github.io/pymdown-extensions/ which shows how to use it. I will actually be removing this since it this can cause issues if the maintainers make changes.

docs/user_guide/index.md
gianelli-lagos-aws: Thank you! Ohh I see, makes sense. I will work on the plain Markdown files then.

karthikbekalp (dismissed):

docs/user_guide/getting-started.md
```diff
@@ -0,0 +1,66 @@
...
+2. Launch Cinema 4D after installation
+3. Verify the submitter appears in `Extensions` > `AWS Deadline Cloud Submitter`
+
+## Updating the Submitter
+
+To update the submitter to the latest version, download and run the latest submitter installer.
+
+## Step 2: Submit Your First Render (2 minutes)
+
+1. Open Cinema 4D and load a test scene. **Or use our sample scenes:** Go to [`test_scenes`](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/tree/mainline/test/integ/test_scenes), select a scene folder, navigate to its `scene` subfolder, and run `scene.py` in Cinema 4D via `Extensions` > `User Scripts` > `Run Script` (this will generate a .c4d file for you)
```
crowecawcaw, line 31: Nit: I think we should out the test scene guidance. The user guide is more geared towards end users using the GUI instead of developers using code. Users will probably also have a scene already.

docs/user_guide/getting-started.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Step 3: Monitor Your Renders
+
+If you haven't already, install the Deadline Cloud Monitor from the requirements above.
+
+After submitting a job, open Deadline Cloud Monitor (DCM) to view the job's progress. The submitter will create a job with a single step and with one task per frame.
+![DCM](images/DCM.png)
+
+**To view rendering logs:** Right-click on a task and choose "View logs"  
+This is especially useful when your job fails to troubleshoot issues.
```
crowecawcaw:
```suggestion
Viewing logs is especially useful for troubleshooting failed jobs.  
```

docs/user_guide/getting-started.md
crowecawcaw:
```suggestion
1. Open Cinema 4D and load a scene.
```

Nit: The user guide is geared towards end users who are using a GUI instead of developers using code, so we can maybe leave this out.

crowecawcaw (changes requested):

docs/user_guide/getting-started.md
gianelli-lagos-aws: I agree, thank you!

crowecawcaw (approved):

karthikbekalp (approved):

PR #271, opened 2025-08-06
docs: Update Tile Rendering Instructions doc

gianelli-lagos-aws:
Tile Rendering instructions were missing a key step.

Added the step to divide both the Width and Height by the number of tiles per axis (e.g., for 3x3 tiles, divide original dimensions by 3)

Customers will have faster render times.

----


karthikbekalp (approved):
LGTM. Thanks!

joel-wong-aws (approved):

PR #273, opened 2025-08-07
chore: update pyproject metadata to production

joel-wong-aws:
The Cinema 4D integration is already production-grade, but the [PyPi metadata](https://pypi.org/project/deadline-cloud-for-cinema-4d/) doesn't reflect that. Updated the metadata.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ authors = [
 ]
 # https://pypi.org/classifiers/
 classifiers = [
-  "Development Status :: 3 - Alpha",
+  "Development Status :: 5 - Production/Stable",
   "Programming Language :: Python",
   "Programming Language :: Python :: 3 :: Only",
   "Programming Language :: Python :: 3.9",


karthikbekalp (commented):
Hmm, the version is still 0.* ? Does production grade mean that we need to update it 1.0 version now?

joel-wong-aws:
> Hmm, the version is still 0.* ? Does production grade mean that we need to update it 1.0 version now?

I don't think so. All of the [other integrations](https://github.com/aws-deadline) are labeled similarly and have `0.*` releases. This just makes it consistent with other packages. The submitter is production grade.

karthikbekalp (commented):
~~Just curious, our repo release versions are generally 0.*, does this mean we need to update the release versions as well? Or any other issues with us upgrading to "Production"? Is there a bar that we need to meet?~~

This is a duplicate comment.

karthikbekalp:
> > Hmm, the version is still 0.* ? Does production grade mean that we need to update it 1.0 version now?
> 
> I don't think so. All of the [other integrations](https://github.com/aws-deadline) are labeled similarly and have `0.*` releases. This just makes it consistent with other packages. The submitter is production grade.

Cool, thanks for confirming.

karthikbekalp (approved):

AWS-Samuel (approved):

PR #283, opened 2025-08-08
feat: Implement font handling for Cinema 4D submitter for Windows.

karthikbekalp:

One of our customers pointed out that in Cinema 4D we can add different fonts for "Text" but for the fonts to be rendered properly, it needs to be installed on the workers. 

Added logic to install fonts during the job environment phase before the Cinema 4D application is opened. 

There is an issue with fonts in Mac and we will be working on it separately to fix it but it looks as though the issue is on the Cinema 4D side. 

Custom fonts would be rendered in Cinema 4D. 

I tested it by submitting a bunch of scenes with different fonts and confirmed that it worked. 

Yes

I also plan to add unit tests separately. 

Yes. Running them now. 

I also plan to add an integration test separately for this issue. 

Yes. 

No change to documentation is necessary. 

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/assets.py
+++ b/src/deadline/cinema4d_submitter/assets.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import c4d
 
 from .scene import Scene
+from .font_utils import is_asset_a_font, copy_font_to_scene_folder
 
 _FRAME_RE = re.compile("#+")
 
@@ -22,15 +23,36 @@ class AssetIntrospector:
         """
         # Grab tx files (if we need to)
         assets: set[Path] = set()
-        assets.add(Path(Scene.name()))
+
+        path_to_scene_file = Path(Scene.name())
+        path_to_scene_file_dir = path_to_scene_file.parent
+        assets.add(path_to_scene_file)
 
         doc = c4d.documents.GetActiveDocument()
         asset_list: list[dict] = []
-        c4d.documents.GetAllAssetsNew(doc, allowDialogs=False, lastPath="", assetList=asset_list)
+
+        c4d.documents.GetAllAssetsNew(
+            doc,
+            allowDialogs=False,
+            lastPath="",
+            assetList=asset_list,
+            flags=c4d.ASSETDATA_FLAG_WITHFONTS,
+        )
+
         for asset in asset_list:
+            if is_asset_a_font(asset):
+                copy_font_to_scene_folder(asset["assetname"], path_to_scene_file_dir)
+
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
             if exists is True and filename is not None:
                 assets.add(Path(filename))
+
+        # Add all font files from the tempFonts directory to assets
+        fonts_dir = path_to_scene_file_dir / "tempFonts"
+        if fonts_dir.exists():
+            for font_file in fonts_dir.iterdir():
+                if font_file.is_file():
+                    assets.add(font_file)
 
         return assets
--- a/src/deadline/cinema4d_submitter/assets.py
+++ b/src/deadline/cinema4d_submitter/assets.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import c4d
 
 from .scene import Scene
-from .font_utils import is_asset_a_font, copy_font_to_scene_folder
+from .font_utils import is_asset_a_font, copy_font_to_scene_folder, TEMP_FONTS_DIR
 
 _FRAME_RE = re.compile("#+")
 
@@ -49,7 +49,7 @@ class AssetIntrospector:
                 assets.add(Path(filename))
 
         # Add all font files from the tempFonts directory to assets
-        fonts_dir = path_to_scene_file_dir / "tempFonts"
+        fonts_dir = path_to_scene_file_dir / TEMP_FONTS_DIR
         if fonts_dir.exists():
             for font_file in fonts_dir.iterdir():
                 if font_file.is_file():
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -7,6 +7,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Optional
 
+import shutil
+
 import c4d
 import yaml  # type: ignore[import]
 from qtpy import QtWidgets
@@ -26,6 +28,7 @@ from .assets import AssetIntrospector
 from .data_classes import (
     RenderSubmitterUISettings,
 )
+from .font_utils import scene_has_fonts, get_font_manager_environment
 from .scene import Animation, Scene
 from .style import C4D_STYLE
 from .takes import TakeSelection
@@ -278,6 +281,13 @@ def _get_job_template(
         if "jobEnvironments" not in job_template:
             job_template["jobEnvironments"] = []
         job_template["jobEnvironments"].append(override_environment["environment"])
+
+    # Conditionally add FontManager job environment if fonts are detected
+    if adaptor and scene_has_fonts(Path(Scene.name()).parent):
+        font_manager_environment = get_font_manager_environment()
+        if "jobEnvironments" not in job_template:
+            job_template["jobEnvironments"] = []
+        job_template["jobEnvironments"].append(font_manager_environment)
 
     add_timeouts_to_job_template(job_template, settings.timeouts)
 
@@ -625,7 +635,14 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
         temp_dir: Path to the temporary directory
         asset_references: Asset references to update
     """
+
     doc = c4d.documents.GetActiveDocument()
+
+    # Get the original scene file path BEFORE the temp export
+    # This is crucial because Scene.name() will change after SaveProject
+    original_scene_file_path = Path(Scene.name())
+    original_scene_dir = original_scene_file_path.parent
+    original_fonts_dir = original_scene_dir / "tempFonts"
 
     # Save the project to the temporary directory
     temp_file_path = os.path.join(temp_dir, doc.GetDocumentName())
@@ -642,9 +659,26 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
             "Exporting the scene failed. Please fix all the paths for your assets in your scene in Cinema 4D's Window menu bar > Project Asset Inspector."
         )
 
+    temp_fonts_dir = Path(Scene.name()).parent / "tempFonts"
+
+    # Copy fonts from the original scene's tempFonts directory to the temp directory
+
+    if original_fonts_dir.exists() and original_fonts_dir.is_dir():
+        # Create the tempFonts directory in the temp location
+        temp_fonts_dir.mkdir(exist_ok=True, parents=True)
+
+        # Copy all font files from the original tempFonts directory
+        copied_fonts = []
+        for font_file in original_fonts_dir.iterdir():
+            if font_file.is_file():
+                destination = temp_fonts_dir / font_file.name
+                shutil.copy2(font_file, destination)
+                copied_fonts.append(font_file.name)
+
     # If we get here, save was successful
     # Get all files within the temp directory
     temp_assets = set()
+
     for root, _, files in os.walk(temp_dir):
         for file in files:
             file_path = os.path.join(root, file)
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -668,12 +668,10 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
         temp_fonts_dir.mkdir(exist_ok=True, parents=True)
 
         # Copy all font files from the original tempFonts directory
-        copied_fonts = []
         for font_file in original_fonts_dir.iterdir():
             if font_file.is_file():
                 destination = temp_fonts_dir / font_file.name
                 shutil.copy2(font_file, destination)
-                copied_fonts.append(font_file.name)
 
     # If we get here, save was successful
     # Get all files within the temp directory
--- a/src/deadline/cinema4d_submitter/font_installer.py
+++ b/src/deadline/cinema4d_submitter/font_installer.py
@@ -15,6 +15,8 @@ from typing import Set, Tuple
 _TEMP_FONTS_DIR = "tempFonts"
 
 if sys.platform == "win32":
+    from ctypes import wintypes
+
     try:
         import winreg
     except ImportError:
@@ -27,6 +29,7 @@ else:
     winreg = None  # type: ignore
     user32 = None  # type: ignore
     gdi32 = None  # type: ignore
+    wintypes = None  # type: ignore
 
 FONTS_REG_PATH = r"Software\Microsoft\Windows NT\CurrentVersion\Fonts"
 
@@ -106,12 +109,12 @@ def get_font_name(dst_path: str) -> str:
     fontname = os.path.splitext(filename)[0]
 
     # Try to get the font's real name
-    cb = ctypes.wintypes.DWORD()
+    cb = wintypes.DWORD()
     if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), None, GFRI_DESCRIPTION):
         buf = (ctypes.c_wchar * cb.value)()
         if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), buf, GFRI_DESCRIPTION):
             fontname = buf.value
-    is_truetype = ctypes.wintypes.BOOL()
+    is_truetype = wintypes.BOOL()
     cb.value = ctypes.sizeof(is_truetype)
     gdi32.GetFontResourceInfoW(
         filename, ctypes.byref(cb), ctypes.byref(is_truetype), GFRI_ISTRUETYPE
--- a/src/deadline/cinema4d_submitter/font_installer.py
+++ b/src/deadline/cinema4d_submitter/font_installer.py
@@ -12,6 +12,8 @@ import sys
 import traceback
 from typing import Set, Tuple
 
+_TEMP_FONTS_DIR = "tempFonts"
+
 if sys.platform == "win32":
     try:
         import winreg
@@ -67,13 +69,13 @@ def find_fonts(session_dir: str) -> Set[str]:
         full_sub_dir = None
         for path, dirs, files in os.walk(asset_dir):
             for d in dirs:
-                if "tempFonts" in d:
+                if _TEMP_FONTS_DIR in d:
                     full_sub_dir = os.path.join(path, d)
                     logger.debug(f"tempFonts directory: {full_sub_dir}")
                     break
 
         if not full_sub_dir:
-            logger.debug(f"Couldn't recursively find tempFonts in subfolder: {subfolder}")
+            logger.debug(f"Couldn't recursively find {_TEMP_FONTS_DIR} in subfolder: {subfolder}")
             continue
 
         for file_name in os.listdir(full_sub_dir):
@@ -84,7 +86,7 @@ def find_fonts(session_dir: str) -> Set[str]:
                 fonts.add(full_assetpath)
             else:
                 logger.warning(
-                    f"A file that is not a supported font was found in the tempFonts folder: {full_assetpath}"
+                    f"A file that is not a supported font was found in the {_TEMP_FONTS_DIR} folder: {full_assetpath}"
                 )
     return fonts
 
@@ -100,26 +102,22 @@ def get_font_name(dst_path: str) -> str:
     if sys.platform != "win32":
         raise RuntimeError("Font installation is only supported on Windows")
 
-    try:
-        filename = os.path.basename(dst_path)
-        fontname = os.path.splitext(filename)[0]
+    filename = os.path.basename(dst_path)
+    fontname = os.path.splitext(filename)[0]
 
-        # Try to get the font's real name
-        cb = ctypes.wintypes.DWORD()
-        if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), None, GFRI_DESCRIPTION):
-            buf = (ctypes.c_wchar * cb.value)()
-            if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), buf, GFRI_DESCRIPTION):
-                fontname = buf.value
-        is_truetype = ctypes.wintypes.BOOL()
-        cb.value = ctypes.sizeof(is_truetype)
-        gdi32.GetFontResourceInfoW(
-            filename, ctypes.byref(cb), ctypes.byref(is_truetype), GFRI_ISTRUETYPE
-        )
-        if is_truetype:
-            fontname += " (TrueType)"
-
-    except Exception as e:
-        raise e
+    # Try to get the font's real name
+    cb = ctypes.wintypes.DWORD()
+    if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), None, GFRI_DESCRIPTION):
+        buf = (ctypes.c_wchar * cb.value)()
+        if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), buf, GFRI_DESCRIPTION):
+            fontname = buf.value
+    is_truetype = ctypes.wintypes.BOOL()
+    cb.value = ctypes.sizeof(is_truetype)
+    gdi32.GetFontResourceInfoW(
+        filename, ctypes.byref(cb), ctypes.byref(is_truetype), GFRI_ISTRUETYPE
+    )
+    if is_truetype:
+        fontname += " (TrueType)"
 
     return fontname
 
@@ -230,8 +228,7 @@ def _install_fonts(session_dir: str) -> None:
     fonts = find_fonts(session_dir)
 
     if not fonts:
-        logger.info("No custom fonts found, continuing task...")
-        return
+        raise RuntimeError("No custom fonts found")
     for font in fonts:
         logger.info("Installing font: " + font)
         installed, msg = install_font(font)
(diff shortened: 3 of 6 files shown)


src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
...
+
+    :param session_dir: the root folder in which to look for files
+
+    :returns: a set with all found fonts
+    """
+    fonts = set()
+    for subfolder in os.listdir(session_dir):
+        # Only look in assetroot folders
+        if not subfolder.startswith("assetroot-"):
+            continue
```
crowecawcaw, line 69: Is the `assetroot-` prefix a stable part of the session? Seems like it might change over time. Is there another way we can find these directories?

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
karthikbekalp, line 1: This file is identical to the installation script from After effects: https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/font_manager.py

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
...
+)
+
+# Font extensions supported in gdi32.AddFontResourceW
+# OpenType fonts without an extension can also be installed (e.g. Adobe Fonts)
+FONT_EXTENSIONS = [".otf", ".ttf", ".fon", ""]
+
+logger = logging.getLogger(__name__)
+
+
+def find_fonts(session_dir: str) -> Set[str]:
```
crowecawcaw, line 57: Could we simplify this function to just look for font files instead of searching for special directories?

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
...
+        is_truetype = wintypes.BOOL()
+        cb.value = ctypes.sizeof(is_truetype)
+        gdi32.GetFontResourceInfoW(
+            filename, ctypes.byref(cb), ctypes.byref(is_truetype), GFRI_ISTRUETYPE
+        )
+        if is_truetype:
+            fontname += " (TrueType)"
+
+    except Exception as e:
+        raise e
```
crowecawcaw: Does nothing. Better to remove the try-catch

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
...
+            if not os.path.exists(FONT_LOCATION_USER):
+                logger.info(f"Creating User Fonts folder: {FONT_LOCATION_USER}")
+                os.makedirs(FONT_LOCATION_USER)
+
+            dst_dir = FONT_LOCATION_USER
+            registry_scope = winreg.HKEY_CURRENT_USER
+        dst_path = os.path.join(dst_dir, os.path.basename(src_path))
+
+        # Copy the font to the Windows Fonts folder
+        shutil.copy(src_path, dst_path)
```
crowecawcaw, line 160: Is there a way to use font locally without installing them on the worker? I see we clean up the fonts afterwards which is probably ok. But even better if we don't have to mess with the underlying worker at all.

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
...
+    Calls all needed functions for installing fonts
+
+    :param session_dir: directory of the session
+    """
+    logger.info("Looking for fonts to install...")
+    fonts = find_fonts(session_dir)
+
+    if not fonts:
+        logger.info("No custom fonts found, continuing task...")
+        return
```
crowecawcaw: Do we need this log line? If not, we can drop the block

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
...
+    for font in fonts:
+        logger.info("Installing font: " + font)
+        installed, msg = install_font(font)
+        if not installed:
+            raise RuntimeError(f"Error installing font: {msg}")
```
crowecawcaw, line 244: Probably cleaner if `install_font` raises the RuntimeError directly. Same with uninstall_font

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+        file_path (str): Path to the file
+
+    Returns:
+        bool: True if the file is a font file, False otherwise
+    """
+    if not file_path:
+        return False
+
+    # Adobe fonts do not have an extension.
+    font_extensions = [".otf", ".ttf", ".fon", ""]
```
crowecawcaw: Does this treat all extension-less files as fonts? Seems like it would break something or attach the wrong files

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+        raise ValueError("Font name cannot be empty")
+
+    if not scene_location or not scene_location.exists():
+        raise ValueError(f"Scene location does not exist: {scene_location}")
+
+    # Get the font's location
+    font_location = get_font_location(font_name.strip())
+
+    if font_location is None:
+        raise FontNotFoundError(f"Font '{font_name}' not found in system font directories")
```
crowecawcaw: Could this just be a RuntimeError if we're not handling it differently?

crowecawcaw (changes requested):

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,286 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Utility functions for the handling of fonts
+"""
+
+import ctypes
```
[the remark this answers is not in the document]
karthikbekalp: Updated and removed the extra import.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Good question, how about I do this as a follow up? Trying to unblock a customer who is hitting this issue and if its not urgent I'd like to do this in a follow up. :)

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: That's a good question. I don't know currently but can I fix it later as a follow up?

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Makes sense. Removed.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Unfortunately not. I spoke with Maxon support (the company behind Cinema 4D) about this one and looks like the only way to get fonts working on render node is by installing it. :'(

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+        font_dirs = [os.path.join(windir, "Fonts")]
+
+        # User-installed fonts
+        if localappdata:
+            user_fonts_dir = os.path.join(localappdata, "Microsoft", "Windows", "Fonts")
+            font_dirs.append(user_fonts_dir)
+
+        # Adobe fonts
+        if appdata:
+            adobe_livetype = os.path.join(appdata, "Adobe", "CoreSync", "plugins", "livetype")
```
joel-wong-aws: How did you get this path? From the online sources I can find, this should be
`adobe_livetype = os.path.join(appdata, "Adobe", "CoreSync", "plugins", "livetype", "r")`

https://community.adobe.com/t5/substance-3d-designer-discussions/some-adobe-fonts-not-showing-up-in-text-node/td-p/12416316
https://stackoverflow.com/questions/57258711/where-does-adobe-fonts-aka-typkekit-aka-creative-cloud-fonts-store-fonts-on-my

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+        # macOS font directories
+        font_dirs = [
+            "/Library/Fonts",
+            "/System/Library/Fonts",
+            os.path.expanduser("~/Library/Fonts"),
+        ]
+
+        # Adobe fonts on macOS
+        adobe_livetype = os.path.expanduser(
+            "~/Library/Application Support/Adobe/CoreSync/plugins/livetype"
```
joel-wong-aws: Similar here, the source I see show that this should be
`"~/Library/Application Support/Adobe/CoreSync/plugins/livetype/.r"

https://stackoverflow.com/questions/57258711/where-does-adobe-fonts-aka-typkekit-aka-creative-cloud-fonts-store-fonts-on-my
https://github.com/Kalaschnik/adobe-fonts-revealer

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+    # Search for the font in all font directories
+    for font_dir in font_dirs:
+        try:
+            for root, _, files in os.walk(font_dir):
+                for file in files:
+                    # Check if the font name is in the file name (case-insensitive)
+                    if font_name.lower() in file.lower():
+                        file_path = os.path.join(root, file)
+                        # Check if it's a font file by extension and exists
+                        if is_font_file(file_path) and os.path.isfile(file_path):
```
joel-wong-aws: nit: can we move the `os.path.isfile` check into `is_font_file`?

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: > Do we need this log line?

I think the log line is useful if for some reason the worker is unable to find fonts. But I wonder if we should just raise an error instead because a job environment should not have been added to install fonts if there was no fonts to install.

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+    owner: Optional[c4d.BaseList2D] = asset.get("owner", None)
+    if pid == c4d.NOTOK or owner is None:
+        return False
+    # We check the type of the parameter value.
+    # If it is a FontData, we assume it is a font asset.
+    value = owner.GetParameter(pid, c4d.DESCFLAGS_GET_NONE)
+    return isinstance(value, c4d.FontData)
+
+
+def get_font_location(font_name: str) -> Optional[str]:
```
joel-wong-aws, line 107: supernit: this could be more efficient if we passed in multiple font names, traversed the font directories in one go, and then copied them in a batch.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Makes sense. How about I add this part of my follow up while we unblock the customer?

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+    # Get system font directories based on the operating system
+    font_dirs = get_system_font_directories()
+
+    # Search for the font in all font directories
+    for font_dir in font_dirs:
+        try:
+            for root, _, files in os.walk(font_dir):
+                for file in files:
+                    # Check if the font name is in the file name (case-insensitive)
+                    if font_name.lower() in file.lower():
```
joel-wong-aws: In the case where we use a non-regular font (e.g. `SegoeUI` `Bold`) should we copy all versions of a font name to the job?

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: It is possible but unfortunately there is no easy way to fix this though. We are only checking in font directories so it should be pretty unlikely to find a file with the same name as the font no? 

One of our customers is using Adobe fonts so we cannot skip texts without fonts.

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -22,15 +23,36 @@ def parse_scene_assets(self) -> set[Path]:
...
+            if is_asset_a_font(asset):
+                copy_font_to_scene_folder(asset["assetname"], path_to_scene_file_dir)
+
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
             if exists is True and filename is not None:
                 assets.add(Path(filename))
 
+        # Add all font files from the tempFonts directory to assets
+        fonts_dir = path_to_scene_file_dir / "tempFonts"
```
joel-wong-aws: Can we make `tempFonts` a constant?

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: I added a separate error in this case so that its clear that the user has not installed the fonts present in the scene. I hit an issue before where I opened a scene with fonts not installed in my workstation and hence added this error.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -635,9 +652,26 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
...
+    temp_fonts_dir = Path(Scene.name()).parent / "tempFonts"
+
+    # Copy fonts from the original scene's tempFonts directory to the temp directory
+
+    if original_fonts_dir.exists() and original_fonts_dir.is_dir():
+        # Create the tempFonts directory in the temp location
+        temp_fonts_dir.mkdir(exist_ok=True, parents=True)
+
+        # Copy all font files from the original tempFonts directory
+        copied_fonts = []
```
joel-wong-aws: Unused variable `copied_fonts`

src/deadline/cinema4d_submitter/font_utils.py
joel-wong-aws: +1, the empty line ending would likely match anything based on the logic below

karthikbekalp (commented):
Thanks @crowecawcaw  for the review. I'll try to punt any non-trivial work for my follow up so that we can quickly unblock our customers for now.

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: I used the font locations from our [AE repo](https://github.com/aws-deadline/deadline-cloud-for-after-effects/blob/mainline/dist/DeadlineCloudSubmitter_Assets/JobTemplate/scripts/get_user_fonts.py#L30C1-L30C49)

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: Sure, updating.

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: I propose to add it to the follow up PR so that we can unblock our customer. Is it ok?

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,283 @@
...
+    try:
+        filename = os.path.basename(dst_path)
+        fontname = os.path.splitext(filename)[0]
+
+        # Try to get the font's real name
+        cb = ctypes.wintypes.DWORD()
+        if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), None, GFRI_DESCRIPTION):
+            buf = (ctypes.c_wchar * cb.value)()
+            if gdi32.GetFontResourceInfoW(filename, ctypes.byref(cb), buf, GFRI_DESCRIPTION):
+                fontname = buf.value
```
joel-wong-aws: What happens if this is unsuccessful? What happens to `fontname`? Do we need any logging if the conditionals are false?

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: Hmm, interesting. We definitely can. Are we good with adding it as part of follow up?

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # Load the font in the current session, remove font when loading fails
+        if not gdi32.AddFontResourceW(dst_path):
+            os.remove(dst_path)
+            raise OSError(f'AddFontResource failed to load "{src_path}"')
+
+        # Notify running programs
+        user32.SendMessageTimeoutW(
+            HWND_BROADCAST, WM_FONTCHANGE, 0, 0, SMTO_ABORTIFHUNG, 1000, None
+        )
```
joel-wong-aws, line 170: Can we only do this once after all fonts are installed?

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,283 @@
...
+    """
+    logger.info("Looking for fonts to install...")
+    fonts = find_fonts(session_dir)
+
+    if not fonts:
+        logger.info("No custom fonts found, continuing task...")
+        return
+    for font in fonts:
+        logger.info("Installing font: " + font)
+        installed, msg = install_font(font)
```
leon-li-inspire, line 242: Is it possible this method would return False if the font has been already installed by some reasons ? Asking to ensure we don't raise RuntimeError below in this case.

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -0,0 +1,283 @@
...
+            os.remove(dst_path)
+            raise OSError(f'RemoveFontResourceW failed to load "{src_path}"')
+
+        if os.path.exists(dst_path):
+            os.remove(dst_path)
+
+        # Notify running programs
+        user32.SendMessageTimeoutW(
+            HWND_BROADCAST, WM_FONTCHANGE, 0, 0, SMTO_ABORTIFHUNG, 1000, None
+        )
```
joel-wong-aws, line 223: Similar note as above, can we only do this once all fonts are uninstalled?

joel-wong-aws (changes requested):

src/deadline/cinema4d_submitter/assets.py
karthikbekalp: Sure, will do.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Sure, updated.

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,260 @@
...
+        dict[str, Any]: The FontManager job environment configuration
+    """
+    # Read the font installer script from file
+    font_installer_path = Path(__file__).parent / "font_installer.py"
+    with open(font_installer_path, "r", encoding="utf-8") as f:
+        font_installer_script = f.read()
+
+    return {
+        "name": "FontManager",
+        "description": "Manages font installation and cleanup for Cinema4D rendering.",
```
leon-li-inspire: nit: Do we need to explicitly mention this environment was injected by the submitter itself ?

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Hmm, its unlikely as the fonts are installed in the user directory that is only available until the current session. We would delete it later on.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: How about add this into a follow up? The reason I ask this is because this file is identical to AE's script and I want to unblock our customer soon.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: How about add this into a follow up? The reason I ask this is because this file is identical to AE's script and I want to unblock our customer soon.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: How about add this into a follow up? The reason I ask this is because this file is identical to AE's script and I want to unblock our customer soon.

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: Sure. I'll add it.

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: True. How about add this into a follow up? The reason I ask this is because this file is identical to AE's script and I want to unblock our customer soon.

leon-li-inspire (dismissed):

src/deadline/cinema4d_submitter/font_installer.py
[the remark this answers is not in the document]
karthikbekalp: I had to add this back because I started to get `AttributeError: module 'ctypes' has no attribute 'wintypes'` errors which leads to failures for the very first run.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: I thought about this more deeply and I think my assumptions were wrong. 

The location of font installation is "C:\\Users\\job-user\\AppData\\Local\\Microsoft\\Windows\\Fonts" which means it exists for all the job submissions and not only for specific sessions. I'll fix this.

joel-wong-aws (dismissed):

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,446 @@
...
+                    if not is_font_file(file_path):
+                        continue
+
+                    # Check if this font matches
+                    if _is_font_match(font_name, file, file_path):
+                        logger.debug(f"Font match found: {file_path}")
+                        return file_path
+
+        except (PermissionError, OSError) as e:
+            logger.debug(f"Cannot access font directory {font_dir}: {e}")
```
crowecawcaw, line 139: There are a lot of places where we log a debug message if there's an exception. How do these debug messages show up to users? If there's an error and a font isn't attached, is there a way for a user to figure out why?

karthikbekalp: Discussed this offline. The errors would appear in Cinema4D's console and not in the console. I think its nice to have for now but we can always remove in the future if its not useful.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -625,8 +635,15 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
...
         asset_references: Asset references to update
     """
+
     doc = c4d.documents.GetActiveDocument()
 
+    # Get the original scene file path BEFORE the temp export
+    # This is crucial because Scene.name() will change after SaveProject
+    original_scene_file_path = Path(Scene.name())
+    original_scene_dir = original_scene_file_path.parent
+    original_fonts_dir = original_scene_dir / "tempFonts"
```
joel-wong-aws, line 645: nit: this reference to tempFonts can be replaced with a constant

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -642,9 +659,24 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
             "Exporting the scene failed. Please fix all the paths for your assets in your scene in Cinema 4D's Window menu bar > Project Asset Inspector."
         )
 
+    temp_fonts_dir = Path(Scene.name()).parent / "tempFonts"
```
joel-wong-aws, line 662: nit: this reference to tempFonts can be replaced with a constant

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,446 @@
...
+import platform
+import shutil
+from pathlib import Path
+from typing import Optional, List, Any, Dict
+
+# Import fontTools for robust font metadata extraction
+# for detecting Adobe fonts.
+from fontTools import ttLib
+
+TEMP_FONTS_DIR = "tempFonts"
```
joel-wong-aws, line 15: nit: this constant is defined in two places

joel-wong-aws (dismissed):

joel-wong-aws (approved):

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: Yup, we need a separate declaration in `font_installer` because the submitter bundles the file as a script directly to be run as a job environment.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Makes sense. I'll remember to remove this as part of the followup.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Makes sense. I'll remember to remove this as part of the followup.

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -0,0 +1,446 @@
...
+
+    if not font_path or not os.path.isfile(font_path):
+        return None
+
+    try:
+        font = ttLib.TTFont(font_path)
+        names_table = font["name"].names
+
+        # Extract key font information
+        metadata: dict[str, Any] = {
```
crowecawcaw, line 46: Follow up: would be better to define a dataclass for this

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Updated this in the commit: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/283/commits/23c61362c1cc8b525a77c5836caf6ee38e5f9824 to ignore font installation if the font already exists.

crowecawcaw (approved):

PR #294, opened 2025-08-21
fix: Only bundle fonts on Windows as its not supported on Mac.

### LINKED ISSUE #292: Bug: Text objects while submitting on a Mac does not render correctly.
### Describe the bug

We have encountered issues while attempting to render scenes with text with fonts in Cinema 4D on a machine different from the one where the project was created. 

The output when rendered on a different machine does not work render correctly and it defaults to a base font. 

### Expected Behaviour

The fonts are rendered correctly on the machine. 

### Current Behaviour

Cinema 4D just uses default fonts than using the fonts that were bundled. 

### Reproduction Steps

1. Try creating a scene with a Text object and select a font for it. 
2. Submit the job using the submitter. 

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): Only on Mac
2. Version of Cinema 4D: Any
3. Version of this package: Any
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?: Any

karthikbekalp:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/292

Fonts does not work when submitted from MacOS.

I contacted Maxon support and they suggested that we should just use instruct customers to make their text objects editable and submit as that's their recommended option for us. 
So I updated the code to do just that. 

I also added the code that I was working on [font's PR](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/283) followup into this PR. 

I'm also working on unit tests for this feature. But I guess its best to add that into a follow up PR and not make this PR even bigger than it already is. 

Better visibility and only runs the font bundling logic in Windows. Also, updated our docs with this info. 

Ran few manual tests to confirm that no font bundling happens. 

Yes

Yes ( I had to modify the versions in parameter_values.yaml to make it work, I'll have a PR for that shortly. )

```

test/integ/test_cinema4d.py::test_integ[physical]
[gw0] [ 14%] PASSED test/integ/test_cinema4d.py::test_integ[physical] 
test/integ/test_cinema4d.py::test_integ[physical_textured]
[gw0] [ 28%] PASSED test/integ/test_cinema4d.py::test_integ[physical_textured] 
test/integ/test_cinema4d.py::test_integ[redshift]
[gw0] [ 42%] PASSED test/integ/test_cinema4d.py::test_integ[redshift] 
test/integ/test_cinema4d.py::test_integ[redshift_takes]
[gw0] [ 57%] PASSED test/integ/test_cinema4d.py::test_integ[redshift_takes] 
test/integ/test_cinema4d.py::test_integ[redshift_textured]
[gw0] [ 71%] PASSED test/integ/test_cinema4d.py::test_integ[redshift_textured] 
test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
[gw0] [ 85%] PASSED test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters] 
test/integ/test_cinema4d.py::test_integ[physical_multi_takes]
[gw0] [100%] PASSED test/integ/test_cinema4d.py::test_integ[physical_multi_takes] 

======================================================================== slowest 5 durations ======================================================================== 
475.01s call     test/integ/test_cinema4d.py::test_integ[physical_multi_takes]
109.75s call     test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters]
87.64s call     test/integ/test_cinema4d.py::test_integ[redshift_textured]
85.93s call     test/integ/test_cinema4d.py::test_integ[redshift]
80.12s call     test/integ/test_cinema4d.py::test_integ[redshift_takes]
=================================================================== 7 passed in 962.10s (0:16:02) ===================================================================
```

Yes and it does not require any changes to the test setup. 

Yes

No. 

----

### DIFF
--- a/src/deadline/cinema4d_submitter/font_installer.py
+++ b/src/deadline/cinema4d_submitter/font_installer.py
@@ -14,23 +14,6 @@ from typing import Set
 
 FONTS_DIR = "fonts"
 
-if sys.platform == "win32":
-    from ctypes import wintypes
-
-    try:
-        import winreg
-    except ImportError:
-        import _winreg as winreg  # type: ignore
-
-    user32 = ctypes.WinDLL("user32", use_last_error=True)
-    gdi32 = ctypes.WinDLL("gdi32", use_last_error=True)
-else:
-    # Stub for non-Windows platforms
-    winreg = None  # type: ignore
-    user32 = None  # type: ignore
-    gdi32 = None  # type: ignore
-    wintypes = None  # type: ignore
-
 FONTS_REG_PATH = r"Software\Microsoft\Windows NT\CurrentVersion\Fonts"
 
 HWND_BROADCAST = 0xFFFF
@@ -52,6 +35,16 @@ FONT_LOCATION_USER = os.path.join(
 FONT_EXTENSIONS = [".otf", ".ttf", ".fon", ""]
 
 logger = logging.getLogger(__name__)
+
+
+def is_windows() -> bool:
+    """
+    Check if the current platform is Windows.
+
+    Returns:
+        bool: True if running on Windows, False otherwise
+    """
+    return sys.platform == "win32"
 
 
 def _collect_fonts_from_directory(fonts_dir: str) -> Set[str]:
@@ -186,8 +179,13 @@ def get_font_name(dst_path: str) -> str:
 
     :returns: string with the font's name
     """
-    if sys.platform != "win32":
+    if not is_windows():
         raise RuntimeError("Font installation is only supported on Windows")
+
+    # Import Windows-specific modules only when needed
+    from ctypes import wintypes
+
+    gdi32 = ctypes.WinDLL("gdi32", use_last_error=True)  # type: ignore[attr-defined]
 
     filename = os.path.basename(dst_path)
     fontname = os.path.splitext(filename)[0]
@@ -218,15 +216,23 @@ def install_font(src_path: str, scope: str = INSTALL_SCOPE_USER) -> None:
 
     :raises RuntimeError: if font installation fails
     """
-    if sys.platform != "win32":
+    if not is_windows():
         logger.error("Font installation is only supported on Windows")
         return
+
+    # Import Windows-specific modules only when needed
+    try:
+        import winreg
+    except ImportError:
+        import _winreg as winreg  # type: ignore
+
+    gdi32 = ctypes.WinDLL("gdi32", use_last_error=True)  # type: ignore[attr-defined]
 
     try:
         # Determine font destination
         if scope == INSTALL_SCOPE_SYSTEM:
             dst_dir = FONT_LOCATION_SYSTEM
-            registry_scope = winreg.HKEY_LOCAL_MACHINE
+            registry_scope = winreg.HKEY_LOCAL_MACHINE  # type: ignore[attr-defined]
         else:
             # Check if the Fonts folder exists, create it if it doesn't
             if not os.path.exists(FONT_LOCATION_USER):
@@ -234,7 +240,7 @@ def install_font(src_path: str, scope: str = INSTALL_SCOPE_USER) -> None:
                 os.makedirs(FONT_LOCATION_USER)
 
             dst_dir = FONT_LOCATION_USER
-            registry_scope = winreg.HKEY_CURRENT_USER
+            registry_scope = winreg.HKEY_CURRENT_USER  # type: ignore[attr-defined]
         dst_path = os.path.join(dst_dir, os.path.basename(src_path))
 
         # Check if font already exists at destination
@@ -255,10 +261,10 @@ def install_font(src_path: str, scope: str = INSTALL_SCOPE_USER) -> None:
         fontname = get_font_name(dst_path)
 
         # Creates registry if it doesn't exist, opens when it does exist
-        with winreg.CreateKeyEx(
-            registry_scope, FONTS_REG_PATH, 0, access=winreg.KEY_SET_VALUE
+        with winreg.CreateKeyEx(  # type: ignore[attr-defined]
+            registry_scope, FONTS_REG_PATH, 0, access=winreg.KEY_SET_VALUE  # type: ignore[attr-defined]
         ) as key:
-            winreg.SetValueEx(key, fontname, 0, winreg.REG_SZ, filename)
+            winreg.SetValueEx(key, fontname, 0, winreg.REG_SZ, filename)  # type: ignore[attr-defined]
     except Exception as e:
         raise RuntimeError(f"Failed to install font '{src_path}': {str(e)}") from e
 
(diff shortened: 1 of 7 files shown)


src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -69,69 +68,61 @@ def find_fonts(session_dir: str, scene_file_path: str) -> Set[str]:
...
-                    f"A file that is not a supported font was found in the {_TEMP_FONTS_DIR} folder: {full_assetpath}"
-                )
-
-    # If fonts were found in asset root, return them
+    # Approach 1: Recursive search for fonts directories in session_dir
+    logger.debug("Using recursive font search approach")
+    try:
+        for root, dirs, files in os.walk(session_dir):
+            for dir_name in dirs:
+                if dir_name == FONTS_DIR:
```
joel-wong-aws: nit: can we use a guard clause here instead of a (further) indented `if` statement? All the nesting makes it harder to follow

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -69,69 +68,61 @@ def find_fonts(session_dir: str, scene_file_path: str) -> Set[str]:
...
+                    fonts_dir = os.path.join(root, dir_name)
+                    logger.debug(f"Found {FONTS_DIR} directory: {fonts_dir}")
+
+                    try:
+                        for file_name in os.listdir(fonts_dir):
+                            full_path = os.path.join(fonts_dir, file_name)
+                            if os.path.isfile(full_path):
```
joel-wong-aws: Same note here about guard clauses

joel-wong-aws (dismissed):

crowecawcaw (dismissed):
Is there an easy way to do the OS check in a single spot instead of checking in multiple places? We could also import files in code and not at the top of a file if it helps clean up the code.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Sure, updating it.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Sure.

karthikbekalp:
> Is there an easy way to do the OS check in a single spot instead of checking in multiple places? We could also import files in code and not at the top of a file if it helps clean up the code.

I think that's a good idea. I'll refactor.

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -55,88 +37,138 @@
...
+            for dir_name in dirs:
+                # Skip directories that are not fonts directories
+                if dir_name != FONTS_DIR:
+                    continue
+
+                fonts_dir = os.path.join(root, dir_name)
+                logger.info(f"Found {FONTS_DIR} directory: {fonts_dir}")
+
+                # Skip system fonts directories (those under .env directories)
+                if os.sep + ".env" + os.sep in fonts_dir:
```
joel-wong-aws, line 101: Are fonts ever directly under a folder `\.env`? e.g. `\.env\ArialRegular`? If so, then this condition wouldn't be false in that scenario, right?

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -55,88 +37,138 @@
...
+                    logger.info(f"Skipping system fonts directory under .env: {fonts_dir}")
```
joel-wong-aws, line 102: Is it possible for the system fonts on one machine to be different than the system fonts on a different machine? Would it be incorrectly skipped in that case?

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: We wanted to ignore all the fonts under \.env\ as this folder generally includes Cinema 4D system files which can also have fonts in it. Hence, I added this check to skip Cinema 4D fonts directories and in general any installation directories.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: The contents of the `.env` folder within the session directory is basically the installation files from the DCC i.e. Cinema 4D in this case.

joel-wong-aws (approved):

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -24,62 +22,83 @@
...
-            "postscript_name": None,
-            "raw_names": {},
-        }
+        # Create FontMetadata instance
+        metadata = FontMetadata(file_path=font_path)
 
         # Build raw names table for debugging
         try:
-            metadata["raw_names"] = {i: str(names_table[i]) for i in range(len(names_table))}
+            metadata.raw_names = {i: str(names_table[i]) for i in range(len(names_table))}
```
crowecawcaw, line 76: Is `dict[int, str]` the best type for this? Do we just need a list of raw names?

src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -162,19 +181,20 @@ def _is_font_match(font_name: str, filename: str, file_path: str) -> bool:
...
+            if font_name_lower == family_name_lower or font_name_lower in family_name_lower:
+                return True
 
         # Check full name match
-        full_name = metadata.get("full_name", "").lower()
-        if full_name and (font_name_lower == full_name or font_name_lower in full_name):
-            return True
+        if metadata.full_name:
+            full_name_lower = metadata.full_name.lower()
+            if font_name_lower == full_name_lower or font_name_lower in full_name_lower:
```
crowecawcaw, line 196: Nit: equality check is redundant

src/deadline/cinema4d_submitter/platform_utils.py
```diff
@@ -0,0 +1,27 @@
...
+import sys
+
+
+def is_windows() -> bool:
+    """
+    Check if the current platform is Windows.
+
+    Returns:
+        bool: True if running on Windows, False otherwise
+    """
```
crowecawcaw, line 16: Nit: these comments are noisy, probably clearer without them

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -55,88 +37,138 @@
...
-    if fonts:
-        logger.debug(f"Found {len(fonts)} fonts in asset root directories")
-        return fonts
+    except (OSError, PermissionError) as e:
+        logger.warning(f"Could not access fonts directory {fonts_dir}: {e}")
+
+    return fonts
+
+
+def _find_fonts_recursive(session_dir: str) -> Set[str]:
```
crowecawcaw, line 80: I think a lot of this code and `_collect_fonts_from_directory` could be cleaned up by using glob instead: https://docs.python.org/3/library/glob.html

src/deadline/cinema4d_submitter/font_installer.py
```diff
@@ -55,88 +37,138 @@
...
+    logger.debug(f"Scene file path provided: {scene_file_path}")
+
+    # Approach 1: Recursive search for fonts directories in session_dir
+    fonts = _find_fonts_recursive(session_dir)
+    if fonts:
+        return fonts
+
+    # Approach 2: Scene file path approach as fallback
+    logger.debug("No fonts found via recursive search, trying scene-based approach")
+    return _find_fonts_scene_based(scene_file_path)
```
crowecawcaw, line 171: Would it be safer to always do both? In which scenarios would the user want one or the other?

leon-li-inspire (approved):

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: Hmm, good point. But now that I look closely at the variable, it looks like we don't ever use that field and I think I'll go ahead and remove it.

src/deadline/cinema4d_submitter/font_utils.py
karthikbekalp: Makes sense, I'll remove it.

src/deadline/cinema4d_submitter/platform_utils.py
karthikbekalp: Noted, I'll remove it.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: I think its a good idea. I'll fix this in the followup PR.

karthikbekalp:
Hmm, looks like this PR got merged. I'll work on the unresolved comments from this PR in my upcoming PR.

src/deadline/cinema4d_submitter/font_installer.py
karthikbekalp: Hmm, I looked into using `glob` but I think in this particular scenario it was limited and I couldn't use its full potential due to:
- It didn't have any negative clauses that could help me skip the folders under `.env`
- Adobe fonts don't have any extensions, which limits glob's functionality

However, I will update the code to use glob to look for folders as it is more performant than using just os.walk.

PR #324, opened 2025-10-31
fix: Allow other users to access dependencies installed by system installation

### LINKED ISSUE #322: Bug: Deadline Cloud Submitter fails to launch as standard user
### Describe the bug

When launching the AWS Deadline Cloud Submitter in Cinema 4D as a standard (non-admin) user on Windows, the submitter fails to launch with `QtBindingsNotFoundError` error.

When prompted to install the GUI components and font tools packages as standard user, the console shows the packages already exist in 

C:\Program Files\DeadlineCloudSubmitter\Submitters\Cinema4D\ . Specify --upgrade to force replacement

Image

### Expected Behaviour

Once the submitter is installed as Admin and with all the additional GUI Components and font tools packages, it is expected to launch the submitter directly without any issue or having to install the GUI, font tools packages again as standard user.

### Current Behaviour

Fails to launch the submitter with below error in python console in Cinema 4D:

`Traceback (most recent call last):
  File "C:\Program Files\DeadlineCloudSubmitter\Submitters\Cinema4D\cinema_4d_plugins\DeadlineCloud.pyp", line 64, in Execute
    self._import_and_show_submitter()
  File "C:\Program Files\DeadlineCloudSubmitter\Submitters\Cinema4D\cinema_4d_plugins\DeadlineCloud.pyp", line 48, in _import_and_show_submitter
  File "C:\Program Files\DeadlineCloudSubmitter\Submitters\Cinema4D\deadline\cinema4d_submitter\__init__.py", line 111, in <module>
    from .cinema4d_render_submitter import show_submitter  # noqa: E402
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\DeadlineCloudSubmitter\Submitters\Cinema4D\deadline\cinema4d_submitter\cinema4d_render_submitter.py", line 14, in <module>
    from qtpy import QtWidgets
  File "C:\Program Files\DeadlineCloudSubmitter\Submitters\Cinema4D\qtpy\__init__.py", line 293, in <module>
    raise QtBindingsNotFoundError from None
qtpy.QtBindingsNotFoundError: No Qt bindings could be found`

Image

### Reproduction Steps

On Windows:
1. Install Cinema4D 2025.3.3 as Admin
2. Install Deadline Cloud Submitter as Admin → for All Users.
3. Select Deadline Cloud for Cinema 4D 2024 - 2025 and proceed with installation.
4. Install the GUI components & font tools prompted when launching the submitter.
5. Submitter launches successfully as Admin user.
6. Log in as the standard user (non-admin).
7. Launch Cinema 4D.
8. Open a scene file and launch the AWS Deadline Cloud Submitter.
9. Prompts to install the GUI Components and Font tools again.
10. Observe that the submitter fails to launch and throws an error in the python console.

### Environment

Tested On:
1. Window Server 2025, Windows 11 23H2
2. Cinema 4D 2025.3.3
3. DeadlineCloudSubmitter-2025.10.22.0-windows-x64-installer

karthikbekalp:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/322

The system installation failed to work for other users even when run as an Administrator with system install. 

Added logic to add read and execute permissions so that other users can use the dependencies installed. 

Other users when installed as a system install will be able to use the plugin. 

Ran it manually on an EC2 and confirmed that it worked. 

Yes, and also added some. 

Skipped as this code is not covered in integration tests and we want to get this fix in ASAP. 

Yes, but it does not have an impact on the tests. 

This was a bug that is fixed. 

No

If so, then please describe the changes that users of this package must make to update their scripts, or Python applications. Also,
please ensure that the title of your commit follows our conventional commit guidelines in 
[CONTRIBUTING.md](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/CONTRIBUTING.md#conventional-commits) for breaking changes.
*delete text ending here*

----

### DIFF
--- a/src/deadline/cinema4d_submitter/__init__.py
+++ b/src/deadline/cinema4d_submitter/__init__.py
@@ -6,9 +6,79 @@ from pathlib import Path
 
 import c4d
 
-from .platform_utils import is_macos
+from .platform_utils import is_macos, is_windows
 
 _logger = _logging.getLogger(__name__)
+
+
+def _has_windows_admin_privileges() -> bool:
+    """
+    Determine if the current process is running with administrator privileges on Windows.
+
+    Returns:
+        True if running as administrator on Windows, False otherwise
+    """
+    # Early return if not running on Windows
+    if not is_windows():
+        return False
+
+    try:
+        import ctypes
+
+        # Use Windows API to check if the current process has admin privileges
+        # IsUserAnAdmin() returns non-zero if the user is an admin, 0 otherwise
+        return ctypes.windll.shell32.IsUserAnAdmin() != 0  # type: ignore[attr-defined]
+    except Exception:
+        # If any exception occurs (e.g., attribute error, access denied),
+        # assume not running as admin and return False
+        return False
+
+
+def _apply_windows_read_execute_permissions_for_all_users(directory: Path) -> None:
+    """
+    Apply read and execute permissions to all users for a directory on Windows.
+
+    This function uses the Windows icacls command to grant permissions following
+    the principle of least privilege. Users receive only the minimum permissions
+    needed to access and use installed Python packages.
+
+    Permission flags explained:
+    - Users: Applies to the built-in Users group (all non-admin users)
+    - (OI) Object Inherit: Files created in the directory inherit these permissions
+    - (CI) Container Inherit: Subdirectories created inherit these permissions
+    - (RX) Read and Execute: Minimum privileges needed to access and run Python packages
+      - R (Read): Users can read Python files and dependencies
+      - X (Execute): Users can execute Python scripts and binaries
+      - Does NOT include Write, Modify, or Full Control permissions
+    - /T: Apply recursively to all existing subdirectories and files
+
+    Args:
+        directory: The directory path to apply permissions to
+    """
+    # If not running as admin early return
+    if not _has_windows_admin_privileges():
+        return
+
+    # Construct icacls command with appropriate parameters
+    # Grant Users group Read and Execute permissions with inheritance
+    icacls_command = ["icacls", str(directory), "/grant", "Users:(OI)(CI)(RX)", "/T"]
+
+    try:
+        # Execute the command without raising exception on non-zero exit
+        result = subprocess.run(icacls_command, check=False, capture_output=True, text=True)
+
+        if result.returncode == 0:
+            _logger.info(f"Successfully applied Windows permissions to {directory}")
+        else:
+            _logger.warning(
+                f"Failed to apply Windows permissions to {directory}. "
+                f"Exit code: {result.returncode}. "
+                f"Error: {result.stderr.strip() if result.stderr else 'No error output'}"
+            )
+    except Exception as e:
+        _logger.warning(
+            f"Exception occurred while applying Windows permissions to {directory}: {e}"
+        )
 
 
 def has_gui_deps():
@@ -68,6 +138,10 @@ def _install_packages(packages, description):
             f"Missing {description} with non-standard set-up, installing into Cinema 4D's python"
         )
     subprocess.run(install_command, check=False)
+
+    # This ensures that system-wide installations grant access to all users
+    if module_directory.exists():
+        _apply_windows_read_execute_permissions_for_all_users(module_directory)
 
 
 def install_gui():
--- a/src/deadline/cinema4d_submitter/__init__.py
+++ b/src/deadline/cinema4d_submitter/__init__.py
@@ -61,7 +61,10 @@ def _apply_windows_read_execute_permissions_for_all_users(directory: Path) -> No
 
     # Construct icacls command with appropriate parameters
     # Grant Users group Read and Execute permissions with inheritance
-    icacls_command = ["icacls", str(directory), "/grant", "Users:(OI)(CI)(RX)", "/T"]
+    # The SID for /Users group is "S-1-5-32-545". This ensures that all the
+    # users of the workstation have access to the installed packages.
+    # https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers
+    icacls_command = ["icacls", str(directory), "/grant", "*S-1-5-32-545:(OI)(CI)(RX)", "/T"]
 
     try:
         # Execute the command without raising exception on non-zero exit
--- a/src/deadline/cinema4d_submitter/__init__.py
+++ b/src/deadline/cinema4d_submitter/__init__.py
@@ -29,8 +29,6 @@ def _has_windows_admin_privileges() -> bool:
         # IsUserAnAdmin() returns non-zero if the user is an admin, 0 otherwise
         return ctypes.windll.shell32.IsUserAnAdmin() != 0  # type: ignore[attr-defined]
     except Exception:
-        # If any exception occurs (e.g., attribute error, access denied),
-        # assume not running as admin and return False
         return False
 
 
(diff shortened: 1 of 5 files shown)


joel-wong-aws (dismissed):

src/deadline/cinema4d_submitter/__init__.py
```diff
@@ -6,11 +6,81 @@
...
+    Args:
+        directory: The directory path to apply permissions to
+    """
+    # If not running as admin early return
+    if not _has_windows_admin_privileges():
+        return
+
+    # Construct icacls command with appropriate parameters
+    # Grant Users group Read and Execute permissions with inheritance
+    icacls_command = ["icacls", str(directory), "/grant", "Users:(OI)(CI)(RX)", "/T"]
```
epmog: Does this work on non-english systems? We used to manually do icacls commands but it didn't work on any other language.

karthikbekalp: Good call, I was not aware of that.

Right now, we grant all the "Users" the permissions but we can make it language agnostic? https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers 

The security identifier for "Users" seem to be `S-1-5-32-545`. I'll test that out now.

karthikbekalp: Using the SID works! Updated the PR!

joel-wong-aws (dismissed):

joel-wong-aws (dismissed):

src/deadline/cinema4d_submitter/__init__.py
```diff
@@ -6,11 +6,84 @@
...
+    # If not running as admin early return
+    if not _has_windows_admin_privileges():
+        return
+
+    # Construct icacls command with appropriate parameters
+    # Grant Users group Read and Execute permissions with inheritance
+    # The SID for /Users group is "S-1-5-32-545". This ensures that all the
+    # users of the workstation have access to the installed packages.
+    # https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers
+    icacls_command = ["icacls", str(directory), "/grant", "*S-1-5-32-545:(OI)(CI)(RX)", "/T"]
```
edwards-aws, line 65: I'd recommend not using `icacls`. We've had issues in the past with `icacls` being slow, issues with parsing the standard output, but more importantly sometimes admins block `icacls`. So `icacls` won't always be available on the machine this is running on.

So we have a few choices here
1) Accept the risk with using `icacls` but give the user an actionable error message when it fails.
2) We've done something similar in the past with `pywin32`. This package doesn't have `pywin32` as a direct dependency, but Deadline Cloud does have `pywin32` dependency so it should be available through that. Keep in mind that `pywin32` comes with it's own baggage and we would like to remove the dependency at some point. https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/config/config_file.py#L215-L252
3) We could just use `ctypes`. It's a bit more work, but it's probably worth it long term to not have the dependency. Usually if `pywin32` has an API you can find the equivalent win32 API with a google, for example https://learn.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-addaccessallowedaceex and https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-lookupaccountnamew. We have lots of ctypes examples in the [openjd-sessions-for-python](https://github.com/OpenJobDescription/openjd-sessions-for-python/tree/mainline/src/openjd/sessions/_win32) repo

test/unit/deadline_submitter_for_cinema4d/test_init.py
```diff
@@ -0,0 +1,138 @@
...
+        with patch("deadline.cinema4d_submitter.subprocess.run") as mock_run:
+            _apply_windows_read_execute_permissions_for_all_users(test_dir)
+
+            # Assert
+            mock_has_admin.assert_called_once()
+            mock_run.assert_not_called()
+
+    @patch("deadline.cinema4d_submitter._has_windows_admin_privileges")
+    @patch("deadline.cinema4d_submitter.subprocess.run")
+    def test_executes_icacls_command_when_admin(self, mock_run, mock_has_admin):
```
edwards-aws, line 111: nit: I'm not sure how much the mocking-based tests provide here. We could instead have a temp directory with wrong permissions and validate that the tests set permissions correctly. As a bonus this would let us change the implementation without needing to rewrite tests and the tests would also serve as regression tests.

src/deadline/cinema4d_submitter/__init__.py
```diff
@@ -6,11 +6,84 @@
...
+
+    try:
+        import ctypes
+
+        # Use Windows API to check if the current process has admin privileges
+        # IsUserAnAdmin() returns non-zero if the user is an admin, 0 otherwise
+        return ctypes.windll.shell32.IsUserAnAdmin() != 0  # type: ignore[attr-defined]
+    except Exception:
+        # If any exception occurs (e.g., attribute error, access denied),
+        # assume not running as admin and return False
```
edwards-aws: nit: This comment isn't completely accurate. Exceptions only get thrown from `ctypes` if the function is being called incorrectly see https://docs.python.org/3/library/ctypes.html#exceptions

src/deadline/cinema4d_submitter/__init__.py
```diff
@@ -6,11 +6,84 @@
...
+    # Early return if not running on Windows
+    if not is_windows():
+        return False
+
+    try:
+        import ctypes
+
+        # Use Windows API to check if the current process has admin privileges
+        # IsUserAnAdmin() returns non-zero if the user is an admin, 0 otherwise
+        return ctypes.windll.shell32.IsUserAnAdmin() != 0  # type: ignore[attr-defined]
```
edwards-aws, line 30: nit: This is just more something to be aware of. The downside to using [IsUserAnAdmin](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-isuseranadmin#remarks) is that we don't get any error information from it. We'd have to call [CheckTokenMembership](https://learn.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-checktokenmembership) instead if we wanted more error info.

That being said if we don't care about any error state, this is fine.

src/deadline/cinema4d_submitter/__init__.py
karthikbekalp: Hmm, I was not aware of that. I'll work on this.

src/deadline/cinema4d_submitter/__init__.py
karthikbekalp: > That being said if we don't care about any error state, this is fine.

Currently, we use this to test if Cinema 4D was opened as "Admin" or not. If not, we would just not modify the permissions for the packages. 

But I think you bring up a good point, documentation is necessary for such a change because without it customers would not know the right way to do system installation. I'll work on this.

test/unit/deadline_submitter_for_cinema4d/test_init.py
karthikbekalp: > We could instead have a temp directory with wrong permissions and validate that the tests set permissions correctly.

That's a good idea. Let me modify that.

src/deadline/cinema4d_submitter/__init__.py
karthikbekalp: Currently, how this logic works is : 

1. Customer installs the submitter installer with as "Admin" user and proceeds with "system installation". 
2. Customer opens Cinema 4D as "Admin" and installs the dependencies. (This is the point where the GUI dependencies start installing from this code) 
3. Customer uses Cinema 4D with a different user. 

I think this code is probably just going to be temporary and will be fixed once we bundle the GUI dependencies ourselves in the submitter. 

> Accept the risk with using icacls but give the user an actionable error message when it fails.

~~I'll discuss with the team and get back to this.~~

I spoke with the team and it looks like giving a helpful message is something that we will prefer.

src/deadline/cinema4d_submitter/__init__.py
karthikbekalp: Removed the comment.

test/unit/deadline_submitter_for_cinema4d/test_init.py
karthikbekalp: Hmm, looks like `icacls` command only works if run in Admin. Hence, I'll just keep the test as is for now because we don't want to start running tests in admin just for this test.

edwards-aws (approved):

joel-wong-aws (approved):

PR #326, opened 2025-11-02
fix: Silently log errors and continue font handling.

karthikbekalp:

The Cinema 4D font utilities module was using exception-based error handling that could interrupt the job submission workflow. When font operations encountered errors (missing fonts, permission issues, non valid paths, etc.), the raised exceptions would block the entire submission process, preventing users from submitting render jobs even when the font issues were non-critical.

Converted the font_utils.py module from exception-based error handling to a resilient log-and-continue model.

Job submissions continue even when font operations fail. 

Yes, also added few unit tests. 

Running it. 

No

This change does not require documentation. 

No

If so, then please describe the changes that users of this package must make to update their scripts, or Python applications. Also,
please ensure that the title of your commit follows our conventional commit guidelines in 
[CONTRIBUTING.md](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/CONTRIBUTING.md#conventional-commits) for breaking changes.
*delete text ending here*

----

### DIFF
--- a/src/deadline/cinema4d_submitter/font_utils.py
+++ b/src/deadline/cinema4d_submitter/font_utils.py
@@ -122,6 +122,7 @@ def get_font_location(font_name: str) -> Optional[str]:
         Optional[str]: Path to the font file if found, None otherwise
     """
     if not font_name or not font_name.strip():
+        logger.error("Failed to locate font: font name is empty")
         return None
 
     font_name = font_name.strip()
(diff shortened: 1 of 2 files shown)


src/deadline/cinema4d_submitter/font_utils.py
```diff
@@ -348,17 +348,11 @@ def copy_font_to_scene_folder(font_name: str, scene_location: Path) -> None:
     try:
         shutil.copy2(font_location, destination)
         logger.debug(f"Successfully copied font from {font_location} to {destination}")
-
-        # Verify the copied file
-        if not is_font_file(str(destination)):
```
baxeaz, line 353: Should this basically never get hit so might as well remove it?

karthikbekalp: This code is run but its not useful. At line 314, we already check if a file is a font file before copying so checking again after copy is not useful. Hence, removed it...

joel-wong-aws (approved):

Cherie-Chen:
Font issue could be critical for some jobs. Is it possible to popup confirmation dialog to confirm is customer wants to continue submitting the job?

Approving to move fast, but can revisit it later

Cherie-Chen (approved):

karthikbekalp:
> Font issue could be critical for some jobs. Is it possible to popup confirmation dialog to confirm is customer wants to continue submitting the job?
> 
> Approving to move fast, but can revisit it later

Yup. That's a good callout. I'll cut an enhancement issue for this.

karthikbekalp:
Cut the issue: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/327

PR #342, opened 2025-11-18
feat: Add detailed Cinema 4D + Redshift logging option while submitting

karthikbekalp:

Currently, there are times when the current logs are not sufficient. 
We need a way to help customers to run the same job with detailed logging which they can share with us to deep dive. 

Added a new queue environment that starts running Cinema 4D + Redshift with debug logging enabled. 
This can slow down the rendering (hence its not the default), but it can give us more information in the logs which we can then use to debug or share with Maxon. 
Also, I added logic to extract the bugreports generated by Cinema 4D. 

Customers can easily resubmit a job with detailed logging enabled which can be then be shared with us. 

I manually tested this feature and also added unit tests.

image

As this is a job parameter, we can also use it in Resubmit job flows. 

image

Yes

Integration test require a little tweaking as this changes the template. I'll be making a follow-up PR with that fix. 

Yes. But this does not require a change to integ_helpers. 

Yes

No

----

### DIFF
--- a/docs/user_guide/faq-and-glossary.md
+++ b/docs/user_guide/faq-and-glossary.md
@@ -86,7 +86,7 @@ A: The submitter includes built-in error detection to catch common issues like m
 
 **Q: How do I enable detailed logging for debugging rendering issues?**
 
-A: The submitter includes a "Activate detailed logging" checkbox in the Job-Specific Settings that captures detailed logs for troubleshooting. When enabled:
+A: The submitter includes an "Activate detailed logging" checkbox in the Job-Specific Settings that captures detailed logs for troubleshooting. When enabled:
 
 1. The system automatically enables debug logging during rendering
 2. After rendering completes, all logs are printed to the job output
--- a/docs/user_guide/faq-and-glossary.md
+++ b/docs/user_guide/faq-and-glossary.md
@@ -84,6 +84,31 @@ A: Yes, you can set job priority levels in the submitter to control render queue
 
 A: The submitter includes built-in error detection to catch common issues like missing assets before submission. This can be deactivated in the submitter.
 
+**Q: How do I enable detailed logging for debugging rendering issues?**
+
+A: The submitter includes a "Activate detailed logging" checkbox in the Job-Specific Settings that captures detailed logs for troubleshooting. When enabled:
+
+1. The system automatically enables debug logging during rendering
+2. After rendering completes, all logs are printed to the job output
+
+To view the logs:
+1. Open Deadline Cloud Monitor
+2. Navigate to your completed job
+3. Right-click on a task and select "View logs"
+4. Enable the "View logs for all tasks" button
+5. Scroll through the task run logs to find the "Shut down DetailedLogging" section for detailed logs
+
+The logs are output in HTML format with timestamps. To save and view them in a web browser without timestamps:
+
+```bash
+cd deadline-cloud-for-cinema-4d/scripts
+python clean_redshift_detailed_logs.py /path/to/detailed_logs.log
+```
+
+Or run without arguments and enter the path when prompted. This generates a `redshift_log_cleaned.html` file that's easier to read and compare.
+
+Use detailed logging when you need to debug Redshift rendering issues, investigate rendering artifacts or errors, analyze renderer performance, or capture information for technical support.
+
 **Q: I'm getting permission errors with a system-wide installation on Windows. How do I fix this?**
 
 A: If users encounter permission errors when accessing the submitter after a system-wide installation:
--- a/docs/user_guide/submitter-features.md
+++ b/docs/user_guide/submitter-features.md
@@ -84,7 +84,7 @@ To view the detailed logs:
 2. Navigate to your completed job
 3. Right-click on a task and select "View logs"
 4. Enable the "View logs for all tasks" button.
-4. Scroll through the task run logs to find the "Shut down DetailedLogging" for detailed logs.
+5. Scroll through the task run logs to find the "Shut down DetailedLogging" for detailed logs.
 
 The Redshift logs are output in HTML format and include timestamps at the start of each line. If you want to save and view the logs in a web browser without timestamps, you can use the provided cleanup utility:
 
--- a/docs/user_guide/submitter-features.md
+++ b/docs/user_guide/submitter-features.md
@@ -41,6 +41,7 @@ Settings specific to your Cinema 4D render:
 - **Takes** - Select which Cinema 4D takes to render
 - **Override Frame Range** - Override the frame range from your scene settings
 - **Automatic Error Checking** - Optional checkbox to activate/deactivate error checking during rendering
+- **Activate detailed logging** - Enable detailed logging to capture detailed logs for debugging rendering issues. When enabled, debug logs are automatically captured and printed to the job output for easy review
 - **Task Run Timeout** - Maximum time allowed for each task to complete
 - **Cinema 4D Launch Timeout** - Maximum time allowed for Cinema 4D to start up
 - **Cinema 4D Shutdown Timeout** - Maximum time allowed for Cinema 4D to shut down cleanly
@@ -55,6 +56,58 @@ Settings specific to your Cinema 4D render:
 **Host Requirements** (optional) - Allows you to specify which types of hosts will be eligible for picking up tasks for this job.
 
 The submitter handles the technical details so you can focus on your creative work.
+
+## Detailed Logging for Debugging
+
+The **Activate detailed logging** feature helps you troubleshoot rendering issues by capturing detailed logs.
+
+### When to Use Detailed Logging
+
+Enable detailed logging when you need to:
+- Debug Redshift rendering issues or unexpected behavior
+- Investigate rendering artifacts or errors
+- Analyze renderer performance and behavior
+- Capture detailed information for technical support
+
+### How It Works
+
+When you enable the "Activate detailed logging" checkbox in the Job-Specific Settings:
+
+1. **Log Capture** - The system automatically enables Redshift debug logging by setting the `REDSHIFT_DEBUGCAPTURE` environment variable
+2. **Log Output** - After rendering completes, all Redshift logs are automatically printed to the job output for easy review
+
+### Viewing the Logs
+
+To view the detailed logs:
+
+1. Open Deadline Cloud Monitor
+2. Navigate to your completed job
+3. Right-click on a task and select "View logs"
+4. Enable the "View logs for all tasks" button.
+4. Scroll through the task run logs to find the "Shut down DetailedLogging" for detailed logs.
+
+The Redshift logs are output in HTML format and include timestamps at the start of each line. If you want to save and view the logs in a web browser without timestamps, you can use the provided cleanup utility:
+
+1. Download the detailed log file from Deadline Cloud Monitor (right-click on task → "Download logs")
+2. Run the cleanup script with the log file path:
+   ```bash
+   cd deadline-cloud-for-cinema-4d/scripts
+   python clean_redshift_detailed_logs.py /path/to/detailed_logs.log
+   ```
+   Or run it without arguments and enter the path when prompted:
+   ```bash
+   python clean_redshift_detailed_logs.py
+   ```
+3. Open the generated `redshift_log_cleaned.html` file in your web browser
+
+The cleanup script automatically extracts the Redshift HTML logs from the detailed log file and removes timestamp prefixes (e.g., `2024/11/17 14:23:45-08:00`) from each line, making the logs easier to read and compare.
+
+### Important Notes
+
+- Detailed logging is **disabled by default** to minimize overhead
+- Only enable it when you need to debug specific issues
+- Logs are captured on a per-job basis - each job has its own log output
+- The feature works across Windows, Linux worker nodes
 
 ---
 
--- a/scripts/clean_redshift_detailed_logs.py
+++ b/scripts/clean_redshift_detailed_logs.py
@@ -1,0 +1,150 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+Utility script to extract and clean Redshift debug logs from Cinema 4D detailed log files.
+
+This script extracts HTML Redshift debug logs from Cinema 4D detailed log files and
+removes timestamp prefixes to make them easier to read and compare. It's particularly
+useful when viewing detailed logs captured by the Cinema 4D submitter's detailed logging feature.
+
+Usage:
+    python clean_redshift_detailed_logs.py [log_file_path]
+
+If no log file path is provided, the script will prompt for one.
+
+The script will:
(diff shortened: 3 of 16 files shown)


vaish-gujjari (approved):

docs/user_guide/faq-and-glossary.md
```diff
@@ -84,6 +84,31 @@ A: Yes, you can set job priority levels in the submitter to control render queue
 
 A: The submitter includes built-in error detection to catch common issues like missing assets before submission. This can be deactivated in the submitter.
 
+**Q: How do I enable detailed logging for debugging rendering issues?**
+
+A: The submitter includes a "Activate detailed logging" checkbox in the Job-Specific Settings that captures detailed logs for troubleshooting. When enabled:
```
joel-wong-aws: nit: `a "Activate detailed logging"` -> `an "Activate detailed logging"`

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -135,8 +135,11 @@ def _build_ui(self, settings):
         self.activate_error_checking_chck = QCheckBox("Activate automatic error checking", self)
         lyt.addWidget(self.activate_error_checking_chck, 5, 0)
 
+        self.activate_detailed_logging_chck = QCheckBox("Activate detailed logging", self)
```
joel-wong-aws: Should this be `Activate detailed Redshift logging` since it only applies to Redshift?

Also, what does it do in the case of physical or Arnold renders?

scripts/clean_redshift_detailed_logs.py
```diff
@@ -0,0 +1,150 @@
...
+            content = file.read()
+    except FileNotFoundError:
+        print(f"Error: File '{log_file_path}' not found.")
+        return None
+    except Exception as e:
+        print(f"Error reading file: {e}")
+        return None
+
+    # Find the start marker
+    start_pattern = r"REDSHIFT DEBUG LOG:.*?={70,}"
```
joel-wong-aws: Why `.*?={70,}`?

scripts/clean_redshift_detailed_logs.py
```diff
@@ -0,0 +1,150 @@
...
+
+    end_pos = start_pos + end_match.start() - 2
+
+    # Extract the content between markers
+    extracted_content = content[start_pos:end_pos].strip()
+
+    return extracted_content
+
+
+def remove_timestamps(input_text):
```
joel-wong-aws, line 84: Why do we need to remove timestamps? I would imagine timestamps are helpful for debugging?

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
```diff
@@ -55,6 +55,17 @@ parameterDefinitions:
   allowedValues:
   - '1'
   - '0'
+- name: DetailedLogging
```
joel-wong-aws, line 58: Do we need to update all the integ tests with parameter?

docs/user_guide/faq-and-glossary.md
```diff
@@ -84,6 +84,31 @@ A: Yes, you can set job priority levels in the submitter to control render queue
...
+
+1. The system automatically enables debug logging during rendering
+2. After rendering completes, all logs are printed to the job output
+
+To view the logs:
+1. Open Deadline Cloud Monitor
+2. Navigate to your completed job
+3. Right-click on a task and select "View logs"
+4. Enable the "View logs for all tasks" button
+5. Scroll through the task run logs to find the "Shut down DetailedLogging" section for detailed logs
```
diane-hj, line 99: Nit. Should it be Detailed Logging? 👀

docs/user_guide/submitter-features.md
```diff
@@ -56,6 +57,58 @@ Settings specific to your Cinema 4D render:
...
+
+### Viewing the Logs
+
+To view the detailed logs:
+
+1. Open Deadline Cloud Monitor
+2. Navigate to your completed job
+3. Right-click on a task and select "View logs"
+4. Enable the "View logs for all tasks" button.
+4. Scroll through the task run logs to find the "Shut down DetailedLogging" for detailed logs.
```
diane-hj: Nit. double 4s :p

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    found_logs = FoundLogs()
+    conda_prefix = os.environ.get("CONDA_PREFIX", "")
+    home_dir = os.path.expanduser("~")
+
+    print("Searching for log files...")
+
+    # Check for Redshift log.html: $CONDA_PREFIX/redshiftlocaldata/log/log.latest.0/log.html
+    found_logs.redshift = _find_redshift_log(
+        conda_prefix, ["redshiftlocaldata", "log", "log.latest.0", "log.html"]
```
joel-wong-aws, line 103: This path is only correct for the Deadline Cloud Conda package. Is there a way we can generalize it to all workers? These paths are controlled by these env vars: https://help.maxon.net/r3d/softimage/en-us/Content/html/Custom%2BInstall%2BLocations.html

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    found_logs = FoundLogs()
+    conda_prefix = os.environ.get("CONDA_PREFIX", "")
+    appdata = os.environ.get("APPDATA", "")
+
+    print("Searching for log files...")
+
+    # Check for Redshift log.html: $CONDA_PREFIX\cinema4d\RedshiftData\Log\Log.Latest.0\log.html
+    found_logs.redshift = _find_redshift_log(
+        conda_prefix, ["cinema4d", "RedshiftData", "Log", "Log.Latest.0", "log.html"]
```
joel-wong-aws, line 127: This path is only correct for the Deadline Cloud Conda package. Is there a way we can generalize it to all workers? These paths are controlled by these env vars: https://help.maxon.net/r3d/softimage/en-us/Content/html/Custom%2BInstall%2BLocations.html

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -0,0 +1,205 @@
...
+    print(f"\n{'='*80}")
+    print(f"END OF {log_type}")
+    print(f"{'='*80}\n")
+
+
+if __name__ == "__main__":
+    enabled = sys.argv[1] if len(sys.argv) > 1 else "0"
+
+    if enabled != "1":
+        print("Detailed logging is disabled, skipping log output.")
```
joel-wong-aws, line 185: Can we use a `return` here? Then we don't need an indented `else`

docs/user_guide/submitter-features.md
```diff
@@ -56,6 +57,58 @@ Settings specific to your Cinema 4D render:
...
+
+The Redshift logs are output in HTML format and include timestamps at the start of each line. If you want to save and view the logs in a web browser without timestamps, you can use the provided cleanup utility:
```
diane-hj, line 89: Nit. I think this part has duplicated contents with `faq-and-glossary.md`. What do you think of making this file a single source of truth and having the FAQ link here? That way, we can avoid maintaining duplicate documentation.

docs/user_guide/faq-and-glossary.md
karthikbekalp: I will update this.

scripts/clean_redshift_detailed_logs.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # Find the end marker
+    end_pattern = r"END OF REDSHIFT DEBUG LOG"
+    end_match = re.search(end_pattern, content[start_pos:])
+
+    if not end_match:
+        print("Error: Could not find 'END OF REDSHIFT DEBUG LOG' marker in the file.")
+        return None
+
+    end_pos = start_pos + end_match.start() - 2
```
diane-hj: Could you let me know what is this -2 for please?

scripts/clean_redshift_detailed_logs.py
karthikbekalp: This is so that it only detects the log after 
```
REDSHIFT DEBUG LOG:
=============================
```

scripts/clean_redshift_detailed_logs.py
karthikbekalp: This is only a utility function to remove the timestamps from our worker logs to cleanup the logs when customers send us over their task run logs. 

For example, the logs would now look like which is not a valid html: 
```
2025/11/18 12:06:52-08:00 <div class="DEBUG line"><div>20:06:50   4044MB DEBUG:        </div><div>	Shut down ok
2025/11/18 12:06:52-08:00 </div></div>
2025/11/18 12:06:52-08:00 <div class="DEBUG line"><div>20:06:50   4044MB DEBUG:        </div><div>PostFX: Shut down
2025/11/18 12:06:52-08:00 </div></div>
2025/11/18 12:06:52-08:00 <div class="DEBUG line"><div>20:06:50   4044MB DEBUG:        </div><div>Shutdown GPU Devices...
```

This script would remove the timestamps to something like this: 

```
<div class="DEBUG line"><div>20:06:50   4044MB DEBUG:        </div><div>	Shut down ok
</div></div>
<div class="DEBUG line"><div>20:06:50   4044MB DEBUG:        </div><div>PostFX: Shut down
</div></div>
<div class="DEBUG line"><div>20:06:50   4044MB DEBUG:        </div><div>Shutdown GPU Devices...
```

src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
karthikbekalp: Yes, as explained in my PR description, a follow up CR is coming up with the integ test changes. 
Not only this but also the extra queue environments.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+    print(f"\n{'='*80}")
+    print(f"{log_type}: {log_file}")
+    print(f"{'='*80}\n")
+
+    try:
+        with open(log_file, "r", encoding="utf-8", errors="replace") as f:
+            print(f.read())
+    except Exception as e:
+        print(f"Error: Error reading log file {log_file}: {e}")
```
diane-hj: Super Nit. To be consistent with other print lines,should we remove "Error:" in the front?

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: That's a good idea. How about I add it in follow-up? We want to ship this to one of our customers right away...

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: That's a good idea. How about I add it in follow-up? We want to ship this to one of our customers right away...

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Sure, adding it right away.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def print_log_file(log_file, log_type="LOG"):
+    """Print the contents of a log file.
+
+    Args:
+        log_file: Path to the log file to print.
+        log_type: Type of log file (for display purposes).
+    """
+    print(f"\n{'='*80}")
```
diane-hj, line 165: Could you help me understand how we decided on 80 equals signs here? I noticed the log parsing script uses ={70,} to match separators, so I'm curious about the relationship between these two numbers and how we arrived at 80 👀 :

docs/user_guide/faq-and-glossary.md
karthikbekalp: I think its a drawback in OpenJD, we cannot have space within the name.

docs/user_guide/submitter-features.md
karthikbekalp: Fixing it in the upcoming revision.

docs/user_guide/submitter-features.md
karthikbekalp: Yeah, I was wondering about that. I think right now it makes sense to have it in both so a little duplication is ok when compared to higher visibility for customers. 

But I can remove it if its not required in my follow up PR, what do you think?

scripts/clean_redshift_detailed_logs.py
karthikbekalp: For ignoring the "=========" at the end.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Sure, updated.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: The clean redshift logs script is a nice little script to clean the logs when we have a log file. Generally useful for just devs like us. 
I think it could be just 80 and looks like the LLMs just counted until 70.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Hmm, looks like as this is not a function, we cannot use return here. I'd like to keep it like this for now if you don't mind.

scripts/clean_redshift_detailed_logs.py
karthikbekalp: I refactored to remove the regex to avoid the sonarcloud issues entirely.

diane-hj (approved):

scripts/clean_redshift_detailed_logs.py
joel-wong-aws: That's a bit brittle, we should investigate ways to make it more consistent. We can punt that

scripts/clean_redshift_detailed_logs.py
joel-wong-aws: Thanks for explaining!

joel-wong-aws (approved):

PR #346, opened 2025-11-19
chore: add validation to detect pipe characters in asset paths

### LINKED ISSUE #338: Feature request: Prevent or warn customers from submitting jobs with "|" in the file or path name
### Describe the problem

Currently, if there is "|" in the path or file name we fail to sync the asset on Windows workers with the following error. 

If there is "|" in asset file name. 
```
2025/11/08 16:57:20-08:00 [Errno 22] Invalid argument: 'C:\\Sessions\\session-abc\\assetroot-abc\\tex\\Map-COL-with"|".jpg.af4d5BFa'
```

If there is "|" in path name. 

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Sessions\\session-abc\\assetroot-abc\\tex with | '
```

### Proposed Solution

Instead of uploading the assets and failing when downloading the assets we should be preventing customers from submitting these scene or at least let them know that Windows doesn't allow "|" character. 

### Example Use Cases

Whenever someone submits a job with "|" in the file name or path name.

vaish-gujjari:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/338

Currently, if there is "|" in the path or file name we fail to sync the asset on Windows workers with the following error.

If there is "|" in asset file name.

2025/11/08 16:57:20-08:00 [Errno 22] Invalid argument: 'C:\\Sessions\\session-abc\\assetroot-abc\\tex\\Map-COL-with"|".jpg.af4d5BFa'
If there is "|" in path name.

OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Sessions\\session-abc\\assetroot-abc\\tex with | '

Created a path_validator module with a validate_asset_paths() function that iterates through all asset paths and checks for the pipe character "|". When detected, it logs a warning through the WarningCollectorHandler which displays the message in the submitter UI, alerting users before job submission that these paths will cause sync failures on Windows workers.

Screenshot 2025-11-20 at 2 50 43 PM

Users will now receive clear warnings in the submitter UI when their asset paths contain pipe characters, allowing them to fix the issue before submitting jobs. This prevents job failures on Windows workers and saves time by catching the problem early rather than discovering it after submission when assets fail to sync.

Added comprehensive unit tests in test_path_validator.py covering:
- Detection of pipe characters in filenames
- Detection of pipe characters in directory names
- Handling multiple paths with pipe characters
- Validation that clean paths don't trigger warnings
- Edge case of empty asset sets
All tests pass along with linting (ruff, black, mypy) validation.

No

No

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
+++ b/src/deadline/cinema4d_submitter/adaptor_cinema4d_job_template.yaml
@@ -66,6 +66,17 @@ parameterDefinitions:
   allowedValues:
   - '1'
   - '0'
+- name: DetailedLogging
+  type: STRING
+  userInterface:
+    control: CHECK_BOX
+    label: Enable Detailed Logging
+    groupLabel: Cinema4D Settings
+  description: Enable detailed Cinema 4D + Redshift renderer logging for debugging.
+  default: '0'
+  allowedValues:
+  - '1'
+  - '0'
 steps:
 - name: Render
   parameterSpace:
--- a/test/integ/test_scenes/physical/expected_job_bundle/parameter_values.yaml
+++ b/test/integ/test_scenes/physical/expected_job_bundle/parameter_values.yaml
@@ -7,6 +7,8 @@ parameterValues:
   value: ''
 - name: ActivateErrorChecking
   value: '1'
+- name: DetailedLogging
+  value: '0'
 - name: Frames
   value: '1'
 - name: CondaPackages
--- a/test/integ/test_scenes/physical/expected_job_bundle/template.yaml
+++ b/test/integ/test_scenes/physical/expected_job_bundle/template.yaml
@@ -54,6 +54,17 @@ parameterDefinitions:
   allowedValues:
   - '1'
   - '0'
+- name: DetailedLogging
+  type: STRING
+  userInterface:
+    control: CHECK_BOX
+    label: Enable Detailed Logging
+    groupLabel: Cinema4D Settings
+  description: Enable detailed Cinema 4D + Redshift renderer logging for debugging.
+  default: '0'
+  allowedValues:
+  - '1'
+  - '0'
 steps:
 - name: Main
   parameterSpace:
--- a/test/integ/test_scenes/physical_multi_takes/expected_job_bundle/parameter_values.yaml
+++ b/test/integ/test_scenes/physical_multi_takes/expected_job_bundle/parameter_values.yaml
@@ -7,6 +7,8 @@ parameterValues:
   value: ''
 - name: ActivateErrorChecking
   value: '1'
+- name: DetailedLogging
+  value: '0'
 - name: MainFrames
   value: '1'
 - name: _0123456789abcdef0123456789abcdef0123456789abcdef012345678Frames
--- a/test/integ/test_scenes/physical_multi_takes/expected_job_bundle/template.yaml
+++ b/test/integ/test_scenes/physical_multi_takes/expected_job_bundle/template.yaml
@@ -46,6 +46,17 @@ parameterDefinitions:
   allowedValues:
   - '1'
   - '0'
+- name: DetailedLogging
+  type: STRING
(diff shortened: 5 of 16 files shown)


src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_client.py
```diff
@@ -44,7 +44,7 @@ def graceful_shutdown(self, signum: int, frame: FrameType | None):
         sys.exit(0)
 
     def map_path(self, path: str) -> str:
-        """
+        r"""
```
karthikbekalp: Typo?

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -26,6 +27,9 @@ def parse_scene_assets(self) -> set[Path]:
         # Clear any previous warnings before processing
         warning_collector.clear_warnings()
 
+        # Clear any previous warnings before processing
+        warning_collector.clear_warnings()
```
karthikbekalp: Any reason we have a duplicate?

karthikbekalp (changes requested):
I think the changes look good, but can you also add images on how the feature will look like?

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_client.py
vaish-gujjari: Hi Karthik, no. So when we run "hatch run fmt" we're getting the below issue:

test/unit/deadline_submitter_for_cinema4d/test_warning_collector.py::TestWarningCollector::test_get_warnings_returns_copy <unknown>:47: SyntaxWarning: invalid escape sequence '\p'
~\Desktop\deadline-cloud-for-cinema-4d\src\deadline\cinema4d_adaptor\Cinema4DClient\cinema4d_client.py:47: SyntaxWarning: invalid escape sequence '\p'
"""
<unknown>:47: SyntaxWarning: invalid escape sequence '\p'
~\Desktop\deadline-cloud-for-cinema-4d\src\deadline\cinema4d_adaptor\Cinema4DClient\cinema4d_client.py:47: SyntaxWarning: invalid escape sequence '\p'
"""
<unknown>:47: SyntaxWarning: invalid escape sequence '\p'
<unknown>:47: SyntaxWarning: invalid escape sequence '\p'
~\Desktop\deadline-cloud-for-cinema-4d\src\deadline\cinema4d_adaptor\Cinema4DClient\cinema4d_client.py:47: SyntaxWarning: invalid escape sequence '\p'
"""

By adding r before """,  we're treating all the characters as literals and not interpret as escape sequences..

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: Good catch! This was an oversight on my part. I've cleaned it up.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_client.py
karthikbekalp: Hmm, I don't think adding "r" is the right solution here because: 

1. The code currently in mainline passes the formatter test (i.e. `hatch run fmt` is successful)

2. I think we want to deep dive why we are getting this error in the first place. The error is raised by "test_get_warnings_returns_copy" which is a submitter code trying to check client code which does not look right. Maybe there's an issue in your workspace? 

```
test/unit/deadline_submitter_for_cinema4d/test_warning_collector.py::TestWarningCollector::test_get_warnings_returns_copy :
```

3. The ":47: SyntaxWarning: invalid escape sequence '\p'" is interesting because it does bring up the issue that we do have values in the docstrings that could be misinterpreted as escape characters. 

Let's try to debug this offline once you head over... :)

karthikbekalp (changes requested):

PR #349, opened 2025-11-21
feat: Add Cinema 4D detailed debug logs.

karthikbekalp:

This is a follow up to https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/342 and adds Cinema 4D debug logging (we only added Redshift and Cinema 4D crash reports earlier)

Add Cinema 4D debug logging. 
There was also a comment in my previous PR to use the different Redshift env variables for CMF. So updated that. 

We will now even log the Cinema 4D debug logging. 

Yes

Currently running it. 

Yes but doesn't require any changes to helper. 

Yes

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
+++ b/src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
@@ -3,6 +3,10 @@
 from dataclasses import dataclass, field
 import os
 import sys
+import tempfile
+
+# Cinema 4D detailed log filename constant
+C4D_DETAILED_LOG_FILENAME = "c4d_detailed_logs.txt"
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -11,30 +15,102 @@ class FoundLogs:
...
+    2. REDSHIFT_LOCALDATAPATH environment variable (if set)
+    3. Platform-specific default paths
+
+    Returns:
+        List of paths to check for Redshift logs.
+    """
+    paths = []
+
+    # Add conda prefix path first if available (most likely in rendering environment)
+    conda_prefix = os.environ.get("CONDA_PREFIX")
```
joel-wong-aws: Can we make `CONDA_PREFIX` a constant? We use it in a couple places, and are also inconsistent about whether we pass it around or not. Maybe we should make a function to retrieve the path?

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -11,30 +15,102 @@ class FoundLogs:
...
+
+    Returns paths in priority order:
+    1. CONDA_PREFIX-based paths (for conda environments)
+    2. REDSHIFT_LOCALDATAPATH environment variable (if set)
+    3. Platform-specific default paths
+
+    Returns:
+        List of paths to check for Redshift logs.
+    """
+    paths = []
```
joel-wong-aws: Should paths be a set? Some of these paths may be duplicates

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -11,30 +15,102 @@ class FoundLogs:
...
+    """Search for Cinema 4D detailed log file.
+
+    Args:
+        conda_prefix: The CONDA_PREFIX environment variable value.
+
+    Returns:
+        List containing the Cinema 4D detailed log path if found, empty list otherwise.
+    """
+    # Try CONDA_PREFIX first if available
+    if conda_prefix and os.path.exists(conda_prefix):
```
joel-wong-aws: It seems like we have a couple different approaches for searching for logs between all these functions. Can we make them more consistent? e.g. they could all be organized as:
1. Generate the list of all possible paths for the log file (and differentiate on Windows and Linux within the function)
2. Check each path for logs and save the first matching one

Right now, the logic is a bit confusing with separate Linux and Windows functions calling shared ones which within them again separate Windows and Linux cases.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -170,36 +267,71 @@ def print_log_file(log_file, log_type="LOG"):
...
+        log_files: List of log file paths to print.
+        log_type: Type identifier for the log (e.g., "REDSHIFT DEBUG LOG").
+        description: Human-readable description of the log type.
+        not_found_message: Message to display when no logs are found.
+    """
+    if log_files:
+        print(f"\nFound {len(log_files)} {description}")
+        for log_file in log_files:
+            print_log_file(log_file, log_type)
     else:
```
joel-wong-aws: nit: guard clauses could be used here

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
```diff
@@ -11,30 +15,102 @@ class FoundLogs:
...
+            print(f"Found Cinema 4D detailed log: {c4d_log_path}")
+            return [c4d_log_path]
+        else:
+            print("Cinema 4D detailed log not found in CONDA_PREFIX")
     else:
-        print("Redshift log not found at expected location")
+        print("CONDA_PREFIX not set or doesn't exist")
+
+    # Fallback to temp directory
+    temp_dir = tempfile.gettempdir()
```
joel-wong-aws: This temp directly is generally readable by all users on the system. Can we set file permissions using `mkdtemp` so that it's only readable by the current user, and then pass that location around as an env var?

vaish-gujjari (dismissed):
Reviewed and approved!

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Updated!

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Updated!

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Updated. I refactored the whole logic to be easy to read but I think I should have punted on this effort considering we want to give the updated release to customers.

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Updated

src/deadline/cinema4d_submitter/detailed_logging_scripts/print_logs.py
karthikbekalp: Updated!

src/deadline/cinema4d_submitter/detailed_logging_scripts/setup_logging.py
```diff
@@ -1,25 +1,88 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-"""Verify Redshift debug logging environment variable is enabled."""
+"""Configure environment variables for Cinema 4D detailed logging."""
 import os
 import sys
+import tempfile
 
+# Constants
+C4D_DETAILED_LOG_FILENAME = "c4d_detailed_logs.txt"
```
vaish-gujjari, line 8: Can we define these in a separate file that can be accessed in multiple files than defining them separately in each file? :)

karthikbekalp: Good question, unfortunately we run the scripts as queue environments so we can not make separate files as the queue environment would not have these separate files...

joel-wong-aws (approved):

vaish-gujjari (approved):

PR #353, opened 2025-11-24
fix(submitter): exclude Maxon DB assets from job bundle references

### LINKED ISSUE #300: Bug: Maxon DB assets included in `asset_references.yaml`
### Describe the bug

When I submit a job that references Maxon DB assets (e.g. the Showroom Template example), the Maxon DB assets are included in the `asset_references.yaml` file, even though they do not exist on the local filesystem, and still render correctly without them being included in the job submission (since they're downloaded from Maxon's asset DB).

### Expected Behaviour

Maxon DB assets should not be included in `asset_references.yaml`. We should suggest using localized assets instead of asset references using the "Bundled job submission" option.

### Current Behaviour

There is a warning message about missing assets in the job bundle during submission. You can bypass this error by simply clicking "OK" and the scene renders correctly anyway.

### Proposed Solution

Filter out Maxon DB assets starting with `asset:` or `asset:` [here](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/dc934845322f275fd03607093fd79ee181aecfd1/src/deadline/cinema4d_submitter/assets.py#L48).

### Reproduction Steps

1. Open Cinema 4D 2024 (or 2025)
2. Choose the Showroom Template scene from the quick start
3. Use the AWS Deadline Cloud Submitter to submit the job without the job bundle setting selected

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.): Reproduced on Windows 11 Enterprise 24H2, likely affects other operating systems as well
2. Version of Cinema 4D: Reproduced on Cinema 4D 2024, likely affects 2025 as well
3. Version of this package: 0.8.2
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer? Affects submitter 2025.08.26 and earlier

vaish-gujjari:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/300

When I submit a job that references Maxon DB assets (e.g. the Showroom Template example), the Maxon DB assets are included in the asset_references.yaml file, even though they do not exist on the local filesystem, and still render correctly without them being included in the job submission (since they're downloaded from Maxon's asset DB).

Added a prefix filter in AssetIntrospector.parse_scene_assets() to exclude Maxon Database assets from the collected asset list. The solution checks if asset filenames start with "asset:" or "assetdb://" prefixes and logs excluded assets to the Cinema 4D console for user visibility. These prefixes identify cloud-based assets that Cinema 4D automatically downloads from Maxon's servers at render time, so they don't need to be included in the job bundle's asset references.

Screenshot 2025-11-24 at 2 02 00 PM

Screenshot 2025-11-25 at 12 17 18 PM

**User Experience:**
- Eliminates false "missing assets" warnings during job submission
- Removes misleading "Job submission confirmation" popups about S3 uploads for Maxon DB assets
- Prevents unnecessary "sync attachments" sessions in task logs
- Provides transparency via info logs showing which Maxon DB assets were excluded

**Technical:**
- Maxon DB assets (HDRIs, materials, models from Maxon's asset database) are no longer included in asset_references.yaml
- Job bundles are cleaner and only contain actual local filesystem assets
- No impact on rendering - Maxon DB assets still work correctly as Cinema 4D handles them automatically
- Console logging allows users to verify expected assets were excluded

**Backward Compatibility:**
- Fully backward compatible - only filters out non-existent cloud assets that were causing false warnings

**Recommended testing:**
- Opened Cinema 4D 2024/2025 and loaded the Showroom Template scene (contains Maxon DB assets)
- Submitted job using AWS Deadline Cloud Submitter without "Bundled job submission" option
- Verified no "missing assets" warnings appear during submission
- Verified no S3 upload confirmation dialog appears
- Checked Cinema 4D console contains entries like: [Deadline Cloud] *** EXCLUDING Maxon DB asset: assetdb:///HDRIs/Studio/hdr-light-studio_17.exr
- Verified job renders successfully with Maxon DB assets
- Checked task logs confirm no unnecessary "sync attachments" sessions
- Tested with scenes containing both local assets and Maxon DB assets to ensure local assets are still properly included and only Maxon DB assets are logged as excluded

Yes, I have made changes to the submitter. Specifically:
File modified: src/deadline/cinema4d_submitter/assets.py

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/assets.py
+++ b/src/deadline/cinema4d_submitter/assets.py
@@ -11,9 +11,12 @@ from .platform_utils import is_windows
 from .scene import Scene
 from .font_utils import is_asset_a_font, copy_font_to_scene_folder, FONTS_DIR
 from .warning_collector import warning_collector
+from .warning_logging_handler import WarningCollectorHandler
 from .path_validator import validate_asset_paths
 
 logger = logging.getLogger(__name__)
+if not any(isinstance(h, WarningCollectorHandler) for h in logger.handlers):
+    logger.addHandler(WarningCollectorHandler())
 
 _FRAME_RE = re.compile("#+")
 
@@ -48,8 +51,6 @@ class AssetIntrospector:
             flags=c4d.ASSETDATA_FLAG_WITHFONTS,
         )
 
-        print(f"[Deadline Cloud] Total assets found: {len(asset_list)}")
-
         for asset in asset_list:
             # Only process fonts on Windows. Mac font functionality is not supported
             if is_windows() and is_asset_a_font(asset):
@@ -58,13 +59,9 @@ class AssetIntrospector:
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
 
-            # Debug: print all filenames to see what we're getting
-            if filename:
-                print(f"[Deadline Cloud] Asset: {filename} (exists={exists})")
-
             # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
             if filename is not None and filename.startswith(("asset:", "assetdb://")):
-                print(f"[Deadline Cloud] *** EXCLUDING Maxon DB asset: {filename}")
+                logger.warning(f"Excluding Maxon DB asset from job bundle: {filename}")
             elif exists is True and filename is not None:
                 assets.add(Path(filename))
 
--- a/src/deadline/cinema4d_submitter/assets.py
+++ b/src/deadline/cinema4d_submitter/assets.py
@@ -62,7 +62,9 @@ class AssetIntrospector:
             # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
             if filename is not None and filename.startswith(("asset:", "assetdb://")):
                 logger.warning(f"Excluding Maxon DB asset from job bundle: {filename}")
-            elif exists is True and filename is not None:
+                continue
+
+            if exists is True and filename is not None:
                 assets.add(Path(filename))
 
         # Add all font files from the fonts directory to assets (Windows only)
--- a/src/deadline/cinema4d_submitter/assets.py
+++ b/src/deadline/cinema4d_submitter/assets.py
@@ -61,7 +61,10 @@ class AssetIntrospector:
 
             # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
             if filename is not None and filename.startswith(("asset:", "assetdb://")):
-                logger.warning(f"Excluding Maxon DB asset from job bundle: {filename}")
+                logger.warning(
+                    f"Excluding Maxon DB asset from job bundle: {filename}\n"
+                    "To include assets with job submission, use 'File > Save Project with Assets' to localize them first."
+                )
                 continue
 
             if exists is True and filename is not None:
--- a/src/deadline/cinema4d_submitter/assets.py
+++ b/src/deadline/cinema4d_submitter/assets.py
@@ -49,7 +49,7 @@ class AssetIntrospector:
         )
 
         print(f"[Deadline Cloud] Total assets found: {len(asset_list)}")
-        
+
         for asset in asset_list:
             # Only process fonts on Windows. Mac font functionality is not supported
             if is_windows() and is_asset_a_font(asset):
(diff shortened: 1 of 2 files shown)


src/deadline/cinema4d_submitter/assets.py
```diff
@@ -12,6 +13,8 @@
 from .warning_collector import warning_collector
 from .path_validator import validate_asset_paths
 
+logger = logging.getLogger(__name__)
```
karthikbekalp, line 17: Discussed offline, this was not necessary in this iteration of the code. But I think it will be required when we add this part of the issues before submission window (just like in the font handling code)?

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -45,14 +48,24 @@ def parse_scene_assets(self) -> set[Path]:
...
             if is_windows() and is_asset_a_font(asset):
                 copy_font_to_scene_folder(asset["assetname"], path_to_scene_file_dir)
 
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
-            if exists is True and filename is not None:
+
+            # Debug: print all filenames to see what we're getting
+            if filename:
+                print(f"[Deadline Cloud] Asset: {filename} (exists={exists})")
```
karthikbekalp: Hmm, I'm not sure if this is required?

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -45,14 +48,24 @@ def parse_scene_assets(self) -> set[Path]:
...
+
+            # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
+            if filename is not None and filename.startswith(("asset:", "assetdb://")):
```
karthikbekalp, line 63: I think its best to use guard clauses here and break from the loop. This can help us easily understand how the conditions are structured.

karthikbekalp (changes requested):
Discussed offline and added few comments.

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: yep, modified the same

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: got it

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: no, not really, just an additional debugging helper

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -52,6 +58,12 @@ def parse_scene_assets(self) -> set[Path]:
 
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
+
+            # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
+            if filename is not None and filename.startswith(("asset:", "assetdb://")):
+                logger.warning(f"Excluding Maxon DB asset from job bundle: {filename}")
```
karthikbekalp: Hmm, should we update the log to suggest to use localized assets if they still want to include the assets with job submission? According to the recommendation here: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/300 ?

vaish-gujjari: modified the warning message, Karthik

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -52,6 +58,15 @@ def parse_scene_assets(self) -> set[Path]:
...
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
+
+            # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
+            if filename is not None and filename.startswith(("asset:", "assetdb://")):
+                logger.warning(
+                    f"Excluding Maxon DB asset from job bundle: {filename}\n"
+                    "To include assets with job submission, use 'File > Save Project with Assets' to localize them first."
+                )
+                continue
```
joel-wong-aws, line 69: Can we add unit tests for this change?

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -52,6 +58,15 @@ def parse_scene_assets(self) -> set[Path]:
 
             filename = asset.get("filename", None)
             exists = asset.get("exists", False)
+
+            # Filter out Maxon DB assets (starting with "asset:" or "assetdb://") as they don't exist on local filesystem
+            if filename is not None and filename.startswith(("asset:", "assetdb://")):
+                logger.warning(
+                    f"Excluding Maxon DB asset from job bundle: {filename}\n"
```
joel-wong-aws, line 65: Can we add a note describing what the behaviour will be if the user continues? i.e. that the assets will be download directly from Maxon during the render

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: 1. Yep, added!

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: Noted, added to the warning message

joel-wong-aws (approved):

karthikbekalp (approved):

PR #356, opened 2025-12-01
fix: resolve  token per-take for marked takes submissions

### LINKED ISSUE #350: Bug: Cinema4D multi-take output files overwritten
### Describe the bug

Image
Image

### Expected Behaviour

expected output folders:

### Current Behaviour

Example incorrect result:

### Reproduction Steps

- Mark takes for rendering

Image

### Environment

At minimum:
1. Operating system: MacOS 

vaish-gujjari:

Modified submitter to resolve output paths per-take instead of globally, and moved adaptor token resolution from initialization to render time when take context is available.

Fixes data loss bug where multiple marked takes would overwrite each other's frames. Each take now renders to its own directory as expected.

Added unit tests for per-take path resolution and verified with test scene containing multiple marked takes with child take selected.

Yes

If yes, please include the results of both the automated tests and any manual tests that you ran.
Also, check if there is a change to the `integ_test_helpers.py` file within cinema4d_submitter as a result of your submitter change.
Yes, submitter changes were made. Here are the test results:

Automated Tests:

Added 3 new unit tests in test_take_output_paths.py:
- test_multiple_marked_takes_resolve_paths_individually ✓
- test_child_take_selected_does_not_affect_other_takes ✓
- test_multipass_paths_resolved_per_take ✓

Manual Tests:
- Submitted test scene with 4 marked takes (V, E, R, A) with child take selected
- Verified asset references correctly include separate output directories for each take
- Confirmed job logs show all take-specific paths registered

integ_test_helpers.py Check:
No changes required to integ_test_helpers.py - the modifications to create_job_bundle() maintain the same function signature and return behavior, only changing internal path resolution logic.

No

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -236,7 +236,7 @@ class Cinema4DHandler:
             # Reloading the document will allow the next frame to have correct data.
             self._reload_document()
 
-        self.render_data = self.doc.GetActiveRenderData()
+        self.render_data = self._get_take_render_data(self.doc)
         self.render_data[c4d.RDATA_FRAMESEQUENCE] = c4d.RDATA_FRAMESEQUENCE_MANUAL
         self.render_kwargs[FRAME_KEY] = int(self.render_kwargs.get(FRAME_KEY, data[FRAME_KEY]))
         frame = self.render_kwargs[FRAME_KEY]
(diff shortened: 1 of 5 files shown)


src/deadline/cinema4d_submitter/scene.py
karthikbekalp, line 181: nit: Can we add a type hint for this function i.e. this returns a string?

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -188,6 +188,14 @@ def replace_render_path_tokens(path, doc=None, take=None, render_data=None):
         if render_data is None:
             render_data = Scene.get_render_data(doc=doc, take=take)
 
+        # Store the current take to restore it later
+        take_data = doc.GetTakeData()
+        original_take = take_data.GetCurrentTake() if take_data else None
+
+        # Temporarily set the document's current take to the one we're processing
+        if take and take_data:
+            take_data.SetCurrentTake(take)
```
karthikbekalp, line 197: Hmm, a little confused. Why are we doing this? 

Wouldn't `Scene.get_render_data(doc=doc, take=take)` already give the expected take in the render_data?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -478,24 +478,68 @@ def create_job_bundle(
...
+                return take
+            for child in take.GetChildren():
+                result = search_take(child, target_name)
+                if result:
+                    return result
+            return None
+
+        return search_take(main_take, name)
+
+    for take_data in submit_takes:
```
karthikbekalp, line 500: nit: I think we can simplify this code to just have one function that takes the submit_takes and the path and just creates a list of paths based on different take data. 

The function would look like : 

get_output_directories_for_takes(settings.override_output_path, submit_takes) 

And this function would return a list of all the directories with different take values. 

For example, where the submit takes have 3 takes named "my_take_1" to "my_take_3"
["/some_path/my_take_1/my output","/some_path/my_take_2/my output","/some_path/my_take_2/my output" ]

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -246,9 +246,21 @@ def start_render(self, data: dict) -> None:
         self.render_data[c4d.RDATA_FRAMETO] = frame_time
         self.render_data[c4d.RDATA_FRAMESTEP] = 1
 
-        if self.render_data[c4d.RDATA_PATH]:
+        # Resolve and set output paths with tokens like $take, $prj
```
karthikbekalp, line 249: Hmm, Cinema 4D should automatically resolve the tokens like $take, $prj while rendering. Does it not do that?

karthikbekalp (changes requested):
Added some comments. I'll discuss this PR offline with you.

PR #357, opened 2025-12-02
fix: resolve take-specific output paths and prevent symlink nesting i…

### LINKED ISSUE #350: Bug: Cinema4D multi-take output files overwritten

vaish-gujjari:

- Modified get_render_data() to temporarily switch to the queried take before getting render data, then restore the original take
- Added take-specific resolved output paths to each step's init_data in the job template
- Skipped symlinks during asset collection with followlinks=False and os.path.islink() checks
- Filtered output directories to only include paths under the document directory
- Removed output directories from auto-detected attachments and only add them for selected takes

Screenshot 2025-12-01 at 5 22 23 PM

Screenshot 2025-12-10 at 11 46 09 AM

- Multi-take submissions now correctly render each take to its own output folder regardless of which take is highlighted in the UI
- Prevents deeply nested symlink paths in job bundles
- Eliminates "unknown asset paths" warnings during submission
- Ensures only checked takes are submitted, not all takes in the document

- Tested multi-take submission with V, E, R, A checked while different takes were highlighted
- Verified each step renders to correct output folder (A→A, E→E, R→R, V→V)
- Confirmed symlink paths are no longer nested in session directories
- Validated asset_references.yaml contains only selected takes' output directories
- Unit tests pass with mock object handling

Yes

Yes

Yes

Screenshot 2025-12-08 at 2 33 40 PM

No

----


src/deadline/cinema4d_submitter/scene.py
```diff
@@ -205,6 +208,11 @@ def get_output_paths(take=None) -> Tuple[str, str]:
         Returns the default and multi-pass output paths.
         """
         doc = c4d.documents.GetActiveDocument()
+
+        # Only return output paths for marked takes or when no specific take is provided
+        if take is not None and not take.IsChecked():
```
joel-wong-aws: If the user selects "All takes", does that mean it includes unchecked takes? Is the `take.IsChecked()` condition required here?

test/unit/deadline_submitter_for_cinema4d/test_scene.py
```diff
@@ -24,9 +24,8 @@ def test_scene():
 
 
 def test_get_output_directores():
-    render_data = {0: 0}
-    doc = Scene.get_output_directories(render_data=render_data)
-    assert doc is not None
+    result = Scene.get_output_directories()
+    assert result is not None
```
joel-wong-aws: I think we need more unit tests for this, especially for the 4 different take selections (current, marked, all, main)

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -129,15 +129,18 @@ def renderer(render_data=None) -> str:
...
+    def get_output_directories(doc=None, take=None) -> set[str]:
         """
         Returns a list of directories files will be output to.
         """
-        doc = c4d.documents.GetActiveDocument()
+        if doc is None:
+            doc = c4d.documents.GetActiveDocument()
+
+        # Only return output directories for marked takes or when no specific take is provided
+        if take is not None and not take.IsChecked():
```
joel-wong-aws: Similar comment about the `take.IsChecked()` conditional as below.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -158,6 +158,15 @@ def _get_parameter_values(
     return parameter_values
 
 
+def get_child_takes(take):
+    child_takes = take.GetChildren()
+    all_takes = child_takes
+    if child_takes:
+        for child_take in child_takes:
+            all_takes.extend(get_child_takes(child_take))
+    return all_takes
```
joel-wong-aws: supernit: this function already existed and was just moved around. Can we put it back where it started to reduce merge conflicts?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -219,11 +235,21 @@ def _get_job_template(
...
+                    take_obj = t
+                    break
+
+            # Get the output paths for this specific take
+            output_path, multi_pass_path = Scene.get_output_paths(take=take_obj)
+
             # Update the init data of the step
             init_data = step["stepEnvironments"][0]["script"]["embeddedFiles"][0]
             init_data["data"] = (
-                "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '{{Param.OutputPath}}'\nmulti_pass_path: '{{Param.MultiPassPath}}'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
```
joel-wong-aws, line 231: Does this mean we need to remove the `{{Param.OutputPath}}` and `{{Param.MultiPassPath}}` job parameters from the job template?

joel-wong-aws (changes requested):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Hi Joel, since there's another function that's using get_child_takes function so I made it a regular function instead of a nested one..

src/deadline/cinema4d_submitter/scene.py
vaish-gujjari: Yep, all takes includes unchecked takes as well, Joel.. This is added just to avoid any confusion in the output path logs.. Not really, this is not a required change- I can revert this !

karthikbekalp:
I discussed with @vaish-gujjari offline, we can make this PR much simpler with only specific changes that's required to fix the issue. 

Feel free to ping me once you have the PR up. :)

test/unit/deadline_submitter_for_cinema4d/test_scene.py
vaish-gujjari: Added, Joel, thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: No, Joel.. modified the code in such a way that {{Param.OutputPath}} and {{Param.MultiPassPath}} will be used when "Main" take is selected and will be replaced with output path and multipasspath values when only "child" takes are selected. Thank you

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -205,6 +214,13 @@ def _get_job_template(
             take_frame_param["userInterface"]["groupLabel"] = take_data.ui_group_label
             job_template["parameterDefinitions"].append(take_frame_param)
 
+    # Get take objects for path resolution
+    doc = c4d.documents.GetActiveDocument()
+    take_data_obj = doc.GetTakeData()
+    main_take = take_data_obj.GetMainTake()
+
+    all_takes = [main_take] + get_child_takes(main_take)
```
karthikbekalp: Hmm, this is after the user has chosen the takes correct i.e. post clicking the submit button? 

We should already have the chosen takes in `takes` from this function argument no, do we need to retrieve this again?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -225,12 +241,30 @@ def _get_job_template(
...
-                "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '{{Param.OutputPath}}'\nmulti_pass_path: '{{Param.MultiPassPath}}'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
-                % take_data.name
-            )
+
+            # Check if paths contain $take token
+            if "$take" in settings.output_path or "$take" in settings.multi_pass_path:
+                # Get the output paths for this specific take
+                output_path, multi_pass_path = Scene.get_output_paths(take=take_obj)
+                init_data["data"] = (
+                    "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '%s'\nmulti_pass_path: '%s'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
```
karthikbekalp: Hmm, can we simplify this to something like: 

```
if "$take" in settings.output_path or "$take" in settings.multi_pass_path:

     init_data["data"] = (
                    "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '%s'\nmulti_pass_path: '%s'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
Comment view
                    % (take_data.name, output_path.replace("$take", take_data.name), multi_pass_path.replace("$take", take_data.name))
```

Is there a reason why we want to use "Scene.get_output_paths(take=take_obj)"?

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -194,7 +193,7 @@ def replace_render_path_tokens(path, doc=None, take=None, render_data=None):
             "_rBc": render_data.GetDataInstance(),
             "_frame": doc.GetTime().GetFrame(doc.GetFps()),
         }
-        if take:
+        if take is not None:
```
karthikbekalp: Curious, any reason why we are changing this expression to be explicitly checking for not None? Do we pass false values to this argument somewhere?

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -129,15 +129,14 @@ def renderer(render_data=None) -> str:
...
         Returns a list of directories files will be output to.
         """
-        doc = c4d.documents.GetActiveDocument()
+        if doc is None:
+            doc = c4d.documents.GetActiveDocument()
+
         doc_path = doc.GetDocumentPath()
-        if not take:
-            take_data = doc.GetTakeData()
-            take = take_data.GetCurrentTake()
```
karthikbekalp, line 140: Curious, any reason we removed this? We might need to have some defaults no?

karthikbekalp (changes requested):
Added some comments. Let me know if you want to discuss this offline. :)

src/deadline/cinema4d_submitter/scene.py
vaish-gujjari: Hi Karthik, the default value for take in the function replace_render_path_tokens is set to None. The "if take is not None:" check is added to ensure we are not passing None to the render_path_data dictionary.

karthikbekalp: > The "if take is not None:" check is added to ensure we are not passing None to the render_path_data dictionary

Hmm, not sure I understand. :)

Isn't `if take:` and `if take is not None:` equivalent as long as take is either None or an object? i.e. if the default value for take is None, shouldn't `if take:` also prevent passing None to the `render_path_dictionary`?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
joel-wong-aws: Ah, makes sense, thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -225,12 +241,30 @@ def _get_job_template(
...
+
             # Update the init data of the step
             init_data = step["stepEnvironments"][0]["script"]["embeddedFiles"][0]
-            init_data["data"] = (
-                "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '{{Param.OutputPath}}'\nmulti_pass_path: '{{Param.MultiPassPath}}'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
-                % take_data.name
-            )
+
+            # Check if paths contain $take token
+            if "$take" in settings.output_path or "$take" in settings.multi_pass_path:
```
joel-wong-aws: If we go into this conditional, should we also remove the `OutputPath` and `MultiPassPath` from the job template parameters, since they're unused?

test/unit/deadline_submitter_for_cinema4d/test_scene.py
```diff
@@ -23,10 +24,108 @@ def test_scene():
...
+            mock_c4d.RDATA_MULTIPASS_SAVEIMAGE = 3
+            mock_c4d.RDATA_MULTIPASS_FILENAME = 4
+            yield mock_c4d
+
+    @mock.patch("deadline.cinema4d_submitter.scene.Scene.get_render_data")
+    @mock.patch("deadline.cinema4d_submitter.scene.Scene.replace_render_path_tokens")
+    def test_no_take_with_saveimage(
+        self, mock_replace, mock_get_render, mock_doc, mock_render_data
+    ):
+        render_data = {1: True, 2: "output/image.png", 3: False}
```
joel-wong-aws: nit: can we use constants for the keys in this dictionary instead of 1, 2, 3, etc.? That would make it easier to tell what we're trying to test.

test/unit/deadline_submitter_for_cinema4d/test_scene.py
```diff
@@ -23,10 +24,108 @@ def test_scene():
...
+        render_data = {1: True, 2: "output/image.png", 3: True, 4: "multipass/mp.png"}
+        mock_get_render.return_value = render_data
+        mock_replace.side_effect = ["output/image.png", "multipass/mp.png"]
+
+        result = Scene.get_output_directories(doc=mock_doc)
+
+        assert result == {
+            os.path.normpath("/project/path/output"),
+            os.path.normpath("/project/path/multipass"),
+        }
```
joel-wong-aws, line 87: These 5 tests are all effectively testing the same thing now. When I was requesting more tests, I was imagining that we would test the differentiating flows where the behaviour is different. With the latest code, the biggest differentiated case is where $take is in the output path name - can we remove all but one of these tests (or change them to test the portion of the existing logic where the takes are retrieved), and then add a test for having `$take` in the output path?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Right, yep, we can! modifying the code.. good one, thanks, Joel

test/unit/deadline_submitter_for_cinema4d/test_scene.py
vaish-gujjari: yep, okay

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Ah, yes, not really! removed this block of code, thanks, Karthik

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Yep, simplified this logic, Karthik. The reason we were calling get_output_paths was to get the final path to replace $take and any other tokens that were present.. Now that the logic is much more simpler, we don't have to call the function to replace the tokens.. Thanks

PR #365, opened 2025-12-17
fix: replace  value in path and add warning for missing  for multiple…

### LINKED ISSUE #350: Bug: Cinema4D multi-take output files overwritten

vaish-gujjari:

- Modified get_render_data() to temporarily switch to the queried take before getting render data, then restore the original take
- Added take-specific resolved output paths to each step's init_data in the job template
- Skipped symlinks during asset collection with followlinks=False and os.path.islink() checks
- Filtered output directories to only include paths under the document directory
- Removed output directories from auto-detected attachments and only add them for selected takes

Screenshot 2025-12-01 at 5 22 23 PM

Screenshot 2025-12-10 at 11 46 09 AM

- Multi-take submissions now correctly render each take to its own output folder regardless of which take is highlighted in the UI
- Prevents deeply nested symlink paths in job bundles
- Eliminates "unknown asset paths" warnings during submission
- Ensures only checked takes are submitted, not all takes in the document

- Tested multi-take submission with V, E, R, A checked while different takes were highlighted
- Verified each step renders to correct output folder (A→A, E→E, R→R, V→V)
- Confirmed symlink paths are no longer nested in session directories
- Validated asset_references.yaml contains only selected takes' output directories
- Unit tests pass with mock object handling

Yes

Yes

No

No

----

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -206,6 +206,9 @@ def _get_job_template(
             take_frame_param["userInterface"]["groupLabel"] = take_data.ui_group_label
             job_template["parameterDefinitions"].append(take_frame_param)
 
+    # Check if paths contain $take token
+    has_take_token = "$take" in settings.output_path or "$take" in settings.multi_pass_path
+
     # Replicate the default step, once per render take, and adjust its settings
     default_step = job_template["steps"][0]
     job_template["steps"] = []
@@ -228,9 +231,17 @@ def _get_job_template(
         else:
             # Update the init data of the step
             init_data = step["stepEnvironments"][0]["script"]["embeddedFiles"][0]
+            output_path = settings.output_path
+            multi_pass_path = settings.multi_pass_path
+
+            if has_take_token:
+                # Replace $take token with actual take name, replacing spaces with underscores
+                take_name_for_path = take_data.name.replace(" ", "_")
+                output_path = settings.output_path.replace("$take", take_name_for_path)
+                multi_pass_path = settings.multi_pass_path.replace("$take", take_name_for_path)
             init_data["data"] = (
-                "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '{{Param.OutputPath}}'\nmulti_pass_path: '{{Param.MultiPassPath}}'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
-                % take_data.name
+                "scene_file: '{{Param.Cinema4DFile}}'\ntake: '%s'\noutput_path: '%s'\nmulti_pass_path: '%s'\nactivate_error_checking: '{{Param.ActivateErrorChecking}}'\nuse_cached_text: '{{Param.UseCachedText}}'"
+                % (take_data.name, output_path, multi_pass_path)
             )
 
     # If Arnold is one of the renderers, add Arnold-specific parameters
@@ -474,16 +485,8 @@ def create_job_bundle(
     if settings.export_job_bundle_to_temp and temp_dir:
         export_to_temp_folder(temp_dir, asset_references)
 
-    submit_takes = takes["main_data_list"]
     job_bundle_path = Path(job_bundle_dir)
-    if settings.take_selection == TakeSelection.MAIN:
-        submit_takes = takes["main_data_list"]
-    elif settings.take_selection == TakeSelection.ALL:
-        submit_takes = takes["take_data_list"]
-    elif settings.take_selection == TakeSelection.MARKED:
-        submit_takes = takes["marked_data_list"]
-    elif settings.take_selection == TakeSelection.CURRENT:
-        submit_takes = takes["current_data_list"]
+    submit_takes = get_submit_takes(settings, takes)
 
     # Add overrides to asset references and update the paths with C4D render path tokens.
     if settings.override_output_path:
@@ -655,6 +658,41 @@ def get_conda_packages(doc: Any) -> str:
     return packages
 
 
+def get_submit_takes(
+    settings: RenderSubmitterUISettings, takes: dict[str, list[TakeData]]
+) -> list[TakeData]:
+    """
+    Determine which takes will be submitted based on take selection setting.
+    """
+    if settings.take_selection == TakeSelection.MAIN:
+        return takes["main_data_list"]
+    elif settings.take_selection == TakeSelection.ALL:
+        return takes["take_data_list"]
+    elif settings.take_selection == TakeSelection.MARKED:
+        return takes["marked_data_list"]
+    elif settings.take_selection == TakeSelection.CURRENT:
+        return takes["current_data_list"]
+    return takes["main_data_list"]
+
+
+def check_take_token_warnings(
+    settings: RenderSubmitterUISettings, submit_takes: list[TakeData]
+) -> None:
+    """
+    Check if multiple takes are selected without $take token in output paths.
+    Adds a warning if output files will overwrite each other.
+    """
+    if (
+        len(submit_takes) > 1
+        and "$take" not in settings.output_path
+        and "$take" not in settings.multi_pass_path
+    ):
+        warning_collector.add_warning(
+            "Multiple takes are selected but output paths do not contain the $take token. "
+            "Output files from different takes will overwrite each other in the same destination folder."
+        )
+
+
 def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> None:
     """
     Exports the current Cinema 4D project to a temporary folder and updates the asset references.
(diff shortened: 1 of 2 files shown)


src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -655,6 +658,41 @@ def get_conda_packages(doc: Any) -> str:
...
+    settings: RenderSubmitterUISettings, submit_takes: list[TakeData]
+) -> None:
+    """
+    Check if multiple takes are selected without $take token in output paths.
+    Adds a warning if output files will overwrite each other.
+    """
+    if (
+        len(submit_takes) > 1
+        and "$take" not in settings.output_path
+        and "$take" not in settings.multi_pass_path
```
karthikbekalp: I think as we discussed offline this should be an "or" instead?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -655,6 +658,41 @@ def get_conda_packages(doc: Any) -> str:
...
+    ):
+        warning_collector.add_warning(
+            "Multiple takes are selected but output paths do not contain the $take token. "
+            "Output files from different takes will overwrite each other in the same destination folder."
```
karthikbekalp: nit: I think we should recommend customers to use "$take" in the output/multipass output path so that they can avoid overwriting? Something like : 

```
Output paths missing $take token. This will cause different takes will overwrite each other. Use $take in your path to avoid this.
```

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -655,6 +658,41 @@ def get_conda_packages(doc: Any) -> str:
...
+
+
+def check_take_token_warnings(
+    settings: RenderSubmitterUISettings, submit_takes: list[TakeData]
+) -> None:
+    """
+    Check if multiple takes are selected without $take token in output paths.
+    Adds a warning if output files will overwrite each other.
+    """
+    if (
```
karthikbekalp: nit: I think we can add guard clauses here? 
If the len(submit_takes) == 1, we return immediately?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -759,7 +797,10 @@ def on_create_job_bundle_callback(
         """
         Callback function for creating a job bundle when submitting the job.
         """
-        # Check for warnings and show warning dialog if needed
+        # Determine which takes will be submitted and check for warnings
+        submit_takes = get_submit_takes(settings, takes)
+        check_take_token_warnings(settings, submit_takes)
```
karthikbekalp: Hmm, also considering `check_take_token_warnings` already takes in `settings, takes` I think it would be cleaner if we move `submit_takes = get_submit_takes(settings, takes)` into `check_take_token_warning`. What do you think?

karthikbekalp (changes requested):
I think we are almost there! Thanks for your patience!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Hi, yes, this is now updated in the latest commit

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -655,6 +658,39 @@ def get_conda_packages(doc: Any) -> str:
...
+    """
+    Determine which takes will be submitted based on take selection setting.
+    """
+    if settings.take_selection == TakeSelection.MAIN:
+        return takes["main_data_list"]
+    elif settings.take_selection == TakeSelection.ALL:
+        return takes["take_data_list"]
+    elif settings.take_selection == TakeSelection.MARKED:
+        return takes["marked_data_list"]
+    elif settings.take_selection == TakeSelection.CURRENT:
```
karthikbekalp: nit: Should these be `if` instead as if any of the conditions above are true we would return immediately?

vaish-gujjari: yeah! modified..

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: yep, modified

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: modified..

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Good one ! modified!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -655,6 +658,40 @@ def get_conda_packages(doc: Any) -> str:
...
+    """
+    Check if multiple takes are selected without $take token in output paths.
+    Adds a warning if output files will overwrite each other.
+    """
+    submit_takes = get_submit_takes(settings, takes)
+    if len(submit_takes) == 1:
+        return
+    if "$take" not in settings.output_path or "$take" not in settings.multi_pass_path:
+        warning_collector.add_warning(
+            "Multiple takes are selected but output paths do not contain the $take token. "
```
karthikbekalp, line 702: nit: I think `Output paths missing $take token` the next line is redundant.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -759,7 +796,9 @@ def on_create_job_bundle_callback(
         """
         Callback function for creating a job bundle when submitting the job.
         """
-        # Check for warnings and show warning dialog if needed
+        # check for warnings
```
karthikbekalp: nit: I don't think this comment is necessary.

test/unit/deadline_submitter_for_cinema4d/test_cinema4d_render_submitter.py
```diff
@@ -198,3 +200,65 @@ def test_get_job_template_skips_font_manager_no_fonts(self, tmp_path):
...
+        settings.multi_pass_path = "/path/$take/multipass"
+        takes = {
+            "main_data_list": [
+                TakeData("Main", "Main", "standard", "", None, "1-10", set(), False),
+                TakeData("Take1", "Take1", "standard", "", None, "1-10", set(), False),
+            ]
+        }
+        check_take_token_warnings(settings, takes)
+
+        assert warning_collector.has_warnings()
```
karthikbekalp, line 248: nit: We should also assert that the warning we get is for "$take" but this is not a blocking issue.

karthikbekalp (approved):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks!

baxeaz (approved):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -655,6 +658,40 @@ def get_conda_packages(doc: Any) -> str:
...
+    Check if multiple takes are selected without $take token in output paths.
+    Adds a warning if output files will overwrite each other.
+    """
+    submit_takes = get_submit_takes(settings, takes)
+    if len(submit_takes) == 1:
+        return
+    if "$take" not in settings.output_path or "$take" not in settings.multi_pass_path:
+        warning_collector.add_warning(
+            "Multiple takes are selected but output paths do not contain the $take token. "
+            "This will cause different takes will overwrite each other. Use $take in your path to avoid this."
```
baxeaz, line 703: nit: cause different takes **to** overwrite each other

PR #374, opened 2026-01-02
docs: Add support page for raising issues or feature requests.

karthikbekalp:

We want to add some nice instructions for customers so that they can easily share their issues or raise issues. 

Added a support page. 

Customers have a nice page and instructions to follow to share the issues with us. 

Only docs. 

This is a docs change. 

Nope just a docs change. 
----

### DIFF
--- a/docs/user_guide/support.md
+++ b/docs/user_guide/support.md
@@ -8,18 +8,22 @@ Before reaching out for help, try these troubleshooting steps. They often resolv
 
 ### Troubleshooting Checklist
 
+- **✅ Render one frame locally** - Before submitting to the cloud, render at least one frame locally in Cinema 4D to verify your scene renders correctly. This helps identify scene-specific issues vs. cloud rendering issues.
+
 - **✅ Update to the latest submitter** - We release updates frequently with bug fixes and improvements. Your issue may already be fixed in a newer version. To check if you're running the latest version:
-  1. Find your current version: Open Deadline Cloud Monitor, select any submitted job, and look for **"Submitter Integration Version"** in the job properties panel
+  1. Find your current version: The version is displayed in the submitter window title
   2. Compare with the latest release: Visit the [releases page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/releases) to see the most recent version
   3. If your version is older, update the submitter and test again before reporting the issue
 
-- **✅ Save Cinema 4D Project with Assets** - Enable the **"Save Cinema 4D Project with Assets"** checkbox in the Job-Specific Settings tab of the submitter. This creates a temporary copy of your project with all assets and fixes file paths, helping identify missing files and organizing assets for render farms. [Learn more about this feature](submitter-features.md#job-specific-settings).
+- **✅ Check your Job-Specific Settings** - Review the Job-Specific Settings tab in the submitter. In particular, enable the **"Save Cinema 4D Project with Assets"** checkbox. This creates a temporary copy of your project with all assets and fixes file paths, helping identify missing files and organizing assets for render farms. [Learn more about this feature](submitter-features.md#job-specific-settings).
 
 - **✅ Try different Cinema 4D versions** - If you're experiencing issues, test with Cinema 4D 2024, 2025, or 2026 to see if the problem is version-specific.
 
-- **✅ Check existing GitHub issues** - Search the [GitHub Issues page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues) to see if someone else has already reported your problem and found a solution.
+- **✅ Check existing GitHub issues** - Search the [GitHub Issues page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues?q=is%3Aissue) to see if someone else has already reported your problem and found a solution.
 
 - **✅ Try a different OS fleet** - Submit your job to both Windows and Linux fleets if available. Windows generally has better support and compatibility for Cinema 4D features.
+
+- **✅ Review session logs** - Download and review the session logs from Deadline Cloud Monitor. These logs often contain error messages that generally pinpoint the issue.
 
 - **✅ Create a scene project file** - Use Cinema 4D's **File > Save Project with Assets** to create a self-contained project that includes all dependencies. Zip this file for easy sharing with support.
 
@@ -33,7 +37,7 @@ Contact AWS Support for:
 - AWS account issues
 - Billing questions
 - General AWS service questions
-- Deadline Cloud service availability
+- Deadline Cloud internal server errors in Deadline Cloud Monitor or while using CLI
 
 You can also report Cinema 4D submitter/adaptor issues through AWS Support, but note that these requests may take longer as they need to be routed to the maintainers of this repository. For faster response on Cinema 4D-specific issues, we recommend using GitHub Issues (see below).
 
@@ -57,19 +61,13 @@ You can also report Cinema 4D submitter/adaptor issues through AWS Support, but 
 
 **Before creating a new bug report:**
 
-1. [Search existing bugs](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues?q=is%3Aopen+label%3Abug) to see if your problem has already been reported
+1. [Search existing bugs](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues?q=label%3Abug) to see if your problem has already been reported or fixed in latest versions.
 2. If you find an existing issue that matches your problem:
    - Add a 👍 reaction to help us prioritize
    - Comment with any additional details or reproduction steps you can provide
    - This helps us understand how many users are affected
 
-**If no existing issue matches**, create a new issue with the "bug" label and include:
-
-1. **Clear description** - What happened vs. what you expected
-2. **Steps to reproduce** - Detailed steps so others can recreate the issue
-3. **Environment details** - See the checklist below
-4. **Error messages** - Full error text, not just screenshots
-5. **Logs** - Relevant log files (see below for how to gather them)
+**If no existing issue matches**, [create a new bug report](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/new?template=bug.yml) using our bug report template. The template will guide you through providing all the necessary information.
 
 ### Feature Requests
 
@@ -97,7 +95,7 @@ When contacting support or creating a GitHub issue, always include:
 
 - **Cinema 4D version** - e.g., Cinema 4D 2025.1.0
 - **Operating System** - e.g., Windows 11, macOS 14.2
-- **Submitter version** - Found in Extensions > AWS Deadline Cloud Submitter > About
+- **Submitter version** - Copy the entire contents of the About panel (Extensions > AWS Deadline Cloud Submitter > About)
 - **Renderer** - Standard, Redshift, Arnold, etc.
 - **Fleet configuration** - Worker OS (Windows/Linux), memory requirements, disk space, GPU type if using Redshift
 - **Error messages** - Complete error text, not paraphrased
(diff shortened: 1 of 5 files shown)


docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
+# Getting Support
+
+Need help with the Cinema 4D Deadline Cloud submitter? This page will guide you through troubleshooting steps and how to get support when you need it.
+
+## Before You Contact Support
+
+Before reaching out for help, try these troubleshooting steps. They often resolve common issues and will help you provide better information if you do need to contact support.
+
+### Troubleshooting Checklist
+
```
joel-wong-aws, line 10: Should we add a checklist item to download and review the session logs, including the launch Cinema 4D step?

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+  2. Compare with the latest release: Visit the [releases page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/releases) to see the most recent version
+  3. If your version is older, update the submitter and test again before reporting the issue
+
+- **✅ Save Cinema 4D Project with Assets** - Enable the **"Save Cinema 4D Project with Assets"** checkbox in the Job-Specific Settings tab of the submitter. This creates a temporary copy of your project with all assets and fixes file paths, helping identify missing files and organizing assets for render farms. [Learn more about this feature](submitter-features.md#job-specific-settings).
+
+- **✅ Try different Cinema 4D versions** - If you're experiencing issues, test with Cinema 4D 2024, 2025, or 2026 to see if the problem is version-specific.
+
+- **✅ Check existing GitHub issues** - Search the [GitHub Issues page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues) to see if someone else has already reported your problem and found a solution.
```
joel-wong-aws: Can we add a note to include closed issues in the search?

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+
+[Open a GitHub Issue](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues)
+
+## How to Report Issues on GitHub
+
+### Bug Reports
+
+**Before creating a new bug report:**
+
+1. [Search existing bugs](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues?q=is%3Aopen+label%3Abug) to see if your problem has already been reported
```
joel-wong-aws: Should the search include closed bug reports in case they are fixed in newer submitter versions or have workarounds?

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+2. If you find an existing issue that matches your problem:
+   - Add a 👍 reaction to help us prioritize
+   - Comment with any additional details or reproduction steps you can provide
+   - This helps us understand how many users are affected
+
+**If no existing issue matches**, create a new issue with the "bug" label and include:
```
joel-wong-aws: We have an actual "Bug Report" template, it's not just a label :)

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+
+1. **Clear description** - What happened vs. what you expected
+2. **Steps to reproduce** - Detailed steps so others can recreate the issue
+3. **Environment details** - See the checklist below
+4. **Error messages** - Full error text, not just screenshots
+5. **Logs** - Relevant log files (see below for how to gather them)
```
joel-wong-aws: These sections are all included in the bug report, I don't think we need to duplicate them here - if there's anything here not in the bug report, we should probably add it to the bug report e.g. prompting the user for logs

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+Need help with the Cinema 4D Deadline Cloud submitter? This page will guide you through troubleshooting steps and how to get support when you need it.
+
+## Before You Contact Support
+
+Before reaching out for help, try these troubleshooting steps. They often resolve common issues and will help you provide better information if you do need to contact support.
+
+### Troubleshooting Checklist
+
+- **✅ Update to the latest submitter** - We release updates frequently with bug fixes and improvements. Your issue may already be fixed in a newer version. To check if you're running the latest version:
+  1. Find your current version: Open Deadline Cloud Monitor, select any submitted job, and look for **"Submitter Integration Version"** in the job properties panel
```
joel-wong-aws: This is now in the title of the submitter! No need to go into the properties panel

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+  2. Compare with the latest release: Visit the [releases page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/releases) to see the most recent version
+  3. If your version is older, update the submitter and test again before reporting the issue
+
+- **✅ Save Cinema 4D Project with Assets** - Enable the **"Save Cinema 4D Project with Assets"** checkbox in the Job-Specific Settings tab of the submitter. This creates a temporary copy of your project with all assets and fixes file paths, helping identify missing files and organizing assets for render farms. [Learn more about this feature](submitter-features.md#job-specific-settings).
```
joel-wong-aws: Can we make this more generic to say "check your Job-specific settings" and then we can mention this option in particular?

docs/user_guide/support.md
joel-wong-aws, line 10: Should we add a checklist item to run one frame locally to verify the scene is correctly rendering locally?

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+
+Different types of issues should be directed to different support channels:
+
+### AWS General Support
+
+Contact AWS Support for:
+- AWS account issues
+- Billing questions
+- General AWS service questions
+- Deadline Cloud service availability
```
joel-wong-aws: Can we define this more closely? What does this look like? InternalServerError only?

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+
+## What to Include in Your Support Request
+
+### Required Information Checklist
+
+When contacting support or creating a GitHub issue, always include:
+
+- **Cinema 4D version** - e.g., Cinema 4D 2025.1.0
+- **Operating System** - e.g., Windows 11, macOS 14.2
+- **Submitter version** - Found in Extensions > AWS Deadline Cloud Submitter > About
```
joel-wong-aws: Should we suggest that they copy the whole contents of the about tab to the ticket?

docs/user_guide/support.md
```diff
@@ -0,0 +1,140 @@
...
+
+1. In the Cinema 4D submitter, check **"Activate detailed logging"** in Job-Specific Settings
+2. Submit your job
+3. After the job completes (or fails), retrieve the logs:
+   - Open Deadline Cloud Monitor
+   - Navigate to your job
+   - Right-click on a task → **View logs**
+   - Enable **"View logs for all tasks"**
+   - Look for the "Shut down DetailedLogging" section
+   - Click **"Download logs"** to save the log files locally for sharing with support
```
joel-wong-aws: Can we navigate them to just download the session logs instead?

joel-wong-aws (changes requested):

docs/user_guide/support.md
karthikbekalp: Sure, I assumed they would have done that but its best to be explicit.

karthikbekalp: Sure, we can add that. Best to be explicit.

docs/user_guide/support.md
karthikbekalp: You are right! I forgot about that. I'll update this. Just quoting the title from the submitter should be sufficient...

docs/user_guide/support.md
karthikbekalp: Cool, we can do that.

docs/user_guide/support.md
karthikbekalp: We could although as we already request the customers to use the latest submitter, just looking at the open issues should be sufficient. But if it can help customers then why not.

docs/user_guide/support.md
karthikbekalp: I answered this above, but sure why not.

docs/user_guide/support.md
karthikbekalp: Makes sense, I'll update this.

docs/user_guide/support.md
karthikbekalp: Makes sense. I'll update the Bug report to include this.

docs/user_guide/support.md
karthikbekalp: Yup, we could do that.

docs/user_guide/support.md
karthikbekalp: Makes sense. I'll update this to instruct to download the entire session logs from Download logs > Entire session .log

docs/user_guide/support.md
karthikbekalp: I'll rephraze to suggest if there are errors encountered in Deadline cloud monitor or in deadline CLI commands.

karthikbekalp (commented):
Thanks for your comments, making changes right away.

docs/user_guide/support.md
joel-wong-aws: Sometimes, customers run into problems where there is an existing solution but it is hidden or not immediately obvious as the solution. Searching through closed issues can help in this case!

docs/user_guide/support.md
```diff
@@ -0,0 +1,135 @@
...
+  3. If your version is older, update the submitter and test again before reporting the issue
+
+- **✅ Check your Job-Specific Settings** - Review the Job-Specific Settings tab in the submitter. In particular, enable the **"Save Cinema 4D Project with Assets"** checkbox. This creates a temporary copy of your project with all assets and fixes file paths, helping identify missing files and organizing assets for render farms. [Learn more about this feature](submitter-features.md#job-specific-settings).
+
+- **✅ Try different Cinema 4D versions** - If you're experiencing issues, test with Cinema 4D 2024, 2025, or 2026 to see if the problem is version-specific.
+
+- **✅ Check existing GitHub issues** - Search the [GitHub Issues page](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues?q=is%3Aissue) to see if someone else has already reported your problem and found a solution.
+
+- **✅ Try a different OS fleet** - Submit your job to both Windows and Linux fleets if available. Windows generally has better support and compatibility for Cinema 4D features.
```
joel-wong-aws: Since these users aren't technical, they might not know what "OS" stands for. Would it be better to title this "Try a different fleet operating system"?

docs/user_guide/support.md
```diff
@@ -0,0 +1,135 @@
...
+
+**Before creating a new feature request:**
+
+1. [Search existing enhancements](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues?q=is%3Aopen+label%3Aenhancement) to see if someone has already suggested your idea
+2. If you find an existing request that matches:
+   - Add a 👍 reaction to show your support
+   - Comment with your specific use case - this strengthens the request and helps us understand different needs
+   - The more users who express interest, the higher priority it receives
+
+**If no existing request matches**, create a new issue with the "enhancement" label and explain:
```
joel-wong-aws: Similar to bug reports, we have a full "Feature Request" GitHub issue template which has these sections

joel-wong-aws (dismissed):

docs/user_guide/support.md
karthikbekalp: Updated!

docs/user_guide/support.md
karthikbekalp: Updated!

joel-wong-aws (approved):

PR #378, opened 2026-01-08
fix: use per-take output paths when $take token is present in multi-take submissions

### LINKED ISSUE #350: Bug: Cinema4D multi-take output files overwritten

vaish-gujjari:

ntegration tests were failing because the code started using the $take token in output paths. When $take is present in the path, the code needs to replace it with the actual take name and embed the resolved paths directly in the init data. However, when $take is not present, the code should use parameter references like {{Param.OutputPath}} instead of hardcoding paths.

Added a has_take_token parameter to _get_job_template() that determines whether to replace $take tokens with actual take names (and embed resolved paths) or use parameter references. When has_take_token=True, the code replaces $take with the sanitized take name and embeds the full paths. When has_take_token=False, it uses parameter references like {{Param.OutputPath}} and {{Param.MultiPassPath}}.

Integration tests now pass because the job template correctly handles both scenarios: scenes with $take tokens in paths (which get resolved per-take) and scenes without $take tokens (which use parameter references). This maintains backward compatibility while supporting the new $take token feature.

Verified that integration tests pass with the changes, and manually tested both scenarios: submissions with $take in output paths and submissions without $take tokens.

No

Yes
Screenshot 2026-01-08 at 10 48 45 AM

Yes
Screenshot 2026-01-08 at 11 09 46 AM

Screenshot 2026-01-08 at 11 19 16 AM

No

No

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -182,6 +182,7 @@ def _get_job_template(
     settings: RenderSubmitterUISettings,
     renderers: set[str],
     takes: list[TakeData],
+    has_take_token: bool = False,
 ) -> dict[str, Any]:
     if os.getenv("DEADLINE_COMMAND_TEMPLATE"):
         template = "default_cinema4d_job_template.yaml"
@@ -219,9 +220,6 @@ def _get_job_template(
             take_frame_param["userInterface"]["groupLabel"] = take_data.ui_group_label
             job_template["parameterDefinitions"].append(take_frame_param)
 
-    # Check if paths contain $take token
-    has_take_token = "$take" in settings.output_path or "$take" in settings.multi_pass_path
-
     # Replicate the default step, once per render take, and adjust its settings
     default_step = job_template["steps"][0]
     job_template["steps"] = []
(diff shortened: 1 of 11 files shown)


src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -182,6 +182,7 @@ def _get_job_template(
     settings: RenderSubmitterUISettings,
     renderers: set[str],
     takes: list[TakeData],
+    has_take_token: bool = False,
```
karthikbekalp: Curious, any reason we modified this over here?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -423,7 +432,7 @@ def get_child_takes(take):
 
     for take in all_takes:
         take_name = take.GetName()
-        display_name = take_name[:64]
+        display_name = _STRIPPED_DISPLAY_CHARS.sub("_", take_name)[:64]
```
joel-wong-aws: Is there a reason we need to change the display name?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
joel-wong-aws, line 185: It's not clear to me what we're trying to achieve with this PR. The title is `fix: update integ tests...`, but this has a functional effect (i.e. it's not just integ tests that are changing), so the commit doesn't represent the changes. Are we trying to fix a bug or just make integration tests work?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -244,18 +242,25 @@ def _get_job_template(
...
-            output_path = settings.output_path
-            multi_pass_path = settings.multi_pass_path
 
             if has_take_token:
-                # Replace $take token with actual take name, replacing spaces with underscores
-                take_name_for_path = take_data.name.replace(" ", "_")
+                # Replace $take token with sanitized take name for file path safety
+                # Use original name (not truncated display_name) but sanitize it
+                take_name_for_path = _STRIPPED_DISPLAY_CHARS.sub("_", take_data.name).replace(
+                    " ", "_"
```
joel-wong-aws: Can we simplify by adding space to the list of stripped display chars?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: We are trying to fix a bug that's causing the integ tests to fail, Joel. 
We were passing the parsed paths to initData for all paths including the ones which doesn't have $take.. to handle that, I added a new variable has_take_token where in we'er checking for $take in the paths before we pass them into the _get_job_template function and update the initData in such a way that if there is no $take in the path, we keep the initData as is (passing Param.OutputPath) and only modify it when there is a $take token.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: I tried doing this, Joel but the physical multi takes integ test is failing as the display_name is used for step names and should preserve spaces (for testcase lik: "he llo", but take_name_for_path should replace spaces.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
joel-wong-aws: If there is a bug we're trying to fix, can we update the commit message (and PR title) to reflect that? This change is more than just fixing integration tests, it contains functional changes, and when developers read the commit message, it should be ideal if they can easily identify what changed.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
joel-wong-aws: We (at least now) only use `_STRIPPED_DISPLAY_CHARS` in one place, right? So the display_name concerns aren't relevant anymore? Also, since this isn't related to display names anymore, can we rename it to `_STRIPPED_PATH_CHARS`

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Makes sense, will update the commit message !

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: ah, yep, I can update that

joel-wong-aws (approved):

PR #396, opened 2026-02-16
fix: warn on .rs file detection and add FAQ for nested Redshift proxy files

### LINKED ISSUE #270: Bug: Redshift Proxy of Proxies fails with job attachments
### Describe the bug

When a Redshift proxy file is used(ProxyA) that contains another Redshift proxy file ( ProxyB), when you submit and render Deadline fails to find ProxyB (not caught in JA) and will will no render it. 

### Workaround

If you add the folder where the proxies live (in my case the same level as the scene file) as a job attachments input FOLDER, then the render succeeds.

### Expected Behaviour

The JA process should find the redshift proxies if possible and remap them.

### Current Behaviour

Fails to render proxy of proxies.

### Reproduction Steps

1. Make some objects.
1. Export them to a Redshift file (rs file)
1. Make a new scene.
1. Create a reshift proxy, and put it to the above.
1. Export that into a new Redshift proxy file .
1. Make a new scene.
1. Create a Redshift proxy of the 2nd proxy-of-proxies.
1. Render as Redshift and submit.

### Environment

At minimum:
1. Operating system Windows
2. Version of Cinema 4D 2025.3.2
3. Version of this package 
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

vaish-gujjari:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/270

When a Redshift proxy file (.rs) references another nested .rs proxy file, the Cinema 4D submitter's job attachments process does not detect or include the nested dependency. This causes renders to fail on Deadline Cloud because the nested proxy (e.g., ProxyB) cannot be found on the worker. The Cinema 4D SDK cannot read .rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either, so automatic detection is not feasible.

Since we cannot programmatically resolve nested .rs dependencies, we removed the previous nested proxy scanning logic (which attempted to parse binary .rs files for references) and replaced it with a simple, always-on warning. Now, whenever at least one .rs file is detected in the scene assets, the submitter logs a warning informing the user that nested .rs files will not be automatically included in the job bundle and that they should manually verify and add any nested proxy dependencies. A corresponding FAQ entry was also added to the user guide explaining this limitation.

**Link to the forum post:** https://developers.maxon.net/forum/topic/16370/cinema-4d-redshift-nested-redshift-proxy-files-not-detected-by-project-asset-inspector/2

Screenshot 2026-02-16 at 12 15 59 PM

Users will now be proactively warned about the nested proxy limitation every time they submit a scene containing .rs files, rather than silently failing at render time. This gives them the opportunity to manually add nested dependencies (e.g., by including the proxy folder as a job attachments input folder) before submission.

Tested by submitting nested proxies, .rs files, non .rs files, and mixed asset scenes in Cinema 4D. Unit tests were added covering: warning is shown when at least one .rs file is present, no warning when no .rs files exist, warning fires with mixed asset types including .rs, and case-insensitive .RS extension detection.

Yes
Screenshot 2026-02-16 at 1 03 50 PM

Yes
Screenshot 2026-02-16 at 2 08 33 PM

Yes

Yes

No

### DIFF
--- a/docs/user_guide/faq-and-glossary.md
+++ b/docs/user_guide/faq-and-glossary.md
@@ -92,7 +92,7 @@ A: The submitter includes built-in error detection to catch common issues like m
 
 **Q: Are nested Redshift proxy files detected when submitting Cinema 4D jobs to Deadline?**
 
-A: No, Redshift proxy files are not detected when submitting Cinema 4D jobs to Deadline. This is a limitation of the Redshift .rs file format. When you export a Cinema 4D scene containing RS Proxy objects to.rs, all referenced proxy data is flattened/inlined into a single file - no external references are preserved. The Cinema 4D SDK cannot read.rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either.
+A: No, Redshift proxy files are not detected when submitting Cinema 4D jobs to Deadline. This is a limitation of the Redshift *.rs file format. When you export a Cinema 4D scene containing RS Proxy objects to *.rs, all referenced proxy data is flattened/inlined into a single file - no external references are preserved. The Cinema 4D SDK cannot read *.rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either.
 
 **Q: How do I enable detailed logging for debugging rendering issues?**
 
--- a/docs/user_guide/faq-and-glossary.md
+++ b/docs/user_guide/faq-and-glossary.md
@@ -93,6 +93,7 @@ A: The submitter includes built-in error detection to catch common issues like m
 **Q: Are nested Redshift proxy files detected when submitting Cinema 4D jobs to Deadline?**
 
 A: No, Redshift proxy files are not detected when submitting Cinema 4D jobs to Deadline. This is a limitation of the Redshift *.rs file format. When you export a Cinema 4D scene containing RS Proxy objects to *.rs, all referenced proxy data is flattened/inlined into a single file - no external references are preserved. The Cinema 4D SDK cannot read *.rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either.
+Link to the forum post: https://developers.maxon.net/forum/topic/16370/cinema-4d-redshift-nested-redshift-proxy-files-not-detected-by-project-asset-inspector/2
 
 **Q: How do I enable detailed logging for debugging rendering issues?**
 
--- a/docs/user_guide/faq-and-glossary.md
+++ b/docs/user_guide/faq-and-glossary.md
@@ -90,6 +90,10 @@ A: Yes, you can set job priority levels in the submitter to control render queue
 
 A: The submitter includes built-in error detection to catch common issues like missing assets before submission. This can be deactivated in the submitter.
 
+**Q: Are nested Redshift proxy files detected when submitting Cinema 4D jobs to Deadline?**
+
+A: No, Redshift proxy files are not detected when submitting Cinema 4D jobs to Deadline. This is a limitation of the Redshift .rs file format. When you export a Cinema 4D scene containing RS Proxy objects to.rs, all referenced proxy data is flattened/inlined into a single file - no external references are preserved. The Cinema 4D SDK cannot read.rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either.
+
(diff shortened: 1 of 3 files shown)


docs/user_guide/faq-and-glossary.md
```diff
@@ -90,6 +90,10 @@ A: Yes, you can set job priority levels in the submitter to control render queue
 
 A: The submitter includes built-in error detection to catch common issues like missing assets before submission. This can be deactivated in the submitter.
 
+**Q: Are nested Redshift proxy files detected when submitting Cinema 4D jobs to Deadline?**
+
+A: No, Redshift proxy files are not detected when submitting Cinema 4D jobs to Deadline. This is a limitation of the Redshift .rs file format. When you export a Cinema 4D scene containing RS Proxy objects to.rs, all referenced proxy data is flattened/inlined into a single file - no external references are preserved. The Cinema 4D SDK cannot read.rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either.
```
karthikbekalp: nit: `to *.rs,` similarly, `read *.rs files`  (Also Grammar-wise, do we need to add "The" for Cinema 4D SDK?

src/deadline/cinema4d_submitter/assets.py
```diff
@@ -71,6 +71,16 @@ def parse_scene_assets(self) -> set[Path]:
             if exists is True and filename is not None:
                 assets.add(Path(filename))
 
+        # Warn if any Redshift proxy files are present in assets
+        if any(str(asset).lower().endswith(".rs") for asset in assets):
```
karthikbekalp: nit: Its always better to abstract this to a separate function so that we don't keep increasing the size of this function...

karthikbekalp (commented):
Added some nit comments but on the whole looks good to me. 
Don't forget to run integration tests! (Please add the integration test report into the PR description)

docs/user_guide/faq-and-glossary.md
vaish-gujjari: Changed .rs → *.rs in three places, added missing spaces before *.rs, and yes — "The Cinema 4D SDK" is grammatically correct as-is since it's referring to a specific SDK.

src/deadline/cinema4d_submitter/assets.py
vaish-gujjari: Extracted the .rs warning logic into _warn_if_redshift_proxies_detected() and parse_scene_assets now just calls it with a single line.

docs/user_guide/faq-and-glossary.md
```diff
@@ -90,6 +90,10 @@ A: Yes, you can set job priority levels in the submitter to control render queue
 
 A: The submitter includes built-in error detection to catch common issues like missing assets before submission. This can be deactivated in the submitter.
 
+**Q: Are nested Redshift proxy files detected when submitting Cinema 4D jobs to Deadline?**
+
+A: No, Redshift proxy files are not detected when submitting Cinema 4D jobs to Deadline. This is a limitation of the Redshift *.rs file format. When you export a Cinema 4D scene containing RS Proxy objects to *.rs, all referenced proxy data is flattened/inlined into a single file - no external references are preserved. The Cinema 4D SDK cannot read *.rs files to discover nested dependencies, and the Redshift Core doesn't expose this functionality either.
+
```
karthikbekalp, line 97: Forgot to tell this before, but can we add the forum post link here?

vaish-gujjari: Added

karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/assets.py
karthikbekalp: Thanks!

docs/user_guide/faq-and-glossary.md
karthikbekalp: Thanks!

karthikbekalp (approved):

crowecawcaw (approved):

PR #399, opened 2026-03-03
fix: auto-deduplicate duplicate take names instead of raising an error

### LINKED ISSUE #398: Bug: Ensure customers either get a warning message or make the take names unique while submitting
### Describe the bug

Currently, if a customer creates multiple takes with the same name, then we get an error after submitting to Deadline Cloud. 

### Expected Behaviour

We should either: 

- Give customers a helpful message to make sure their take names are unique. 
- Append some random characters at the end to make them unique. 

### Current Behaviour

Image

### Reproduction Steps

See the current behavior steps to reproduce. 

### Environment

This affects all environments. 

### Logs

_No response_

vaish-gujjari:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/398

When a customer creates multiple takes with the same name in Cinema 4D, submission to Deadline Cloud fails with a RuntimeError. This is a poor user experience — the error is unhelpful and blocks submission entirely.

Refactored deduplicate_take_names by extracting its internal logic into single-responsibility helper functions:

- _find_duplicate_take_names — identifies which take names appear more than once.
- _validate_duplicate_name_lengths — raises a RuntimeError if any duplicated take name is already at the 64-character limit, since a suffix can't be appended without truncating the display name.
- _generate_unique_name — finds the next available _{n} suffix that doesn't collide with any existing take name.
- _apply_take_name — writes the new name onto the take and truncates display_name to 64 characters.
- deduplicate_take_names now orchestrates these helpers: find duplicates, validate lengths, then loop through takes and rename collisions. The behavior is identical — e.g. take, take, take_1 still correctly produces take_2, take_3, take_1 — but each piece is independently testable and easier to reason about.

The callback calls warn_duplicate_take_names before the warning dialog, so the user sees the dedup warning alongside any other warnings and can choose to continue or cancel.

Duplicate take names no longer block submission with a RuntimeError. Instead, they are automatically made unique and the user is informed via the existing warning dialog. No changes to the adaptor, init-data, or run-data.

16 unit tests across two test classes:

TestDeduplicateTakeNames (14 tests):

1. test_no_duplicates_no_changes — unique takes are left unchanged
2. test_duplicate_names_get_suffixed — two identical names become MyTake_1, MyTake_2
3. test_three_duplicates_get_suffixed — three identical names get _1, _2, _3
4. test_multiple_duplicate_groups — separate duplicate groups (A, B) are suffixed independently
5. test_mixed_unique_and_duplicate — unique takes stay unchanged, only duplicates get suffixed
6. test_display_name_updated_on_dedup — display_name is updated alongside name
7. test_dedup_does_not_set_frames_parameters — dedup leaves frames_parameter_name as None
8. test_suffix_collision_with_existing_name — my_take, my_take, my_take_1 correctly produces my_take_2, my_take_3, my_take_1 with no collisions
9. test_frames_parameters_set_after_dedup — generate_take_parameter_names produces unique frame parameter names after deduplication
10. test_empty_list_no_error — empty input doesn't error
11. test_single_take_no_changes — single take is left unchanged
12. test_duplicate_name_at_64_chars_raises_error — RuntimeError raised when duplicate name is already 64 chars
13. test_duplicate_name_at_63_chars_does_not_raise — 63-char duplicates are suffixed normally without error
14. test_display_name_truncated_to_64_chars — display_name is capped at 64 chars when suffix pushes it over

TestWarnDuplicateTakeNames (2 tests):

1. test_no_duplicates_no_warning — no warning when all names are unique
2. test_duplicate_names_warning_is_added — warning is emitted mentioning the duplicated name and the _1, _2 suffixes

Yes.

Screenshot 2026-03-03 at 10 25 45 AM

Yes.
Screenshot 2026-03-03 at 10 53 13 AM

Yes. Changes are limited to generate_take_parameter_names in cinema4d_render_submitter.py and the on_create_job_bundle_callback that calls it. No changes to integ_test_helpers.py.

Screenshot 2026-03-04 at 9 27 00 AM

Screenshot 2026-03-04 at 9 26 47 AM

Screenshot 2026-03-04 at 9 38 00 AM

Screenshot 2026-03-04 at 2 34 37 PM

No.

No.

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -608,7 +608,7 @@ def deduplicate_take_names(submit_takes: list[TakeData]) -> None:
     _validate_duplicate_name_lengths(duplicated_names)
 
     all_names = {take.name for take in submit_takes}
-    next_suffix: dict[str, int] = {name: 1 for name in duplicated_names}
+    next_suffix: dict[str, int] = dict.fromkeys(duplicated_names, 1)
 
     for take in submit_takes:
         if take.name in next_suffix:
@@ -901,7 +901,6 @@ def _show_submitter(temp_dir: str, parent=None, f=Qt.WindowFlags()):
         """
         Callback function for creating a job bundle when submitting the job.
         """
-        # Check for warnings before submission
         submit_takes = get_submit_takes(settings, takes)
         warn_duplicate_take_names(submit_takes)
 
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -542,10 +542,9 @@ def create_job_bundle(
         take.frame_range != first_frame_range for take in submit_takes
     )
 
-    # If there are multiple frame ranges and we're not overriding the range,
-    # then we create per-take Frames parameters.
-    if per_take_frames_parameters:
-        generate_take_parameter_names(submit_takes)
+    # Always deduplicate take names; only generate per-take Frames parameters
+    # when there are multiple frame ranges and we're not overriding the range.
+    generate_take_parameter_names(submit_takes, set_frames_parameters=per_take_frames_parameters)
 
     renderers: set[str] = {take_data.renderer_name for take_data in submit_takes}
     job_template = _get_job_template(settings, renderers, submit_takes, has_take_token)
@@ -586,38 +585,56 @@ def create_job_bundle(
     }
 
 
-def generate_take_parameter_names(submit_takes: list[TakeData]) -> None:
+def generate_take_parameter_names(
+    submit_takes: list[TakeData], set_frames_parameters: bool = True
+) -> None:
     """
-    This function generates unique take frame range parameter names
-    by combining each takes name with a unique suffix (if required)
-    while still meeting the requirements (letters+numbers+underscores,
-    max 64 chars, etc.) of a parameter name.
+    This function deduplicates take names (appending _1, _2, etc. for duplicates)
+    and generates unique take frame range parameter names by combining each take's
+    name with a unique suffix (if required) while still meeting the requirements
+    (letters+numbers+underscores, max 64 chars, etc.) of a parameter name.
+
+    If set_frames_parameters is False, only deduplication is performed and
+    frames_parameter_name is not set on the takes.
 
     The frame parameter names are saved to the input submit_takes.
     """
+
+    # Deduplicate take names by appending _1, _2, etc. for duplicates
+    name_counts: dict[str, int] = {}
+    for take in submit_takes:
+        name_counts[take.name] = name_counts.get(take.name, 0) + 1
+
+    duplicated_names = {name for name, count in name_counts.items() if count > 1}
+    if duplicated_names:
+        next_suffix: dict[str, int] = {name: 1 for name in duplicated_names}
+        for take in submit_takes:
+            if take.name in next_suffix:
+                suffix = next_suffix[take.name]
+                new_name = f"{take.name}_{suffix}"
+                take.name = new_name
+                take.display_name = new_name[:64]
+                next_suffix[new_name.rsplit("_", 1)[0]] = suffix + 1
+
+        renamed_list = ", ".join(f"'{name}'" for name in sorted(duplicated_names))
+        warning_collector.add_warning(
+            f"Duplicate take names were found: {renamed_list}. "
+            "They have been automatically renamed with _1, _2, etc. suffixes to ensure uniqueness."
+        )
+
+    if not set_frames_parameters:
+        return
 
     # parameter names must start with a letter or underscore
     allowed_first_job_parameter_chars = re.compile("[a-zA-Z_]")
     # parameter names must only contain letters, numbers, or underscores
     removed_job_parameter_chars = re.compile("[^a-zA-Z0-9_]")
 
-    take_names = set()
     parameter_names = set()
 
     for take_number in range(len(submit_takes)):
         take_data = submit_takes[take_number]
-
-        # First, check for duplicate take names since this will result in overwriting files in the output
-        # or other unexpected behaviour
-        # We do this here rather than earlier in submission because we get an error popup
-        # (rather than a quieter console error) for errors here.
         take_name = take_data.name
-        if take_name in take_names:
-            raise RuntimeError(
-                f"You have multiple takes named '{take_name}' with different render settings among the takes. "
-                "Please use unique take names."
-            )
-        take_names.add(take_name)
 
         # Now, determine the frame parameter name
         # remove all disallowed characters
@@ -834,6 +851,10 @@ def _show_submitter(temp_dir: str, parent=None, f=Qt.WindowFlags()):
         """
         Callback function for creating a job bundle when submitting the job.
         """
+        # Deduplicate take names early so the warning shows in the dialog
+        submit_takes = get_submit_takes(settings, takes)
+        generate_take_parameter_names(submit_takes, set_frames_parameters=False)
+
         check_take_token_warnings(settings, takes)
 
         if warning_collector.has_warnings():
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -542,9 +542,6 @@ def create_job_bundle(
         take.frame_range != first_frame_range for take in submit_takes
     )
 
-    # Deduplicate take names (appends _1, _2, etc. for duplicates)
-    deduplicate_take_names(submit_takes)
-
     # If there are multiple frame ranges and we're not overriding the range,
     # then we create per-take Frames parameters.
     if per_take_frames_parameters:
@@ -589,11 +586,14 @@ def create_job_bundle(
     }
 
 
-def deduplicate_take_names(submit_takes: list[TakeData]) -> None:
+def deduplicate_take_names(submit_takes: list[TakeData]) -> set[str]:
     """
-    Checks for duplicate take names among the submitted takes and makes them
-    unique by appending _1, _2, etc. Adds a warning via warning_collector
-    so the user is informed of the renaming.
+    Checks for duplicate take names and makes them unique by appending
+    _1, _2, etc. suffixes. Handles collisions with existing take names
+    (e.g. 'take', 'take', 'take_1' won't produce two 'take_1' entries).
+
+    Returns the set of original names that were duplicated, or an empty set
+    if no duplicates were found.
     """
     name_counts: dict[str, int] = {}
     for take in submit_takes:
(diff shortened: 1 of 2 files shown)


src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -834,6 +851,10 @@ def on_create_job_bundle_callback(
         """
         Callback function for creating a job bundle when submitting the job.
         """
+        # Deduplicate take names early so the warning shows in the dialog
+        submit_takes = get_submit_takes(settings, takes)
+        generate_take_parameter_names(submit_takes, set_frames_parameters=False)
```
karthikbekalp: Hmm, I wonder if we should just simplify this as `generate_take_parameter_names` is doing a bunch of stuff? 

1. We check if the takes have duplicate names (maybe using a separate function), if yes we add a warning. 
2. We use generate_take_parameter_names in line 546 ?

karthikbekalp (changes requested):
I think the changes look really good! Thanks for the quick fix. 
I added a few comments that can help with the readability of the code so if you can get that fixed then we can merge the changes. :)

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: yep, modified.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -542,6 +542,9 @@ def create_job_bundle(
         take.frame_range != first_frame_range for take in submit_takes
     )
 
+    # Deduplicate take names (appends _1, _2, etc. for duplicates)
+    deduplicate_take_names(submit_takes)
```
joel-wong-aws, line 547: `on_create_job_bundle_callback` calls `create_job_bundle`. Is there a reason we need to de-duplicate take names here (within `create_job_bundle`) in addition to deduplicating them in `on_create_job_bundle_callback`?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -601,25 +634,13 @@ def generate_take_parameter_names(submit_takes: list[TakeData]) -> None:
...
     for take_number in range(len(submit_takes)):
         take_data = submit_takes[take_number]
-
-        # First, check for duplicate take names since this will result in overwriting files in the output
-        # or other unexpected behaviour
-        # We do this here rather than earlier in submission because we get an error popup
-        # (rather than a quieter console error) for errors here.
         take_name = take_data.name
-        if take_name in take_names:
-            raise RuntimeError(
```
joel-wong-aws, line 616: Should we keep this in case the customer creates take names that collide with the de-duplication logic? e.g.
`my_take_name`
`my_take_name`
`my_take_name_1`

Would still collide

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -586,6 +589,36 @@ def create_job_bundle(
...
+    for take in submit_takes:
+        if take.name in next_suffix:
+            suffix = next_suffix[take.name]
+            new_name = f"{take.name}_{suffix}"
+            take.name = new_name
+            take.display_name = new_name[:64]
+            next_suffix[new_name.rsplit("_", 1)[0]] = suffix + 1
+
+    renamed_list = ", ".join(f"'{name}'" for name in sorted(duplicated_names))
+    warning_collector.add_warning(
```
joel-wong-aws: Is there a way for users to auto-ignore warnings? I wonder if customers would get frustrated after seeing this warning over and over again when it's not actually causing any issues?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Interesting.. did not think of this case.. I'll modify the code to address this case by checking if any modified name already exists..

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: No, not really, there's no suppression mechanism at the moment, but the warning only appears when the user has multiple takes with at least one take sharing a name with another. If all take names are unique, no warning shows up.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Good catch.. no reason to deduplicate in both places. The call in create_job_bundle was left over from when I was testing both placements. Removed it..

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -834,6 +874,10 @@ def on_create_job_bundle_callback(
         """
         Callback function for creating a job bundle when submitting the job.
         """
+        # Deduplicate take names early so the warning shows in the dialog
```
karthikbekalp: Hmm, the comment is out of date?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -586,6 +586,58 @@ def create_job_bundle(
...
+
+    return duplicated_names
+
+
+def warn_duplicate_take_names(submit_takes: list[TakeData]) -> None:
+    """
+    Deduplicates take names and adds a warning via warning_collector
+    if any duplicates were found.
+    """
+    duplicated_names = deduplicate_take_names(submit_takes)
```
karthikbekalp: Hmm, let's try to make sure that functions do specific things instead of trying to do everything. 
The idea behind making `warn_duplicate_take_names` a separate function was to ensure that we can keep the responsibility separate. 
If `deduplicate_take_names` changes then it would affect this method as well. 
Can we just make sure this function only emits a warning by just checking if there are duplicate names in the submit_takes list without using `deduplicate_take_names`?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -586,6 +586,58 @@ def create_job_bundle(
...
 
 
+def deduplicate_take_names(submit_takes: list[TakeData]) -> set[str]:
+    """
+    Checks for duplicate take names and makes them unique by appending
+    _1, _2, etc. suffixes. Handles collisions with existing take names
+    (e.g. 'take', 'take', 'take_1' won't produce two 'take_1' entries).
+
+    Returns the set of original names that were duplicated, or an empty set
+    if no duplicates were found.
```
karthikbekalp: It would be odd that `deduplicate_take_names` returns a set of original names that were duplicated.

joel-wong-aws (dismissed):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -586,6 +586,58 @@ def create_job_bundle(
...
+            suffix = next_suffix[original_name]
+            new_name = f"{original_name}_{suffix}"
+            # Skip suffixes that collide with existing take names
+            while new_name in all_names:
+                suffix += 1
+                new_name = f"{original_name}_{suffix}"
+            take.name = new_name
+            take.display_name = new_name[:64]
+            all_names.add(new_name)
+            next_suffix[original_name] = suffix + 1
```
karthikbekalp: Hmm, let's avoid making functions that can have side affects later on as it becomes really hard to read the code. 
For example, here we run warning check which also have as a side effect that changes the take names and then we also have generate_take_parameter_names separately. 
Let's consolidate the code. 

1. Before calling the create job bundle, i.e. in line 878 (in `on_create_job_bundle_callback`) we invoke `warn_duplicate_take_names` function which should only add warnings if duplicate take names exist. Something like `check_take_token_warnings`? 
2. In `create_job_bundle` function, we make per take transformations similar to the flow we use for `generate_take_parameter_names`. 

Let me know if you want to discuss offline about this. :)

karthikbekalp (changes requested):
Hmm, let's focus on readability of the code. Right now, the functions can have side effects that can make it difficult to understand the flow. 

I posted some comments about it here, but let me know if you want to discuss about this offline. :)

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: The return value serves warn_duplicate_take_names.. once dedup finishes, the takes are already renamed, so there's no way to know what the original duplicated names were without re-scanning. 
Returning them lets the warning message say "Duplicate take names were found: 'A', 'B'" accurately. An empty set also acts as a quick "nothing changed" check to skip the warning entirely.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
vaish-gujjari: Modified the comment to indicate what the function does..

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -834,6 +901,10 @@ def on_create_job_bundle_callback(
         """
         Callback function for creating a job bundle when submitting the job.
         """
+        # Check for warnings before submission
```
karthikbekalp: nit: I'm not sure if this comment is useful as the function name clearly states `warn_duplicate_take_names` or maybe we can make the function name better to `add_warnings_if_duplicate_takes`?

karthikbekalp (commented):
LGTM, as discussed offline there is a Sonar Cloud issue that would be best if fixed.

karthikbekalp (approved):
Thanks for fixing the comments!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for updating!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for updating!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for updating!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for updating!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for updating!

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: Thanks for updating!

joel-wong-aws (approved):

Issue #400, opened 2026-03-04
Feature request: Scenes with more than 10,000 frames in a take


karthikbekalp:
### Describe the problem

Currently, if a scene has more than 10,000 frames in a take then job submissions will fail because Deadline Cloud does not allow for a job with more than 10,000 frames. 

### Proposed Solution

This would require more thought and design because its not supported by Deadline Cloud right now. 

A workaround could be that the submitter automatically submits multiple jobs divided with max 10,000 tasks per job. 

### Example Use Cases

Any scene with more than 10,000 frames. 

### Current Workaround

To unblock, customers would have to manually change the frame range and re-submit the same job multiple times to render all the frames. For example, say a scene file has 25,000 frames. 

Then, customers would have to do the following:
- First job, same scene with Override frame range: 0-10000 
- Second job, same scene with Override frame range: 10001-20000. 
- Third job, same scene with Override frame range: 20001-25000

The second and third jobs can be submitted within the monitor as well 

### Environment (optional)

This affects all environments.

PR #401, opened 2026-03-09
feat!: Add tile rendering within submitter.

### LINKED ISSUE #210: Feature request: Tile rendering in Cinema 4D

karthikbekalp:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/210

We needed a way for customers to submit tile rendering jobs so that customers can increase their speed of rendering by dividing the scene across multiple machines. 

Unfortunately, the current tile rendering solution was too manual and required another library to stitch the files after the render.

Investigated the issue further and after discussions with Maxon, implemented a solution which is 1-click for customers. 
This also includes tile assembly by using Cinema 4D and hence does not even require extra dependencies. 

**To make the PR manageable for review (this PR is already >800 line changes), I've divided the PR into 2:**

- This PR: Adds the business logic 
- Upcoming PR : Adds the integration tests modified by adding tile rendering + docs update

Much better UX for tile rendering. 
image

I tested it manually by testing various scenes in Cinema 4D. 

image

Yes

Yes. 

I've also created 2 integration tests for tile rendering which I'll add in my upcoming PR. 

Yes. I've also updated integ_test_helpers file.

The documentation change will be added in upcoming PR. 

Yes. This PR adds some changes that needs a new run_data schema. 

----

### DIFF
--- a/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
@@ -119,7 +119,7 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
-        return SemanticVersion(major=0, minor=3)
+        return SemanticVersion(major=0, minor=4)
 
     @staticmethod
     def _get_timer(timeout: int | float) -> Callable[[], bool]:
@@ -514,7 +514,14 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
             if name in run_data:
                 self._action_queue.enqueue_action(Action(name, {name: run_data[name]}))
 
-        self._action_queue.enqueue_action(Action("start_render", {"frame": run_data["frame"]}))
+        # Dispatch based on tile_action: "assemble", "render" (tile), or absent (normal)
+        tile_action = run_data.get("tile_action")
+        if tile_action == "assemble":
+            assemble_data = self._build_assemble_tiles_data(run_data)
+            self._action_queue.enqueue_action(Action("assemble_tiles", assemble_data))
+        else:
+            start_render_data = self._build_start_render_data(run_data)
+            self._action_queue.enqueue_action(Action("start_render", start_render_data))
 
         while self._cinema4d_is_rendering and not self._has_exception:
             time.sleep(0.1)  # busy wait so that on_cleanup is not called
@@ -529,6 +536,35 @@ class Cinema4DAdaptor(Adaptor[AdaptorConfiguration]):
                 "Cinema4D exited early and did not render successfully, please check render logs. "
                 f"Exit code {exit_code}"
             )
+
+    _TILE_KEYS = (
+        "current_tile_column",
+        "current_tile_row",
+        "total_tiles_column",
+        "total_tiles_row",
+    )
+
+    def _build_start_render_data(self, run_data: dict) -> dict:
+        """Build the data dict sent to the start_render action."""
+        data: dict = {"frame": run_data["frame"]}
+        if run_data.get("tile_action") == "render":
+            data["tile_action"] = "render"
+        for key in self._TILE_KEYS:
+            if key in run_data:
+                # OpenJD step parameters arrive as strings; cast to int for the renderer.
+                data[key] = int(run_data[key])
+        return data
+
+    def _build_assemble_tiles_data(self, run_data: dict) -> dict:
+        """Build the data dict sent to the assemble_tiles action."""
+        return {
+            "frame": run_data["frame"],
+            # OpenJD step parameters arrive as strings; cast to int for the renderer.
+            "total_tiles_column": int(run_data["total_tiles_column"]),
+            "total_tiles_row": int(run_data["total_tiles_row"]),
+            "output_path": run_data.get("output_path", ""),
+            "multi_pass_path": run_data.get("multi_pass_path", ""),
+        }
 
     def on_stop(self) -> None:
         """ """
--- a/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
+++ b/src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
@@ -5,6 +5,13 @@ import os
 import traceback
 from typing import Any, Callable, Dict
 
+# The Cinema4D Adaptor adds the `deadline` namespace directory to PYTHONPATH,
+# so that importing just the cinema4d_adaptor should work.
+try:
+    from cinema4d_adaptor.Cinema4DClient import tile_rendering  # type: ignore[import]
+except (ImportError, ModuleNotFoundError):
+    from deadline.cinema4d_adaptor.Cinema4DClient import tile_rendering  # type: ignore[import]
+
 try:
     import c4d  # type: ignore
     import maxon
@@ -31,6 +38,7 @@ OUTPUT_PATH_KEY = "output_path"
 MULTIPASS_PATH_KEY = "multi_pass_path"
 SCENE_FILE_KEY = "scene_file"
 START_RENDER_KEY = "start_render"
+ASSEMBLE_TILES_KEY = "assemble_tiles"
 TAKE_KEY = "take"
 
 
@@ -78,6 +86,7 @@ class Cinema4DHandler:
             TAKE_KEY: self.set_take,
             FRAME_KEY: self.set_frame,
             START_RENDER_KEY: self.start_render,
+            ASSEMBLE_TILES_KEY: self.assemble_tiles,
             OUTPUT_PATH_KEY: self.output_path,
             MULTIPASS_PATH_KEY: self.multi_pass_path,
             USE_CACHED_TEXT_KEY: self.use_cached_text,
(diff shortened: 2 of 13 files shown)


src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -5,6 +5,13 @@
 import traceback
 from typing import Any, Callable, Dict
 
+# The Cinema4D Adaptor adds the `deadline` namespace directory to PYTHONPATH,
+# so that importing just the cinema4d_adaptor should work.
+try:
+    from cinema4d_adaptor.Cinema4DClient import tile_rendering  # type: ignore[import]
+except (ImportError, ModuleNotFoundError):
```
karthikbekalp, line 12: This is required if the adaptor code is in a different location. See `cinema4d_handler.py` which has similar code.

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
```diff
@@ -0,0 +1,434 @@
...
+        print(f"Saved tile ({ctx.tile_col}, {ctx.tile_row}) to {tile_path}")
+
+    # Restore output paths so subsequent tile renders can use them
+    if ctx.tile_output_path:
+        render_data[c4d.RDATA_PATH] = ctx.tile_output_path
+    if ctx.tile_multipass_path:
+        render_data[c4d.RDATA_MULTIPASS_FILENAME] = ctx.tile_multipass_path
+
+
+def _assemble_tiles_pass(
```
karthikbekalp, line 355: We can refactor this but because this feature is urgent for a customer, I'd pushback. Also, we have integration tests that tests this code so it should be sufficient.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -404,6 +433,21 @@ def initialize_render_settings() -> RenderSubmitterUISettings:
...
+def _resolve_take_paths(
+    settings: RenderSubmitterUISettings,
+    take_name: Optional[str],
+    has_take_token: bool,
+) -> tuple[str, str]:
+    """Resolve output_path and multi_pass_path, substituting $take if present."""
+    if has_take_token and take_name is not None:
+        take_name_for_path = _STRIPPED_PATH_CHARS.sub("_", take_name)
+        return (
+            settings.output_path.replace("$take", take_name_for_path),
```
karthikbekalp: Hmm, I'd pushback on this to get the changes in for customers but I can fix this in my upcoming PR.

vaish-gujjari (dismissed):

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
```diff
@@ -530,6 +537,35 @@ def on_run(self, run_data: dict) -> None:
...
+        for key in self._TILE_KEYS:
+            if key in run_data:
+                # OpenJD step parameters arrive as strings; cast to int for the renderer.
+                data[key] = int(run_data[key])
+        return data
+
+    def _build_assemble_tiles_data(self, run_data: dict) -> dict:
+        """Build the data dict sent to the assemble_tiles action."""
+        return {
+            "frame": run_data["frame"],
```
diane-hj, line 561: Might be a noob question. Now that we have single digit frame, how should tile assemble work when we introduce chunking in it(the expected behavior)?
With chunking, frame could be a range like "1-5" — would the assembly step need to loop over each frame in the range and assemble tiles per frame?

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
karthikbekalp: I just went ahead and updated this.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
karthikbekalp: I just went ahead and updated this.

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
```diff
@@ -0,0 +1,434 @@
...
+    c4d.FILTER_PSD: (".psd", c4d.FILTER_PSD),
+    c4d.FILTER_TGA: (".tga", c4d.FILTER_TGA),
+}
+
+EXT_TO_FILTER = {v[0]: v[1] for v in FORMAT_MAP.values()}
+
+
+def get_format_info(format_id: int) -> tuple[str, int]:
+    """Return (extension, save_filter) for a C4D render format ID."""
+    return FORMAT_MAP.get(format_id, (".png", c4d.FILTER_PNG))
```
diane-hj: Power nit: The default (".png", c4d.FILTER_PNG) is duplicated from FORMAT_MAP. Could we re-use a constant to keep it in sync:
DEFAULT_FORMAT = FORMAT_MAP[c4d.FILTER_PNG]
return FORMAT_MAP.get(format_id, DEFAULT_FORMAT)

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_tile_rendering.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_large_negative_count_raises_value_error(self):
+        with pytest.raises(ValueError, match="count must be >= 1, got -10"):
+            _tile_extent(0, -10, 500)
+
+
+class TestTileExtentValidInputs:
+
+    def test_first_tile_normal_case(self):
+        # 1920 // 7 = 274
```
karthikbekalp, line 27: This is required so that its easier to understand the tests following this.

vaish-gujjari (dismissed):

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -185,6 +186,34 @@ def _build_ui(self, settings):
...
+
+        self.tile_rendering_chck = QCheckBox("Enable Tile Rendering", self)
+        tile_layout.addWidget(self.tile_rendering_chck, 0, 0, 1, 2)
+
+        tile_layout.addWidget(QLabel("Columns"), 1, 0)
+        self.tiles_columns_spin = QSpinBox(self)
+        self.tiles_columns_spin.setMinimum(1)
+        # We added limits for now to keep it under 10k tasks for a single take.
+        # As 100 * 100 = 10k + 1 task for reassembly goes over 10k limit.
+        self.tiles_columns_spin.setMaximum(99)
```
diane-hj, line 201: The 99 cap protects single-frame jobs from exceeding the 10k task limit. Is tile rendering only intended for single frames, or does it support multi-frame? If multi-frame, the total task count (frames × columns × rows) could exceed 10k even with the 99 cap. I might be misunderstanding or missing some context here.

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
karthikbekalp: Hmm, that's a good question. 
Although, I'm not sure about the answer about it. Generally tile rendering is used for single frame scenes although I've implemented in such a way that multiple frames also works fine. 
But with chunking, we might have to refactor this further.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Generally, customers use tile rendering for big single frame scenes although the way I've implemented, it should also work for multiple frames. 
Considering tile rendering has a significant impact on the rendering wall clock time, I think we should allow customers to use as much tile rendering as possible. 

Additionally, customers right now have scenes where they have >10,000 tasks. So they can exceed the 10k cap even without tile rendering :( 
https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/400

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
karthikbekalp: Updated!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
diane-hj: Not a blocker and yea, make sense, but I was concerned that this maximum doesn't protect the multiple-frame jobs. Since this PR supports multi-frame + tiles, could we add a validation warning when frames × columns × rows exceeds 10k?

src/deadline/cinema4d_adaptor/Cinema4DAdaptor/adaptor.py
diane-hj: Makes sense — single frame is the primary use case for tiles but we support multi-frame works too. Agreed that chunking + tiles would need further work, but that's fine :) No concerns with the current implementation

diane-hj (approved):

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Discussed offline, we need to fix this with warnings but out of scope for this PR.

vaish-gujjari (approved):

PR #408, opened 2026-03-18
fix: fix yaml delimeter in paths

### LINKED ISSUE #397: Bug: Path names with single quotes with takes do not work
### Describe the bug

From what I checked, path names with single quotes do not work if the scene also has takes. 

This is because of the issue here where the [serialization/de-serialization fails as we use single quotes to denote the end of the string. ](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/blob/mainline/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py#L249-L261)

Here's the PR fix: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/395

### Expected Behaviour

The scenes with single quotes in the path should work without any issues. 

### Current Behaviour

The Cinema 4D job fails while rendering. 

### Reproduction Steps

Follow the instructions here: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/395

### Environment

In all environments. 

### Logs

_No response_

diane-hj:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/397

Cinema 4D jobs fail on the worker when scene files are in paths containing apostrophes (e.g. /Users/artist's work/scene.c4d). This function wraps {{Param.*}} values in single quotes (because {{ is a special YAML character), producing values like '{{Param.Cinema4DFile}}'. After the Deadline service substitutes the parameter with the actual path, the apostrophe in the path breaks the single-quoted YAML string, causing a parse error on the worker.

Created a new `_yaml_utils.py` module with a `_build_embedded_yaml` helper that generates YAML for embedded files undergoing parameter substitution. The function uses a hybrid approach: non-path values are serialized with deadline_yaml_dump to preserve proper YAML handling (e.g. quoting "true" as a string, escaping special characters), while path values (specified via unquoted_keys) are written as unquoted plain YAML scalars that safely contain both apostrophes and double quotes after parameter substitution. The module is separate from cinema4d_render_submitter.py to avoid circular imports with tile_utils.py. Updated all three call sites: `init_data` and `run_data` in cinema4d_render_submitter.py, and the tile assembly step in `tile_utils.py`.

Users can now submit Cinema 4D jobs from paths containing apostrophes and double quotes without rendering failures.

Manual testing: Submitted a Cinema 4D job with an apostrophe/double quote in the scene path, verified it renders successfully on the farm.

- haejung's folder: 

Screenshot 2026-03-17 at 4 36 08 PM

- haejung"s folder:

Screenshot 2026-03-17 at 4 36 29 PM

```
Yes

===================================== 306 passed, 6 skipped in 4.79s =====================================
```
Yes. 8/9 tests pass. The `redshift_textured_with_nonascii_characters` test fails with this error: 
```
-    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/test_scenes/
-      redshift_textured_with_nonascii_characters/generated_bundle/redshift_textured-_\u20BF/
-      _\u0119_\xF1_\u03B2_\u0411_\u062A.c4d"
-    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/test_scenes/
-      redshift_textured_with_nonascii_characters/generated_bundle/tex/checkerboard-_\u20BF/
-      _\u0119_\xF1_\u03B2_\u0411_\u062A.bmp"
+    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/
+      integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/
+      redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A.c4d"
+    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/
+      integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/
+      tex/checkerboard-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A.bmp"
   outputs:
     directories:
     - D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/renders
---
+++
@@ -1,14 +1,14 @@
 parameterValues:
 - name: Cinema4DFile
-  value: "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/
-    test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/redshift_textured-_\u20BF/
-    _\u0119_\xF1_\u03B2_\u0411_\u062A.c4d"
+  value: "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/
+    integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/
+    redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A.c4d"
 - name: SubmitterIntegrationVersion
   value: 0.8.0
 - name: OutputPath
-  value: "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/
-    test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/renders/
-    redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A"
+  value: "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/
+    integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/
+    renders/redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A"
```
The content is the same but line breaks happen in different location of the path. It is failing on mainline as well, need to dig further but think it is separate issue. 

```
FAILED test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters] - AssertionError
====================================== 1 failed, 8 passed in 1708.69s (0:28:28) =======================================
```

I made changes in the submitter file but the UI is not changed. 

No changes needed to README.md, DEVELOPMENT.md, or adaptor schemas but docstring was added to the helper function.

No. The parsed values after parameter substitution are identical for all realistic file paths. The YAML format changes (path values are now unquoted and key ordering differs), but this does not affect the adaptor which consumes the data as a dict.

----


test/unit/deadline_submitter_for_cinema4d/test_cinema4d_render_submitter.py
```diff
@@ -451,3 +454,67 @@ def test_duplicate_names_warning_is_added(self):
...
+        result = _build_embedded_yaml(
+            {
+                "scene_file": "{{Param.Cinema4DFile}}",
+                "take": "Main",
+                "output_path": "{{Param.OutputPath}}",
+            },
+            unquoted_keys={"scene_file", "output_path"},
+        )
+
+        substituted = result.replace("{{Param.Cinema4DFile}}", '/path/with"quotes/scene.c4d')
```
karthikbekalp, line 497: Thanks for the fix. Just one question. What if the path has both single quote and double quote? 
Can we just give customers a warning that don't use both the quotes at the same time? Something like this: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/396?

karthikbekalp (changes requested):
Just one suggestion. Let me know what you think?

karthikbekalp:
> Yes. 8/9 tests pass. The redshift_textured_with_nonascii_characters test fails with this error:

Yeah. Its an issue with the line endings. Feel free to update it as necessary or maybe refactor so that we only compare the strings within the path instead of comparing the files? (Out of scope for this PR, only if you have time :)

test/unit/deadline_submitter_for_cinema4d/test_cinema4d_render_submitter.py
```diff
@@ -451,3 +454,67 @@ def test_duplicate_names_warning_is_added(self):
...
+        )
+
+        substituted = result.replace("{{Param.Cinema4DFile}}", "/Users/artist's work/scene.c4d")
+        substituted = substituted.replace("{{Param.OutputPath}}", "/Users/artist's work/render")
+        substituted = substituted.replace("{{Param.ActivateErrorChecking}}", "1")
+        substituted = substituted.replace("{{Param.UseCachedText}}", "0")
+
+        parsed = yaml.safe_load(substituted)
+        assert parsed["scene_file"] == "/Users/artist's work/scene.c4d"
+        assert parsed["output_path"] == "/Users/artist's work/render"
```
karthikbekalp, line 483: Hmm, quick question. 

We get the result from `_build_embedded_yaml` but then we manually substitute the value and then compare later on?
I'd have imagined that `_build_embedded_yaml` does that substitution itself. 

I don't see the same substitution in our code paths so is it something specific that we need to do every time? Or is it specific for tests? 

Let me know if we can discuss about this test offline.

test/unit/deadline_submitter_for_cinema4d/test_cinema4d_render_submitter.py
diane-hj: Our fix handles both. Unquoted YAML plain scalars can contain both apostrophes and double quotes simultaneously. I verified via local testing - submitted the job and rendered successfully: 
```
2026/03/19 09:27:05-07:00 ADAPTOR_OUTPUT: STDOUT: Mapped path '/sessions/session-9832f83383db47bb93d57a6d49487f25zg7jjn5t/assetroot-530e96ad98f70b0f9bf4/client-software/testing/haejung'smeow"check/Untitled' to '/sessions/session-9832f83383db47bb93d57a6d49487f25zg7jjn5t/assetroot-530e96ad98f70b0f9bf4/client-software/testing/haejung'smeow"check/Untitled'.
```
I will add a unit test for this.

test/unit/deadline_submitter_for_cinema4d/test_cinema4d_render_submitter.py
diane-hj: Yeah, the test simulates the full pipeline: 
-  _build_embedded_yaml generates YAML with {{Param.*}} placeholders (this happens at the submission time) 
- str.replace simulates the OpenJD runtime resolving those placeholders with actual values on the worker. 
- yaml.safe_load verifies the result is valid YAML (this is what the adaptor does when it reads init-data.yaml) 
The substitution is not in our codes but I added so that we can simulate and verify the YAML survives with the real paths with special characters. Please let me know if I misunderstand anything!

karthikbekalp: Ooh, that makes sense. Thanks for clarifying. 
I wonder if this should be an integration test for this ? (So that we can verify that the openjd substitution works correctly?)

diane-hj: Thanks! I added an integration for this and also updated other test scenes' expected bundle to match.

test/unit/deadline_submitter_for_cinema4d/test_cinema4d_render_submitter.py
diane-hj: Added 🙏

karthikbekalp (approved):
Thanks so much for the fix! And also for updating the tests!

crowecawcaw (approved):

Issue #409, opened 2026-03-18
Bug: Make tile rendering work for Cinema 4D 2024


karthikbekalp:
### Describe the bug

The current implementation of tile rendering uses library functions like `BakeOcioViewToBitmap` for OCIO that were not available pre-Cinema 4D 2025. 

### Expected Behaviour

Tile rendering works well in 2024 as well. 

### Current Behaviour

We get an error: 

```
AttributeError: module 'c4d.documents' has no attribute 'BakeOcioViewToBitmap'
```

### Reproduction Steps

Submit a tile rendering job from Cinema 4D 2024. 

### Environment

Requires Cinema 4D 2024.

karthikbekalp:
Cinema 4D is 3 years old and Maxon generally does not provide any updates to older versions. But if you feel free that this is something that affects your workflow, feel free to like this Github issue and we can look into it. :)

PR #410, opened 2026-03-18
docs: Add FAQ explaining Tile rendering only available for C4D 2025+ …

karthikbekalp:

Cinema 4D uses some newer Cinema 4D APIs like `BakeOcioViewToBitmap` to fix the colouring in tile rendering. 

Cinema 4D 2024 is 3 years old and hence just updated the docs to suggest customers to upgrade to newer versions of Cinema 4D. 

Also cut a Github issue in case there are many customers who want us to support the workflow: 
https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/409

Customer awareness for the issue. 

Just doc changes. 

----


diane-hj (approved):
Thanks for providing this document!

vaish-gujjari (approved):

PR #413, opened 2026-03-23
feat: bundle PySide6-Essentials dependency

joel-wong-aws:

Cinema 4D has a custom submitter, so Qt dependencies need to be bundled separately from the main deadline-cloud repo.

This is the Cinema 4D counterpart to aws-deadline/deadline-cloud#1021, which bundled Qt dependencies in the deadline-cloud installer.

Add `PySide6-Essentials == 6.8.3` as a direct dependency in `pyproject.toml`, pinned to 6.8 to align with [VFX Reference Platform 2026](https://vfxplatform.com/#reference-platform).

PySide6-Essentials will be installed as a dependency of this package, ensuring Qt is available for the Cinema 4D submitter GUI without requiring a separate install step.

N/A - dependency addition only.

N/A

No

### Does this change impact security?

No

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ classifiers = [
 dependencies = [
     "deadline >= 0.51.1,< 0.55",
     "openjd-adaptor-runtime >= 0.7,< 0.10",
+    "PySide6-Essentials == 6.8.3",  # Use 6.8 to align with VFXP 2026: https://vfxplatform.com/#reference-platform
     # fonttools is Windows-only due to Cinema 4D technical limitations
     "fonttools >=4.59.2, <4.61; sys_platform == 'win32'",
 ]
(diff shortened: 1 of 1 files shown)


crowecawcaw:
Can you give a map of how dependencies end up bundled in? One tricky part about getting Pyside and related dependencies into the Deadline client installer was making sure we only bundled certain Qt modules into it for licensing reasons. We limited the modules with several layers of filtering that removed unused files. Ideally, this integration reuses the filtered Pyside files so we don't have to reimplement it here. You can double check by looking at the files th installer creates in the Pyside directory and seeing if they match the ones in allowlist.py in that linked PR.

karthikbekalp:
Closing this PR in favor of: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/417

PR #415, opened 2026-03-23
feat!: Add contiguous chunking support in Cinema 4d

diane-hj:

Added task chunking support using the OpenJD TASK_CHUNKING extension. 
OpenJD spec: https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md

The Frame parameter in both job templates was changed from INT to CHUNK[INT] with contiguous range constraint. Two new job parameters were added: ChunkSize (default 1, preserving existing behavior) and TargetChunkDuration (default 0, disabled). The Cinema 4D handler's start_render parses the frame range (e.g. "1-10") and sets RDATA_FRAMEFROM/RDATA_FRAMETO so Cinema 4D renders the full chunk natively in a single RenderDocument call. 

The UI adds chunk size and target duration controls, mutually exclusive with tile rendering. The default (non-adaptor) template script was also updated to parse the range into start/end frames for Cinema 4D's command-line renderer.

Added task chunking support using the `OpenJD TASK_CHUNKING` extension. The Frame parameter in both job templates was changed from INT to CHUNK[INT] with rangeConstraint: CONTIGUOUS, since Cinema 4D's renderer accepts start/end frame parameters rather than arbitrary frame lists. Two new job parameters were added: ChunkSize (default 1, preserving existing behavior) and TargetChunkDuration (default 0, disabled). The Cinema 4D handler's start_render parses the contiguous frame range (e.g. "1-10") and sets RDATA_FRAMEFROM/RDATA_FRAMETO so Cinema 4D renders the full chunk natively in a single RenderDocument call. The UI adds chunk size and target duration controls, mutually exclusive with tile rendering. The default (non-adaptor) template script was also updated to parse the range into start/end frames for Cinema 4D's command-line renderer.

- With chunk size 1 (default), behavior is identical to before — no breaking changes
- Higher chunk sizes reduce per-task overhead, with ~20% improvement observed in testing
- Tile rendering and chunking are mutually exclusive (enforced in UI)
- All frame value handling updated to support the "start-end" range format that CHUNK[INT] produces, including set_frame, start_render, and assemble_tiles
- Requires a worker agent that supports the TASK_CHUNKING extension; service-managed fleets always have a compatible version

- Chunking enabled: 

Screenshot 2026-04-01 at 11 54 40 AM

- Chunkinig Size is 1 which is the default behavior. You can enable tile. 

Screenshot 2026-04-01 at 11 54 55 AM

- Tile is enabled, chunking is not editable 

Screenshot 2026-04-01 at 11 55 13 AM

- Chunk with target run seconds: 

Screenshot 2026-04-01 at 5 52 08 PM

- Unit tests: `_parse_frame_range` (single frame, range, single-frame range), start_render with chunk range, set_frame with range string, adaptor on_run with chunk range
- Integration test: new physical_chunking scene rendering frames 1-4 with chunk_size=2, producing 2 chunks — verified job runs successfully and output images match expected
- All existing integration tests updated with new expected job bundles and verified passing
- Manual testing with various scenes submitted to Deadline Cloud via the Cinema 4D submitter

Yes
```
=========================================== 311 passed, 6 skipped in 8.61s ============================================
```

Yes 
```
  inputs:
     directories: []
     filenames:
-    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/test_scenes/
-      redshift_textured_with_nonascii_characters/generated_bundle/redshift_textured-_\u20BF/
-      _\u0119_\xF1_\u03B2_\u0411_\u062A.c4d"
-    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/integ/test_scenes/
-      redshift_textured_with_nonascii_characters/generated_bundle/tex/checkerboard-_\u20BF/
-      _\u0119_\xF1_\u03B2_\u0411_\u062A.bmp"
+    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/
+      integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/
+      redshift_textured-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A.c4d"
+    - "D:/Users/chaejung/client-software/deadline-cloud-for-cinema-4d/test/
+      integ/test_scenes/redshift_textured_with_nonascii_characters/generated_bundle/
+      tex/checkerboard-_\u20BF_\u0119_\xF1_\u03B2_\u0411_\u062A.bmp"
  

=============================================== short test summary info ===============================================
FAILED test/integ/test_cinema4d.py::test_integ[redshift_textured_with_nonascii_characters] - AssertionError
====================================== 1 failed, 10 passed in 1360.70s (0:22:40) ======================================
```
- Yes 

- Yes, updated docstrings for on_run, set_frame, and start_render to document chunk range support
- No schema changes needed — run_data.schema.json already defines frame as string type, which accepts both "5" and "1-10"
- README and DEVELOPMENT.md do not need updates — chunking is exposed through the existing submitter UI and job template parameters

No. With the default chunk size of 1, the behavior is identical to before. The frame field in run-data was already typed as string in the schema, and the adaptor now accepts both single frames ("5") and ranges ("1-10"). Jobs submitted with older submitter versions (using INT Frame parameter) will continue to work. The new ChunkSize and TargetChunkDuration job parameters have defaults that preserve existing behavior. Requires a worker agent that supports the TASK_CHUNKING extension, but service-managed fleets always have a compatible version.

----


src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -255,6 +255,16 @@ def _pathmap_base_material(self, owner, node_space, node_path, mapped_path) -> b
             transaction.Commit()
         return True
 
+    @staticmethod
+    def _parse_frame_range(frame_value) -> tuple[int, int]:
```
karthikbekalp, line 259: Hmm, considering this requires adaptor change, this needs to be a backwards incompatible change.

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
```diff
@@ -434,7 +434,8 @@ def assemble_tiles(
     """
     tiles_columns = int(data["total_tiles_column"])
     tiles_rows = int(data["total_tiles_row"])
-    frame = int(data["frame"])
+    frame = int(data["frame"].split("-")[0]) if "-" in str(data["frame"]) else int(data["frame"])
```
karthikbekalp, line 440: Can customers use task chunking and tile rendering at the same time? 

I think it might not help for tile rendering because customers specifically wanted to parallel render to get faster rendering.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -186,6 +186,39 @@ def _build_ui(self, settings):
         lyt.addWidget(rendering_options_box, widget_row, 0, 1, 2)
         widget_row += 1
 
+        # Chunking group box
```
karthikbekalp, line 173: Can we reorder the settings a bit? 

Move tile rendering and task chunking above cache text (Customers may not need cache text if they are already using 2026+)?  

Out of scope for this PR though but a quick thing.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -186,6 +186,39 @@ def _build_ui(self, settings):
...
+            "Number of frames to group into each chunk.\n"
+            "Use 1 for one frame per task (default).\n"
+            "Higher values reduce per-task overhead."
+        )
+        chunking_layout.addWidget(self.chunk_size_spin, 0, 1)
+
+        chunking_layout.addWidget(QLabel("Target chunk duration (seconds)"), 1, 0)
+        self.target_chunk_duration_spin = QSpinBox(self)
+        self.target_chunk_duration_spin.setMinimum(0)
+        self.target_chunk_duration_spin.setMaximum(86400)
```
karthikbekalp: Can you add comments on why this range is good for task chunking?

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -186,6 +186,39 @@ def _build_ui(self, settings):
...
         widget_row += 1
 
+        # Chunking group box
+        chunking_group_box = QGroupBox("Task Chunking", self)
+        chunking_layout = QGridLayout(chunking_group_box)
+
+        chunking_layout.addWidget(QLabel("Chunk size"), 0, 0)
+        self.chunk_size_spin = QSpinBox(self)
+        self.chunk_size_spin.setMinimum(1)
+        self.chunk_size_spin.setMaximum(150)
```
karthikbekalp, line 183: Can you add comments on why this range is good for task chunking?

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -249,6 +285,16 @@ def _configure_settings(self, settings):
...
         self.tiles_rows_spin.setEnabled(settings.enable_tile_rendering)
 
+        self.chunk_size_spin.setValue(settings.chunk_size)
+        self.target_chunk_duration_spin.setValue(settings.target_chunk_duration)
+
+        # Apply mutual exclusion between tile rendering and chunking
+        if settings.enable_tile_rendering:
+            self.chunk_size_spin.setEnabled(False)
+            self.target_chunk_duration_spin.setEnabled(False)
+        elif settings.chunk_size > 1:
```
karthikbekalp, line 318: Do we show a dialog or add a warning when both are activated? Maybe we can inform customers not to do that?

test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py
```diff
@@ -392,6 +392,32 @@ def test_on_run(
...
+        mock_server.return_value.server_path = "/tmp/9999"
+        is_rendering_mock = PropertyMock(side_effect=[None, True, False])
+        Cinema4DAdaptor._is_rendering = is_rendering_mock
+        adaptor.on_start()
+
+        # WHEN
+        adaptor.on_run({"frame": "1-10"})
+
+        # THEN
+        mock_sleep.assert_called_once_with(0.1)
```
karthikbekalp, line 422: Hmm, curious why are we mocking if sleep was called here?

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_cinema4d_handler.py
```diff
@@ -756,6 +756,67 @@ def test_start_render_updates_text_was_cached_flag(
...
+
+        mock_render_document.return_value = c4d.RENDERRESULT_OK
+
+        with patch.object(handler, "_cache_text_if_needed", return_value=False):
+            handler.start_render({"frame": "10-20"})
+
+        # FRAMEFROM should be called with start frame, FRAMETO with end frame
+        calls = mock_base_time.call_args_list
+        assert (10, 24) in [c.args for c in calls]
+        assert (20, 24) in [c.args for c in calls]
```
karthikbekalp: nit: Can we make this keyword arguments `start_frame=10, fps=24` or something so that its easier to understand why we check for 10,24 ?

karthikbekalp (changes requested):
Added few minor but blocking comments, but the PR looks so far.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -186,6 +186,39 @@ def _build_ui(self, settings):
...
+        self.target_chunk_duration_spin.setMinimum(0)
+        self.target_chunk_duration_spin.setMaximum(86400)
+        self.target_chunk_duration_spin.setValue(0)
+        self.target_chunk_duration_spin.setToolTip(
+            "When set, the scheduler dynamically adjusts chunk sizes\n"
+            "based on observed runtimes of completed chunks, aiming\n"
+            "for this duration per chunk. Leave at 0 to use a fixed\n"
+            "chunk size for all chunks."
+        )
+        self.target_chunk_duration_spin.setSpecialValueText("Disabled")
```
karthikbekalp: I think we should use "Deactivated" instead of "Disabled" if possible.

karthikbekalp:
Can we also add few docs for this feature?

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
diane-hj: > Can customers use task chunking and tile rendering at the same time?
> 
> I think it might not help for tile rendering because customers specifically wanted to parallel render to get faster rendering.

Nice catch! Task chunking and tile rendering are mutually exclusive — the submitter UI enforces this by disabling one when the other is active. However, since the template always uses CHUNK[INT], the frame value arrives in range format (e.g. "1-1") even with chunk size 1. Added a comment explaining this.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
diane-hj: Thanks. Added comments!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
diane-hj: Added :)

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
diane-hj: We handle this through the UI - enabling tile rendering automatically disables and resets the chunking controls, and setting chunk size > 1 unchecks and disables tile rendering. It kinda automatically handled it, customers won't be able to submit a job with both active at the same time so I think it should be good without a dialog, but please let me know if you think we should add an additional warning dialog on top of this!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
diane-hj: Thanks for cathing these. 🙏  I updated it.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: I think a warning dialog would be good but not a blocker for this PR.

phil-IO-p:
UX feedback:

**1. Chunk Size <--> Target Chunk Duration relationship is confusing**
These two fields interact in a non-obvious way: when Target Chunk Duration is non-zero, it effectively overrides Chunk Size at runtime. But the UI presents them as independent fields with no visual indication of the override. A user setting both will have no idea which one "wins." Consider adding brief inline help text explaining the relationship (e.g., "When Target Chunk Duration is set, this value is used only as the initial chunk size").

**2. "Chunk Size" label is ambiguous"**
The unit isn't clear from the label alone. Is it frames? Seconds? Bytes? The description already says "Number of frames per chunk," so perhaps that should *be* the label. This applies to both the UI spin box label and the job template `label:` fields.

**3. Target Chunk Duration tooltip is too technical**
The tooltip references "the scheduler" but will artists know what that refers to. I suggest rephrasing to something like: *"Target render time per chunk. Deadline Cloud will automatically adjust how many frames are grouped together to hit this target. Set to 0 to always use the fixed frames-per-chunk value above."*

**4. Target Chunk Duration spin box has no unit indicator**
The raw number in the spin box is, dare I say, meaningless without reading the tooltip. Consider adding a 'sec' suffix so the unit is always visible inline.

**5. "Deactivated" special value text**
"Deactivated" implies the feature was previously active and turned off. For a default-0 field most users won't touch, consider something more concise that reads more naturally and is shorter.

**6. Buttons require more padding**
Based on the screenshots provided, I recommend adding more padding to the "Settings," "Export bundle," and "Submit" buttons so that they more closely match the "Help" button on the left. This looks more polished.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
phil-IO-p: +1 to the warning dialog suggestion. Even without a dialog though, the grayed-out controls need *some* inline explanation. A customer who doesn't know these are mutually exclusive will just see disabled fields with no indication why.

A static label like *"Not available when tile rendering is enabled"* next to the disabled chunking controls would go a long way. Same in reverse... when chunk size > 1, the disabled tile rendering checkbox could communicate something like *"Not available when chunk size is greater than 1."*

Without this, customers could assume the UI is broken.

diane-hj: Thanks both! I've added the inline labels as suggested — "Not available when tile rendering is enabled" and "Not available when frames per chunk is greater than 1" appear next to the disabled controls. If we still want a warning dialog on top of this, I'll create a follow-up PR/ticket to track it. 🙏

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
karthikbekalp: I updated the PR title.

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
karthikbekalp: Cool thanks for clarifying!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Thanks for making the change.

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Thanks!

karthikbekalp (dismissed):
Thanks Haejung. The changes look good to me!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Thanks!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
karthikbekalp: Thanks!

test/unit/deadline_adaptor_for_cinema4d/Cinema4DAdaptor/test_adaptor.py
karthikbekalp: Thanks for the comment.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
diane-hj: Oops. Thanks for updating it!!

src/deadline/cinema4d_submitter/ui/components/scene_settings_tab.py
```diff
@@ -170,20 +170,49 @@ def _build_ui(self, settings):
...
+        chunking_layout.addWidget(self.chunk_size_spin, 0, 1)
+
+        chunking_layout.addWidget(QLabel("Target chunk duration"), 1, 0)
+        self.target_chunk_duration_spin = QSpinBox(self)
+        # Min 0 (use fixed chunk size). Max 3600 (1 hour),
+        # a practical upper bound since chunking targets short-running tasks.
+        self.target_chunk_duration_spin.setMinimum(0)
+        self.target_chunk_duration_spin.setMaximum(3600)
+        self.target_chunk_duration_spin.setValue(0)
+        self.target_chunk_duration_spin.setSuffix(" sec")
```
joel-wong-aws: nit: can we handle the correct pluralization of this based on the value in the spinner?

phil-IO-p:
Thanks for addressing UX feedback! Synced w/ you directly on a few non-blocking changes worth addressing.

joel-wong-aws (dismissed):

phil-IO-p (approved):

phil-IO-p (approved):

joel-wong-aws (approved):

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_cinema4d_handler.py
karthikbekalp: Thanks for fixing it!

karthikbekalp (approved):

PR #422, opened 2026-03-30
fix: Reduce the adaptor package size by moving GUI deps outside of re…

karthikbekalp:

The GUI components that are included will bloat the size of adaptors. 

Make the GUI components optional. 

Our submitters will still bundle these but we don't need it for adaptors. 

Smaller adaptor package size. 

Yes

No, this does not change the code in anyway. 

No

No, it does not require documentation. 

No. 

----


epmog (dismissed):

crowecawcaw:
LGTM but we also need to update the manual install instructions

epmog (approved):

crowecawcaw (approved):

PR #456, opened 2026-06-05
 docs: add Python 3.11 version requirement note to DEVELOPMENT.md

gianelli-lagos-aws:

Integration tests fail with ImportError: Module use of python313.dll conflicts with this version of Python when pywin32 is installed using a Python version that doesn't match Cinema 4D's bundled Python 3.11. There was no documentation warning developers about this requirement.

Added a note in DEVELOPMENT.md's Integration Tests section (before the pywin32 install step) explaining that Python 3.11 must be used to match Cinema 4D 2026's bundled Python, and how to verify the version.

Documentation only. Helps developers avoid a confusing DLL mismatch error when setting up integration tests.

Nothing was tested since this is a doc change.

Yes.

No.

----

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -250,13 +250,30 @@ You can also run `c4dpy` and `Commandline.exe` separately and set the license in
    1. `set C4D_LOCATION=<Cinema 4D location>` on Windows Command or `$env:C4D_LOCATION = <Cinema 4D location>` on Windows Powershell.
       1. The default location for `Cinema 4D` on Windows is `C:\Program Files\Maxon Cinema 4D 2026\`. This location would be automatically used if the directory exists.
 2. For running the adaptor tests, we would need to install `pywin32` to the installation paths as its an adaptor dependency.
+
+   > **Important: Python version must match Cinema 4D's bundled Python.**
+   >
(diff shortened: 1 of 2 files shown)


diane-hj (dismissed):
Thanks for updating the docs to be accurate! 🎉

DEVELOPMENT.md
```diff
@@ -250,13 +250,30 @@ You can also run `c4dpy` and `Commandline.exe` separately and set the license in
...
+   > ```
+   > ImportError: Module use of python313.dll conflicts with this version of Python.
+   > ```
+   >
+   > To ensure compatibility:
+   > 1. Install Python 3.11 and make it your default (`python --version` should show 3.11.x)
+   > 2. Install hatch using Python 3.11 (`python -m pip install hatch`)
+   > 3. Use Python 3.11's pip for the pywin32 install below
+   >
+   > You can verify Cinema 4D's Python version by checking the `resource/modules/python/libs/` folder inside your Cinema 4D installation directory.
```
karthikbekalp, line 268: Hmm, I think it can be useful to add this info to the skill in this repo as well? i.e. `/skills/c4d-dev-setup`?

karthikbekalp (changes requested):

karthikbekalp (approved):

PR #460, opened 2026-06-08
fix: show popup when unsupported renderer is selected instead of sile…

gianelli-lagos-aws:

Selecting an unsupported renderer (e.g., Viewport Renderer) caused the submitter to silently fail — no popup, no error message. The submitter just didn't open, with only a ValueError printed to the console that customers wouldn't see.

Added categorized renderer warnings using the logger.warning() pattern. Renderers are classified into verified (no warning), third-party plugin (installation reminder), unsupported (not supported warning), viewport (quality warning), and unknown (unverified warning). Added viewport_renderer to the RendererNames enum so unknown renderers no longer crash the submitter. A _get_renderer_display_name() helper uses Cinema 4D's plugin registry to resolve friendly renderer names. Warnings are surfaced through the existing WarningCollectorHandler and shown in the Issues Detected dialog at submission time.

Screenshot 2026-06-08 at 3 38 17 PM

Users now see categorized warnings in the submission issues dialog for non-verified renderers, with the option to continue anyway. The submitter no longer silently crashes on unknown renderer IDs. No behavior change for verified renderers (Standard, Physical, Redshift).

Yes. Added unit tests in test_scene.py covering: verified renderers return no warning, third-party renderers produce a plugin warning, unsupported renderers produce a not-supported warning, Viewport Renderer produces a quality warning, unknown renderer IDs produce an unverified warning, and Scene.renderer() returns a string (without crashing) for both known and unknown renderer IDs.
============ 323 passed, 6 skipped in 6.36s ==========

Yes. Existing integration tests pass.
===================== 10 passed, 1 xfailed in 1002.88s (0:16:42) =========

Yes.  Renderer warnings are now shown in the Issues Detected dialog at submission time using the logger.warning pattern.

- Did you update all relevant docstrings for Python functions that you modified?

Updated docstring for Scene.renderer() to note the UnsupportedRendererError raise.

- Should the README.md, DEVELOPMENT.md, or other documents in the repository's docs/ directory be updated along with your change?

No.

- Should the schema files for the adaptor's init-data or run-data be updated?

No.

No.

### DIFF
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -33,7 +33,7 @@ from .detailed_logging_utils import get_detailed_logging_environment
 from .font_utils import scene_has_fonts, get_font_manager_environment, FONTS_DIR
 from .warning_collector import warning_collector
 from .platform_utils import is_windows
-from .scene import Animation, Scene
+from .scene import Animation, Scene, UnsupportedRendererError
 from .style import C4D_STYLE
 from .takes import TakeSelection
 from .template_timeout_patcher import add_timeouts_to_job_template
@@ -107,6 +107,8 @@ def show_submitter():
                 w = _show_submitter(temp_dir, None)
             w.setStyleSheet(C4D_STYLE)
             w.exec_()
+    except UnsupportedRendererError as e:
+        c4d.gui.MessageDialog(str(e))
     except Exception:
         print("Deadline UI launch failed")
         import traceback
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -33,7 +33,7 @@ from .detailed_logging_utils import get_detailed_logging_environment
 from .font_utils import scene_has_fonts, get_font_manager_environment, FONTS_DIR
 from .warning_collector import warning_collector
 from .platform_utils import is_windows
-from .scene import Animation, Scene, UnsupportedRendererError
+from .scene import Animation, Scene, get_renderer_warning
 from .style import C4D_STYLE
 from .takes import TakeSelection
 from .template_timeout_patcher import add_timeouts_to_job_template
@@ -77,6 +77,9 @@ def show_submitter():
     if _prompt_save_current_document() is False:
         return
 
+    if _check_renderer_warning() is False:
+        return
+
     try:
         app = QtWidgets.QApplication.instance()
         if not app:
@@ -107,8 +110,6 @@ def show_submitter():
                 w = _show_submitter(temp_dir, None)
             w.setStyleSheet(C4D_STYLE)
             w.exec_()
-    except UnsupportedRendererError as e:
-        c4d.gui.MessageDialog(str(e))
     except Exception:
         print("Deadline UI launch failed")
         import traceback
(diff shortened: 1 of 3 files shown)


crowecawcaw (dismissed):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -107,6 +107,8 @@ def show_submitter():
                 w = _show_submitter(temp_dir, None)
             w.setStyleSheet(C4D_STYLE)
             w.exec_()
+    except UnsupportedRendererError as e:
+        c4d.gui.MessageDialog(str(e))
```
karthikbekalp: Hmm, I wonder if this should still be a warning?

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -13,6 +13,31 @@
...
+
+    def __init__(self, render_id: int):
+        self.render_id = render_id
+        # Try to get the renderer's display name from Cinema 4D's plugin registry
+        renderer_name = None
+        try:
+            plugin = c4d.plugins.FindPlugin(render_id, c4d.PLUGINTYPE_VIDEOPOST)
+            if plugin:
+                renderer_name = plugin.GetName()
+        except Exception:
```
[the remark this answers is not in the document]
karthikbekalp: I think this could be important, any reason we silently pass? Are we expecting some exceptions here?

src/deadline/cinema4d_submitter/scene.py
```diff
@@ -13,6 +13,31 @@
...
+            pass
+        if not renderer_name:
+            renderer_name = str(render_id)
+
+        supported = ", ".join(r.name.replace("_", " ").title() for r in RendererNames)
+        super().__init__(
+            f"Unsupported Renderer\n\n"
+            f'The selected renderer "{renderer_name}" is not supported for Deadline Cloud rendering.\n\n'
+            f"Supported renderers: {supported}.\n\n"
```
karthikbekalp: Hmm, Deadline Cloud does not support Octane. There might be other renderers too that we have not verified if Deadline Cloud supports. 

Can we check it before we show this message?

Here are ones that we have checked: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d#3rd-party-plugins

Also, what happens if a customer tries to render a scene with say "ViewportRenderer"? Does it get proper outputs i.e. are images generated? (If images are generated, then I don't think we should stop customers from submitting such job bundles as it works on Deadline Cloud :)

karthikbekalp (changes requested):
Added few comments. Let's sync offline. :)

src/deadline/cinema4d_submitter/scene.py
gianelli-lagos-aws: Reworked the approach. Instead of blocking anything not in the enum, I added categorized warnings: verified renderers (Standard, Physical, Redshift) show no warning, third-party renderers (Arnold, V-Ray) warn about plugin installation, unsupported renderers (Octane, Corona, Cycles) warn they may not produce expected results, and the Viewport Renderer warns about output quality since it does have proper outputs. Unknown renderers get a generic "not verified" warning. Users can always continue.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
gianelli-lagos-aws: Changed it to a warning! It now uses c4d.gui.QuestionDialog so the user can choose to continue or cancel. The submitter no longer hard-blocks. I also tested the Viewport Renderer on the farm and it does produce output (viewport-quality screenshots), so blocking isn't appropriate.

karthikbekalp (approved):
LGTM! Great work @gianelli-lagos-aws !

diane-hj (approved):
Thank you, Gianelli!

crowecawcaw (approved):

Issue #372, opened 2025-12-19
Bug: Use Project Frame Rate anomoly


andymanaws:
### Describe the bug

version 2025.3.3
Looks like ticking the 'Use Project Frame rate' box in the redshift render settings does not transfer the project frame rate to the deadline nodes. 
Instead they will default to the 30fps. 

### Expected Behaviour

Use Project Frame rate should adhere to the frame rate.

### Current Behaviour

Seems to not use the project frame rate - defaults to 30.

If you explicitly define Frame Rate in the Output setting, then it adheres to that.

### Reproduction Steps

Submit a job with a frame rate not at 30, elect to use project frame and compare the result.

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Version of Cinema 4D
3. Version of this package
4. If this is from a version installed by the Deadline Cloud Submitter installer, then what version of the submitter installer?

Please share other details about your environment that you think might be relevant to reproducing the bug.

karthikbekalp:
Thanks for raising this issue. 
I believe this could be because we don't send or set this parameter at the adaptor side. 

To fix the issue we would have to do the following: 

1. Check our assumption that regardless of the frame rate defined in the render setting, the adaptor defaults to 30 FPS. Maybe we can use the `doc.GetFPS()` to check that in the adaptor? 
    1. Are all the render settings (not just FPS) somehow also getting set to a default? If so we would have to fix that. 
2. Once we have confirmed above, we should send this value from the submitter in init data similar to other settings and set this value at the adaptor. The setting seems to be named `RDATA_FRAMERATE` according to https://developers.maxon.net/docs/py/2024_2_0/modules/c4d.documents/RenderData/index.html

karthikbekalp:
As this issue is fixed by the above PR, I'm closing this ticket.

PR #462, opened 2026-06-09
fix: use RDATA_FRAMERATE instead of doc.GetFps() for frame time mapping

gianelli-lagos-aws:
Fixes: *<https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/372>*

When "Use Project Frame Rate" is unchecked in render settings and a custom FPS is set, both the submitter and adaptor use doc.GetFps() (document FPS) instead of render_data[c4d.RDATA_FRAMERATE] (render settings FPS). In the submitter, this causes wrong frame numbers in the job bundle (e.g., Frames: 0-30 instead of 0-12). In the handler, this causes frames to be mapped to incorrect times when constructing BaseTime objects. Together, these produce renders at wrong frame positions with the wrong frame count.

Changed both cinema4d_handler.py and the submitter's scene.py (Animation.start_frame(), current_frame(), end_frame()) to use render_data[c4d.RDATA_FRAMERATE] instead of doc.GetFps(). RDATA_FRAMERATE holds the effective render FPS regardless of the "Use Project Frame Rate" checkbox state. Both fixes are needed together — fixing only one side makes the mismatch worse.

Renders now use the correct FPS from render settings when mapping frame numbers to time values and generating frame ranges. No behavior change when project FPS and render FPS are the same (majority of cases).

Yes. Updated existing test_start_render_with_chunk_range to use RDATA_FRAMERATE mock (24fps) and verify the code uses the render data FPS. Updated Animation test mocks to include RDATA_FRAMERATE. All 330 passed, 6 skipped.
======= 330 passed, 6 skipped in 5.58s =====

Yes. Added a new physical_custom_fps integration test with a scene where project FPS = 30 and render FPS = 12 (unchecked "Use Project Frame Rate"), with an animated cube. The test produces 13 correct frames (0-12) with the fix, and 31 wrong frames (0-30) without it. 11 passed, 1 xfailed.
==== 11 passed, 1 xfailed in 1140.42s (0:19:00) ====

Yes. Updated Animation.start_frame(), current_frame(), and end_frame() in 
scene.py
 to use render_data[c4d.RDATA_FRAMERATE] instead of doc.GetFps().

No documentation changes needed. The fix is internal to the submitter and adaptor handler and does not change any public interface or schema.

No. The fix corrects incorrect behavior. Renders now produce the expected output when custom FPS is used. No changes to init-data or run-data schemas.


crowecawcaw (dismissed):

test/integ/test_scenes/physical_custom_fps/expected_job_bundle/template.yaml
```diff
@@ -0,0 +1,653 @@
...
+        - daemon
+        - run
+        - --connection-file
+        - '{{ Session.WorkingDirectory }}/connection.json'
+        - --run-data
+        - file://{{ Task.File.runData }}
+        cancelation:
+          mode: NOTIFY_THEN_TERMINATE
+        timeout: 172800
+jobEnvironments:
```
karthikbekalp: I think we can skip the jobEnvironments as its just code for collecting logs.

test/integ/test_scenes/physical_custom_fps/expected_job_bundle/parameter_values.yaml
```diff
@@ -0,0 +1,33 @@
...
+- name: DetailedLogging
+  value: '0'
+- name: UseCachedText
+  value: '0'
+- name: ChunkSize
+  value: 1
+- name: TargetChunkDuration
+  value: 0
+- name: Frames
+  value: '15'
```
karthikbekalp: Can we have multiple frames rendered to verify that FPS indeed was an issue?

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -279,7 +279,7 @@ def start_render(self, data: dict) -> None:
         start_frame, end_frame = self._parse_frame_range(frame_value)
         self.render_kwargs[FRAME_KEY] = start_frame
 
-        fps = self.doc.GetFps()
+        fps = self.render_data[c4d.RDATA_FRAMERATE]
```
karthikbekalp, line 282: Hmm, just to confirm, we have tested that this was an actual bug correct? https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/issues/372#issuecomment-3676129673 

We have had only 1 customer complain about this so its important that we confirm. :)

test/integ/test_scenes/physical_custom_fps/scene/scene.py
```diff
@@ -0,0 +1,70 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+import os
+import sys
+
+import c4d
+from deadline.cinema4d_submitter.integ_test_helpers import internal_create_job_bundle
+
+
+def main():
```
karthikbekalp, line 9: This goes back to my question earlier but can we run the same test without the fix and see if we get a different output ?

test/integ/test_scenes/physical_custom_fps/scene/scene.py
```diff
@@ -0,0 +1,70 @@
...
+    # Set render settings
+    render_data = doc.GetActiveRenderData()
+    render_data[c4d.RDATA_PATH] = "renders/$prj"
+    render_data[c4d.RDATA_RENDERENGINE] = 1023342  # physical
+    render_data[c4d.RDATA_FORMAT] = c4d.FILTER_PNG
+    render_data[c4d.RDATA_MULTIPASS_SAVEIMAGE] = False
+
+    # Uncheck "Use Project Frame Rate" and set render FPS to 12
+    render_data[c4d.RDATA_FRAMERATE_SYNC_WITH_PROJECT] = False
+    render_data[c4d.RDATA_FRAMERATE] = 12.0
```
karthikbekalp, line 49: Hmm, is it required to keep this enabled? I wonder if this could be part of the issue?

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_cinema4d_handler.py
```diff
@@ -853,9 +853,16 @@ def test_start_render_with_chunk_range(
...
+        def mock_getitem(self, key):
+            if key == c4d.RDATA_FRAMERATE:
+                return 24
+            return MagicMock()
+
+        mock_render_data.__getitem__ = mock_getitem
         mock_render_data.GetDataInstance = Mock()
         mock_doc.GetActiveRenderData.return_value = mock_render_data
-        mock_doc.GetFps.return_value = 24
+        mock_doc.GetFps.return_value = 30
```
karthikbekalp: Hmm, a little confusing. Do we need this mock now that we don't use GetFps at all?

karthikbekalp (changes requested):

test/integ/test_scenes/physical_custom_fps/scene/scene.py
gianelli-lagos-aws: Yes! Without the fix the Integration test fails. You end up with a different output.

test/integ/test_scenes/physical_custom_fps/scene/scene.py
gianelli-lagos-aws: In order to have a custom Frame Rate, you need to uncheck this checkbox then you can enter a different frame rate in Render Settings.

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_cinema4d_handler.py
gianelli-lagos-aws: After the fix, this unit test would fail. I kept the mock intentionally at a different value (30) to prove the code doesn't use GetFps(). If someone reverts the fix, the test would fail because it'd pick up 30 instead of the expected 24 from RDATA_FRAMERATE. Happy to remove it if you think it's clearer without!

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
diane-hj: Yes :) This was reproducible on my end.

gianelli-lagos-aws: Yes!  My mentor reproduced the issue, and I verified it through the integration test: with doc.GetFps() (the bug), the test renders the wrong frame (frame 6 instead of 15) and produces a different output image. With RDATA_FRAMERATE (the fix), it renders the correct frame and the output matches expected. The integ test fails without the fix and passes with it.

test/integ/test_scenes/physical_custom_fps/expected_job_bundle/parameter_values.yaml
gianelli-lagos-aws: Yes, I can work on that!

diane-hj: +1. I think you could modify the scene script to add a keyframed animation on the cube (e.g., rotation or position change over the frame range) and render multiple frames.
Something like(but please validate if it works): 
```
track = cube.FindCTrack(c4d.DescID(c4d.ID_BASEOBJECT_REL_POSITION, c4d.VECTOR_X))
curve = track.GetCurve()
curve.AddKey(c4d.BaseTime(0, fps))["key"].SetValue(curve, 0)
curve.AddKey(c4d.BaseTime(30, fps))["key"].SetValue(curve, 100)
```

test/integ/test_scenes/physical_custom_fps/expected_job_bundle/template.yaml
gianelli-lagos-aws: Sounds good!

test/integ/test_scenes/physical_custom_fps/expected_job_bundle/parameter_values.yaml
gianelli-lagos-aws: On lines 18-35 of the scene.py, I have something similar to this! Is this what you are referring to? I think what I could add is the multiple frames part.

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_cinema4d_handler.py
karthikbekalp: >  kept the mock intentionally at a different value (30) to prove the code doesn't use GetFps().  If someone reverts the fix

I think it would be clearer without as currently there is no GetFps call. So if someone were to look at the code they would be confused as I was why we were mocking GetFps. :)

diane-hj (dismissed):
w00t w00t!

karthikbekalp (approved):

crowecawcaw (approved):

PR #480, opened 2026-07-10
feat: Run pre-GUI hooks in the Cinema 4D render submitter

leon-li-inspire:

The Cinema 4D submitter did not run pre-GUI submission hooks, so studios had no way to
pre-populate the submit dialog's fields (job name/description, queue parameters, `deadline:`
job properties) before it opens. deadline-cloud added a Qt-free pre-GUI hook entry point
(`run_pre_gui_hooks` / `PreGuiHookContext`) plus a generic `apply_pre_gui_output` helper in
[aws-deadline/deadline-cloud#1255](https://github.com/aws-deadline/deadline-cloud/pull/1255)
so DCC submitters (Maya, Nuke, and now Cinema 4D) can share one integration. This PR wires
that into Cinema 4D, matching the equivalent Maya and Nuke changes.

- In `_show_submitter`, before building `SubmitJobToDeadlineDialog`, call
  `run_pre_gui_hooks(PreGuiHookContext(bundle_dir=None, submitter_name="cinema4d", ...))`.
  Cinema 4D has no on-disk job bundle at pre-GUI time, so hooks are sourced from
  `DEADLINE_HOOKS_DIR` only (`bundle_dir=None`), gated by `settings.allow_environment_hooks`.
- The confirmation prompt uses `qt_hook_confirmation(parent)`, skipped when
  `settings.auto_accept` is set.
- Apply the merged output with deadline-cloud's generic `apply_pre_gui_output`.
  `RenderSubmitterUISettings` has no `.parameters` list, so `name`/`description` land on the
  settings object and every hook parameter flows into the dialog's shared parameter values.
- Bump the `deadline` floor to `>= 0.60.1` (the release that first ships
  `deadline.client.ui.pre_gui_hooks`; the 0.60.0 tag does not contain the module).

No behavior change when no pre-GUI hooks are configured: the merged output is empty, so the
dialog is seeded with the same `CondaPackages` value as before. The new path only takes effect
when `DEADLINE_HOOKS_DIR` is set and environment hooks are enabled.

- **Unit tests:** Added `test/unit/deadline_submitter_for_cinema4d/test_pre_gui_hooks.py` —
  headless tests exercising the DCC-relevant mapping contract (name/description onto settings;
  all hook parameters flowing to shared values since C4D has no `.parameters` list; empty,
  falsy, and partial output; the `auto_accept` confirm-callback branch). These pass against the
  real merged `apply_pre_gui_output` from #1255.
- **Integration test (xa11y):** Added `test/integ_xa11y/test_cinema4d.py::test_pre_gui_hook` —
  the GUI counterpart to the unit tests. It launches a real Cinema 4D with `DEADLINE_HOOKS_DIR`
  pointing at a fixture hook (and `allow_environment_hooks` / `auto_accept` enabled), drives the
  shipped submitter dialog end to end via xa11y, exports a job bundle, and asserts the hook's
  output survived into it — proving the `run_pre_gui_hooks` + `apply_pre_gui_output` wiring holds
  through an actual export, not just in isolation. The fixture hook emits a fixed name,
  description, and `deadline:priority`; the test asserts `template.yaml` carries
  `name: PREGUI RAN` / `description: populated by pre-GUI hook` and `parameter_values.yaml`
  carries `deadline:priority: 88`.
- **Ran it on a Windows + Cinema 4D 2026 workstation: `1 passed`.** The submitter dialog was
  driven through to Export and the exported bundle contained exactly those three hook values.
- `black --check`, `ruff check`, and `py_compile` all pass.

Docstrings/comments added at the hook call site, in the new unit-test module, and in the
integration test; `test/AGENTS.md` documents the new pre-GUI hook case (fixture layout, the two
settings it enables, and that it uses targeted asserts rather than a golden-bundle diff). No
README/adaptor schema changes are needed — this is submitter-only and does not touch
init-data/run-data.

No. This is additive and backward compatible (no adaptor interface change). It does raise the
`deadline` dependency floor to `>= 0.60.1`, so this PR should only merge once deadline-cloud
0.60.1 (which ships `deadline.client.ui.pre_gui_hooks`) is available to the build — the same
dependency the equivalent Maya and Nuke pre-GUI PRs carry.

----

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline[gui] >= 0.59, < 0.61",
+    "deadline[gui] >= 0.60.1, < 0.61",
     "openjd-adaptor-runtime >= 0.7,< 0.10",
     # fonttools is Windows-only due to Cinema 4D technical limitations
     "fonttools >=4.59.2, <4.64; sys_platform == 'win32'",
@@ -38,7 +38,7 @@ dependencies = [
 
 [project.optional-dependencies]
 gui = [
-    "deadline[gui] >= 0.59, < 0.61",
+    "deadline[gui] >= 0.60.1, < 0.61",
     "PySide6-Essentials == 6.8.3",  # Use 6.8 to align with VFXP 2026: https://vfxplatform.com/#reference-platform
 ]
 
--- a/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
+++ b/src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
@@ -4,6 +4,7 @@ import os
 import re
 import shutil
 import tempfile
+from contextlib import contextmanager
 from copy import deepcopy
 from dataclasses import dataclass
 from pathlib import Path
@@ -14,14 +15,21 @@ import yaml  # type: ignore[import]
 from qtpy import QtWidgets
 from qtpy.QtCore import Qt  # type: ignore[attr-defined]
 
+from deadline.client.config import get_setting, str2bool
 from deadline.client.dataclasses import SubmitterInfo
-from deadline.client.exceptions import DeadlineOperationError
+from deadline.client.exceptions import DeadlineOperationCanceled, DeadlineOperationError
 from deadline.client.job_bundle._yaml import deadline_yaml_dump
 from deadline.client.job_bundle.parameters import JobParameter
 from deadline.client.job_bundle.submission import AssetReferences
 from deadline.client.ui.dialogs.submit_job_to_deadline_dialog import (  # pylint: disable=import-error
     JobBundlePurpose,
     SubmitJobToDeadlineDialog,
+)
+from deadline.client.ui.pre_gui_hooks import (
+    PreGuiHookContext,
+    apply_pre_gui_output,
+    qt_hook_confirmation,
+    run_pre_gui_hooks,
 )
 
 from ._version import version_tuple as adaptor_version_tuple
@@ -110,6 +118,10 @@ def show_submitter():
                 )
             else:
                 w = _show_submitter(temp_dir, None)
+            # _show_submitter returns None when the user declines the pre-GUI hook confirmation
+            # prompt; treat that as a normal cancellation and skip showing the dialog.
+            if w is None:
+                return
             w.setStyleSheet(C4D_STYLE)
             w.exec_()
     except Exception:  # noqa: BLE001 - prevent UI failures from escaping into Cinema 4D
@@ -574,6 +586,106 @@ def save_job_bundle_files(
         deadline_yaml_dump(asset_references.to_dict(), f, indent=1)
 
 
+def _restore_pre_gui_hook_sticky_settings(
+    settings: RenderSubmitterUISettings,
+    pre_gui_hook_sticky_reset: dict[str, tuple[Any, Any]] | None,
+) -> dict[str, Any]:
+    """Restore pre-hook baseline values for sticky fields a pre-GUI hook overwrote.
+
+    ``pre_gui_hook_sticky_reset`` maps each sticky field a pre-GUI hook overwrote to
+    ``(pre_hook_value, hook_value)``. For each, restore the pre-hook value unless the user changed
+    the field in the dialog -- in that case the current value differs from what the hook applied,
+    so we keep the user's value. Hook output therefore stays scoped to the session: it never
+    persists as a stale default after the hook is disabled, nor feeds back as ``jobName`` on the
+    next launch.
+
+    Returns the ``{field_name: hook_value}`` it reset to baseline, so the caller can reapply the
+    hook values after the sticky write and avoid leaving the live ``settings`` object mutated (see
+    :func:`_pre_gui_hook_sticky_baseline`).
+    """
+    if not pre_gui_hook_sticky_reset:
+        return {}
+    restored: dict[str, Any] = {}
+    for field_name, (baseline_value, hook_value) in pre_gui_hook_sticky_reset.items():
+        if getattr(settings, field_name) == hook_value:
+            setattr(settings, field_name, baseline_value)
+            restored[field_name] = hook_value
+    return restored
+
+
+# deadline-cloud's shared-settings tab (SharedJobPropertiesWidget) writes these "deadline:*" job
+# properties onto the matching sticky RenderSubmitterUISettings fields when it gathers settings on
+# Export/Submit. A pre-GUI hook only ever puts them in the shared parameter values (not directly on
+# the settings object), so a hook emitting e.g. deadline:priority would otherwise persist into the
+# scene's sticky settings exactly as an unscoped name/description would. This maps each such shared
+# key to its sticky field so both are scoped out of the sticky write together.
+_HOOK_STICKY_SHARED_PARAM_FIELDS = {
+    "deadline:priority": "priority",
+    "deadline:targetTaskRunStatus": "initial_status",
+    "deadline:maxFailedTasksCount": "max_failed_tasks_count",
+    "deadline:maxRetriesPerTask": "max_retries_per_task",
+    "deadline:maxWorkerCount": "max_worker_count",
+}
+
+
+def _compute_pre_gui_hook_sticky_reset(
+    settings: RenderSubmitterUISettings,
+    name_description_baseline: dict[str, Any],
+    shared_before: dict[str, Any],
+    shared_after: dict[str, Any],
+) -> dict[str, tuple[Any, Any]]:
+    """Build the ``{sticky_field: (pre_hook_value, hook_value)}`` map of every sticky field a
+    pre-GUI hook changed, so create_job_bundle can keep the hook's values out of the sticky write.
+
+    Two routes a hook reaches a sticky field:
+
+    * ``name`` / ``description`` are set directly on ``settings`` by ``apply_pre_gui_output`` —
+      detected by diffing ``name_description_baseline`` (snapshotted before the hook) against the
+      current value.
+    * ``deadline:*`` job properties are routed into the shared parameter values instead; the shared
+      settings tab later writes them onto the matching sticky field (see
+      ``_HOOK_STICKY_SHARED_PARAM_FIELDS``). Each such key the hook added or changed
+      (``shared_before`` -> ``shared_after``) is recorded against its field, with the hook's value
+      so a later user edit in the dialog (current value != hook value) is still kept.
+
+    The baseline for a shared-param field is the field's current value on ``settings``, which
+    ``apply_pre_gui_output`` does not touch — so it is still the pre-hook (scene/sticky) value.
+    """
+    reset: dict[str, tuple[Any, Any]] = {}
+    for attr, baseline in name_description_baseline.items():
+        current = getattr(settings, attr)
+        if current != baseline:
+            reset[attr] = (baseline, current)
+    for shared_key, field_name in _HOOK_STICKY_SHARED_PARAM_FIELDS.items():
+        if shared_key in shared_after and shared_after[shared_key] != shared_before.get(shared_key):
+            reset[field_name] = (getattr(settings, field_name), shared_after[shared_key])
+    return reset
+
+
+@contextmanager
+def _pre_gui_hook_sticky_baseline(
+    settings: RenderSubmitterUISettings,
+    pre_gui_hook_sticky_reset: dict[str, tuple[Any, Any]] | None,
+):
+    """Scope the pre-GUI hook sticky-settings reset to the ``save_sticky_settings`` call only.
+
+    Restores pre-hook baselines for the sticky write, then reapplies the hook values so the live
+    ``settings`` object is unchanged on exit. ``on_create_job_bundle_callback`` runs once per
+    Export/Submit against the dialog's *live* settings object (not a per-call copy), so permanently
+    resetting ``name``/``description`` to baseline would make a second action in the same session
+    build its job from the scene-derived value instead of the hook's -- two different job names for
+    two identical clicks, with the dialog still showing the first. Keeping the reset scoped to the
+    sticky write leaves the in-memory settings exactly as the user saw them, while the persisted
+    sticky settings still exclude the hook output.
+    """
+    restored = _restore_pre_gui_hook_sticky_settings(settings, pre_gui_hook_sticky_reset)
+    try:
+        yield
+    finally:
+        for field_name, hook_value in restored.items():
+            setattr(settings, field_name, hook_value)
+
+
 def create_job_bundle(
     settings: RenderSubmitterUISettings,
     takes: dict[str, list[TakeData]],
@@ -583,6 +695,7 @@ def create_job_bundle(
     attachments: AssetReferences,
     temp_dir: str | None = None,
     host_requirements: dict | None = None,
+    pre_gui_hook_sticky_reset: dict[str, tuple[Any, Any]] | None = None,
 ) -> dict[str, Any]:
     """
     Creates a job bundle and saves sticky settings for rendering.
@@ -671,7 +784,11 @@ def create_job_bundle(
     settings.input_filenames = sorted(attachments.input_filenames)
     settings.input_directories = sorted(attachments.input_directories)
 
-    settings.save_sticky_settings(Scene.name())
+    # Keep pre-GUI hook output out of the persisted sticky settings (the job template above already
+    # captured the effective name/description). Scope the reset to the sticky write only, so the
+    # live settings object keeps the hook values for any later action in the same dialog session.
+    with _pre_gui_hook_sticky_baseline(settings, pre_gui_hook_sticky_reset):
+        settings.save_sticky_settings(Scene.name())
 
     return {
         "known_asset_paths": [
@@ -943,6 +1060,19 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
     asset_references.input_filenames = temp_assets
 
 
+def _pre_gui_hook_confirm_callback(parent):
+    """Choose the confirmation callback for pre-GUI hooks based on the auto_accept setting.
+
+    Returns ``None`` (run hooks without prompting) when ``settings.auto_accept`` is enabled,
+    otherwise the standard Qt confirmation dialog from ``qt_hook_confirmation``. Kept as a small
+    helper so the auto_accept branch can be unit-tested headlessly.
+    """
+    if str2bool(get_setting("settings.auto_accept")):
+        return None
+
+    return qt_hook_confirmation(parent)
+
+
 def _show_submitter(temp_dir: str, parent=None, f=Qt.WindowType.Tool):  # type: ignore[call-overload]
     """
     Creates and returns a submission dialog for rendering jobs.
@@ -991,6 +1121,11 @@ def _show_submitter(temp_dir: str, parent=None, f=Qt.WindowType.Tool):  # type: 
         additional_info=additional_info,
     )
 
+    # Maps each sticky field a pre-GUI hook overwrites -> (pre-hook value, hook value). Populated
+    # after the hooks run below and passed to create_job_bundle, which uses it to keep hook output
+    # out of the persisted sticky settings. See the block near apply_pre_gui_output.
+    pre_gui_hook_sticky_reset: dict[str, tuple[Any, Any]] = {}
+
     def on_create_job_bundle_callback(
         widget: SubmitJobToDeadlineDialog,
         job_bundle_dir: str,
@@ -1026,14 +1161,61 @@ def _show_submitter(temp_dir: str, parent=None, f=Qt.WindowType.Tool):  # type: 
             widget.job_attachments.attachments,
             temp_dir,
             host_requirements,
+            pre_gui_hook_sticky_reset=pre_gui_hook_sticky_reset,
         )
+
+    shared_parameter_values = {
+        "CondaPackages": conda_packages,
+    }
+
+    # Run pre-GUI hooks so studios can pre-populate dialog fields before it opens. Cinema 4D has
+    # no on-disk job bundle at this point, so hooks are sourced from DEADLINE_HOOKS_DIR only
+    # (bundle_dir=None), gated by settings.allow_environment_hooks. The confirmation prompt is
+    # skipped when auto_accept is set; otherwise the standard dialog is shown.
+    try:
+        pre_gui_output = run_pre_gui_hooks(
+            PreGuiHookContext(
+                bundle_dir=None,
+                job_name=render_settings.name,
+                submitter_name="cinema4d",
+                parameters=dict(shared_parameter_values),
+            ),
+            confirm_callback=_pre_gui_hook_confirm_callback(parent),
+        )
+    except DeadlineOperationCanceled:
+        # The user declined the hook confirmation prompt. This is a normal cancellation, not an
+        # error, so abort opening the submitter silently by returning None; show_submitter skips
+        # the dialog. Without this, the exception would surface as a spurious "Deadline UI launch
+        # failed" error for what is a deliberate "No" click.
+        return None
+    # RenderSubmitterUISettings has no `.parameters` list, so apply_pre_gui_output routes
+    # name/description onto it and every hook parameter into shared_parameter_values.
+    # run_pre_gui_hooks returns {} when no hooks run (and raises DeadlineOperationCanceled, handled
+    # above, if the user declines); `or {}` is defensive against any future contract change so the
+    # common no-hooks path can never pass a falsy value into apply_pre_gui_output.
+    #
+    # name/description (and deadline:* job properties routed through shared_parameter_values) are
+    # sticky settings (data_classes.py). Snapshot the pre-hook (scene/sticky) state first, then
+    # record what the hook actually changed so create_job_bundle can scope those to this session
+    # (keep them out of the sticky write). This stops hook output from persisting as a stale default
+    # once the hook is disabled and from feeding back on the next launch, while any edit the user
+    # makes in the dialog still persists. See _compute_pre_gui_hook_sticky_reset.
+    name_description_baseline = {
+        attr: getattr(render_settings, attr) for attr in ("name", "description")
+    }
+    shared_values_before_hook = dict(shared_parameter_values)
+    apply_pre_gui_output(pre_gui_output or {}, render_settings, shared_parameter_values)
+    pre_gui_hook_sticky_reset = _compute_pre_gui_hook_sticky_reset(
+        render_settings,
+        name_description_baseline,
+        shared_values_before_hook,
+        shared_parameter_values,
+    )
 
     submitter_dialog = SubmitJobToDeadlineDialog(
         job_setup_widget_type=SceneSettingsWidget,
         initial_job_settings=render_settings,
-        initial_shared_parameter_values={
-            "CondaPackages": conda_packages,
-        },
+        initial_shared_parameter_values=shared_parameter_values,
         auto_detected_attachments=auto_detected_attachments,
         attachments=attachments,
         on_create_job_bundle_callback=on_create_job_bundle_callback,
--- a/test/integ/fixtures/pregui_hooks/pregui_hook.py
+++ b/test/integ/fixtures/pregui_hooks/pregui_hook.py
@@ -1,0 +1,56 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""Pre-GUI hook fixture for the Cinema 4D xa11y integ test.
+
+deadline-cloud runs this as a subprocess before the submitter dialog is built (see
+``deadline.client.ui.pre_gui_hooks.run_pre_gui_hooks``). It receives the job metadata as JSON on
+stdin and returns the merged pre-GUI output as JSON on stdout: ``name`` / ``description`` land on
+the submitter's settings object, and every key under ``parameters`` flows into the dialog's shared
+parameter values (``deadline:priority`` sets the Priority field).
+
+The test asserts the exported job bundle reflects exactly these values, proving the C4D submitter
+wires ``run_pre_gui_hooks`` + ``apply_pre_gui_output`` correctly (PR #480). This case has no
+``expected/job_bundle/`` golden dir; the assertions compare against the ``_HOOK_*`` constants in
+``test/integ/test_cinema4d.py`` instead, so keep the emitted values in sync with those constants.
+
+The test also sets ``DEADLINE_CLOUD_PREGUI_MARKER``; when present this script writes that file, so
+the test can assert the hook actually ran (separately from asserting its output was applied).
+"""
+
+import json
+import os
+import sys
+
+# Consume the metadata C4D passes on stdin (jobName, parameters, submitterName, ...). We do not
+# branch on it here — the point of the fixture is a deterministic, asserted output — but reading it
+# keeps the subprocess contract honest (a real hook would use it). A malformed/empty stdin is not a
+# failure for this fixture: it still emits the fixed output below, so we deliberately ignore a
+# decode error rather than abort.
+try:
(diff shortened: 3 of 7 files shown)


src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -24,6 +25,12 @@
     JobBundlePurpose,
     SubmitJobToDeadlineDialog,
 )
+from deadline.client.ui.pre_gui_hooks import (  # pylint: disable=import-error
```
leon-li-inspire: why do we need pylint here ?

pyproject.toml
```diff
@@ -30,15 +30,19 @@ classifiers = [
 ]
 
 dependencies = [
-    "deadline[gui] >= 0.59, < 0.61",
+    # run_pre_gui_hooks / PreGuiHookContext / apply_pre_gui_output are required for the pre-GUI
+    # hook integration in the render submitter; they first ship in deadline-cloud 0.60.1 (the
```
leon-li-inspire: Remove this comment as this comment is redundant.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
leon-li-inspire: Good catch — removed. The `# pylint: disable=import-error` wasn't needed here; dropped it from the `pre_gui_hooks` import in `cc8b9a7`.

pyproject.toml
leon-li-inspire: Removed the comment in `cc8b9a7`. The floor stays `deadline[gui] >= 0.60.1, < 0.61` (0.60.1 is the first release that ships `deadline.client.ui.pre_gui_hooks`).

src/deadline/cinema4d_submitter/submitter_api.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@dataclass
+class Cinema4DSubmitterSettings(SubmitterSettings):
+    """Cinema 4D-specific submission settings."""
+
+    renderer: str = ""
+    take_name: str = ""
+    image_width: int = 1920
+    image_height: int = 1080
```
karthikbekalp: Any reason we are explicitly setting the image width and height here?

test/unit/deadline_submitter_for_cinema4d/test_pre_gui_hooks.py
```diff
@@ -0,0 +1,137 @@
...
+from typing import Optional
+from unittest.mock import patch
+
+from deadline.client.ui.pre_gui_hooks import apply_pre_gui_output
+
+from deadline.cinema4d_submitter import cinema4d_render_submitter
+from deadline.cinema4d_submitter.data_classes import RenderSubmitterUISettings
+
+
+def _settings() -> RenderSubmitterUISettings:
```
karthikbekalp, line 28: Thanks for the unit tests, but can we add some integration tests (using xally here?)

src/deadline/cinema4d_submitter/submitter_api.py
```diff
@@ -0,0 +1,141 @@
...
+            os.path.join(settings.project_path, settings.name) if settings.project_path else ""
+        )
+        settings.input_filenames = [scene_file] if scene_file else []
+
+        rd = doc.GetActiveRenderData()
+        if rd:
+            fps = doc.GetFps()
+            start_frame = rd[c4d.RDATA_FRAMEFROM].GetFrame(fps)
+            end_frame = rd[c4d.RDATA_FRAMETO].GetFrame(fps)
+            settings.frame_list = f"{start_frame}-{end_frame}"
```
karthikbekalp: Hmm, I think this part of the code was changed recently? We don't use `doc.GetFps` anymore? https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/commit/be8608b597766adfcfdab190edcfe1405b987093#diff-5305ababdfc27d07f7d36587fa79909c9adefeb8a9c76430813f09b407e34900

karthikbekalp (changes requested):
Thanks for adding hooks to Cinema 4D. 
Reviewed the PR. I added some comments.  I think it would be nice to add more integration tests to ensure we are not breaking anything. 
(Also remember to run the integration tests both xally and the current one to ensure we are not breaking any existing flows).

test/unit/deadline_submitter_for_cinema4d/test_pre_gui_hooks.py
leon-li-inspire: Thanks! Agreed an xa11y integration test would add value here. The four unit tests deliberately cover only the DCC-owned logic that runs headless — `apply_pre_gui_output` mapping onto `RenderSubmitterUISettings`, and the `auto_accept` -> confirm-callback selection. Exercising the full pre-GUI flow (env hooks loaded from `DEADLINE_HOOKS_DIR`, the confirmation dialog, and the merged output landing in the opened submitter) needs a running Cinema 4D, which is exactly what the `test/integ_xa11y` harness drives.

I'd prefer to add that as a focused follow-up rather than in this PR: it needs a hooks-dir fixture + an xa11y case asserting the dialog is pre-populated (and the decline path), and I'd want to validate it against a real C4D on the Windows runner rather than land an unverified GUI test. Tracking it as a follow-up — let me know if you'd rather block this PR on it.

src/deadline/cinema4d_submitter/submitter_api.py
```diff
@@ -0,0 +1,141 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Optional
+
+import yaml
+
+import c4d  # type: ignore[import]
+
+from deadline.client.job_bundle.submission import AssetReferences
+from deadline.client.submitter_api import SubmitterAPI, SubmitterSettings
```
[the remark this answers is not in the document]
leon-li-inspire: Confirmed — verified against the published `deadline 0.60.1` wheel and mainline: there is no `deadline/client/submitter_api.py`, so `from deadline.client.submitter_api import SubmitterAPI, SubmitterSettings` raises ModuleNotFoundError. As you note, nothing imports this module yet, so the existing flow isn't broken.

Scope note: `submitter_api.py` is part of the separate unified-submitter-API effort (branch `feature/unified-submitter-api`), which this pre-GUI-hook branch was cut from — so those commits rode along into this PR. The `SubmitterAPI`/`SubmitterSettings` base class isn't released yet, so this file can't land until deadline-cloud ships it and the floor is bumped. I'd suggest dropping the `submitter_api.py` commits from this PR so it's purely the pre-GUI hook change, and letting them land with the unified-submitter-API work. Will confirm the preferred split before rebasing.

src/deadline/cinema4d_submitter/submitter_api.py
```diff
@@ -0,0 +1,141 @@
...
+        # The "Frames" job parameter has minLength: 1 in the job template, so an
+        # empty frame range (e.g. no active RenderData) would violate the
+        # constraint and fail CreateJob. Only emit it when non-empty.
+        if settings.frame_list:
+            parameter_values.append({"name": "Frames", "value": settings.frame_list})
+
+        # NOTE: The take is passed to the render via the "TAKE" stepEnvironment
+        # variable in the job template, not as a job parameter. The template
+        # defines no "Take" parameter, so emitting one would fail CreateJob with
+        # an unknown-parameter ValidationException.
```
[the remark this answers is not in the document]
leon-li-inspire: Confirmed the finding: `get_settings` captures `take_name` / `output_path` but `get_job_template` never writes them into the `Cinema4D` step environment (`TAKE` stays the template default), so the selected take/output would be dropped. This is part of the unified-submitter-API work (`submitter_api.py`), separate from this PR's pre-GUI hook feature — flagging it to be fixed with that effort rather than here. See the import-scope note above re: splitting these commits out.

src/deadline/cinema4d_submitter/submitter_api.py
leon-li-inspire: This is in the unified-submitter-API code (`submitter_api.py`), not the pre-GUI hook change in this PR. The defaults (1920x1080) are overwritten from `RDATA_XRES`/`RDATA_YRES` when render data exists; whether the API should carry image dimensions at all is a design question for the unified-submitter-API work. Deferring to that effort's owner rather than changing it here.

src/deadline/cinema4d_submitter/submitter_api.py
leon-li-inspire: Good catch — confirmed `submitter_api.py:47` still uses `doc.GetFps()`, whereas mainline's `scene.py` now reads `RDATA_FRAMERATE` (commit be8608b) for the FPS used in frame mapping. This should be aligned to `RDATA_FRAMERATE`, but it's in the unified-submitter-API code (`submitter_api.py`), separate from this PR's pre-GUI hook feature, so I'd fix it with that effort. See the scope note on the import comment above.

test/integ/fixtures/pregui_hooks/pregui_hook.py
```diff
@@ -0,0 +1,34 @@
...
+in sync with ``expected/job_bundle/`` for the ``pregui_hook`` case.
+"""
+
+import json
+import sys
+
+# Consume the metadata C4D passes on stdin (jobName, parameters, submitterName, ...). We do not
+# branch on it here — the point of the fixture is a deterministic, asserted output — but reading it
+# keeps the subprocess contract honest (a real hook would use it).
+try:
```
leon-li-inspire, line 28: What's the goal for this block?

test/integ_xa11y/test_cinema4d.py
```diff
@@ -0,0 +1,770 @@
...
+            _actual_render_dir(actual_dir),
+        )
+
+    # Clean up if the test was successful
+    rmtree(actual_dir, ignore_errors=True)
+
+
+# The directory holding the pre-GUI hook fixture (hooks.yaml + pregui_hook.py). Pointed at by
+# DEADLINE_HOOKS_DIR when launching C4D for the hook case; gated by settings.allow_environment_hooks.
+_PREGUI_HOOKS_DIR = Path(__file__).parent / "fixtures" / "pregui_hooks"
```
leon-li-inspire: Will this path be cleaned up after each test run ?

test/integ_xa11y/test_cases/pregui_hook/input/scene.py
leon-li-inspire, line 19: Is this file still useful ?

test/integ_xa11y/test_cinema4d.py
```diff
@@ -0,0 +1,770 @@
...
+    """
+    case = "pregui_hook"
+    case_folder, actual_dir = _prepare_actual_dir(test_cases_folder_location, case)
+
+    scene_path = _build_cinema4d_scene(cinema4d_location, case_folder, actual_dir, case)
+
+    # Enable the env-hook path in the config the fixture wrote, and point C4D at the hooks dir.
+    _enable_environment_hooks(deadline_farm["env_overlay"])
+
+    _export_job_bundle_via_submitter(
```
leon-li-inspire: Have we tested other cases for hook conditions, such as there is `DEADLINE_HOOKS_DIR` but hooks is not allowed ?

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -933,6 +944,19 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
     asset_references.input_filenames = temp_assets
 
 
+def _pre_gui_hook_confirm_callback(parent):
```
leon-li-inspire, line 1063: Seems like this method not only check whether we allow run pre-gui hooks also check whether we allow to run pre-submission and post-submission hooks right ?

leon-li-inspire: Good question, and worth being precise about. `_pre_gui_hook_confirm_callback` itself is **pre-GUI only** — it reads `settings.auto_accept` and returns either `None` (run the pre-GUI hook without prompting) or `qt_hook_confirmation(parent)` for the pre-GUI confirmation dialog. It does **not** read `allow_environment_hooks`/`allow_bundle_hooks`, and it plays no part in the pre-/post-submission flow.

The nuance you're picking up on is the *setting* it checks: `settings.auto_accept` is global, so the same flag also suppresses the **submission** confirmation dialog (the one that lists pre- and post-submission hooks) over in deadline-cloud core's `create_job_from_job_bundle`. So there are two separate confirmation dialogs — pre-GUI (shown here, in the C4D submitter) and submission (shown by core after Submit) — and `auto_accept` governs both, but this method only wires the pre-GUI one.

Whether hooks run *at all* is a different gate (`allow_environment_hooks` / `allow_bundle_hooks`), enforced by core's `collect_*_hook_sources`, not here.

test/integ/fixtures/pregui_hooks/pregui_hook.py
leon-li-inspire: It consumes the job metadata C4D passes on stdin (jobName, parameters, submitterName, …) so the hook honors the real subprocess contract — a production hook would read and branch on it. This fixture is deliberately deterministic (it emits a fixed, asserted output regardless), so it just reads and discards.

Addressed the CodeQL `py/empty-except` flag on this block in 37fb84f: narrowed `except Exception` → `except (json.JSONDecodeError, ValueError)` and added an explanatory comment in the handler (empty/malformed stdin isn't a failure for this fixture, so the decode error is intentionally ignored).

test/integ_xa11y/test_cinema4d.py
leon-li-inspire: `_PREGUI_HOOKS_DIR` is a static, committed fixture directory (`test/integ/fixtures/pregui_hooks/`, holding `hooks.yaml` + `pregui_hook.py`) — it's a test **input**, not runtime output, so there's nothing to clean up between runs.

The runtime output that *is* cleaned is the case's `actual/` dir: `test_pre_gui_hook` ends with `rmtree(actual_dir, ignore_errors=True)` on success, and (matching the rest of the suite) leaves it in place on failure so the exported bundle can be inspected.

test/integ_xa11y/test_cases/pregui_hook/input/scene.py
leon-li-inspire: Yes — it's required. `scene.py` runs under c4dpy to build and save `pregui_hook.c4d`, which the test-only sidecar plugin then loads (via `DEADLINE_CLOUD_SCENE_PATH`) before opening the submitter. The submitter needs an active document with a real path to open against.

It mirrors the `cube`/standard cases' `scene.py` but omits the render-data setup, because the pre-GUI hook case only exports a bundle and asserts its metadata — it never renders.

test/integ_xa11y/test_cinema4d.py
leon-li-inspire: Not in an integ test yet — good catch, it's a coverage gap worth a follow-up. This case only exercises the **enabled** path (`allow_environment_hooks=true` + `auto_accept=true`).

Two conditions aren't covered here:
- **`DEADLINE_HOOKS_DIR` set but `allow_environment_hooks=false`** → hooks must NOT run and the dialog fields stay at defaults (a negative assertion). Not yet tested at the integ level.
- **Declined confirmation** (`auto_accept=false`, user clicks No) → `DeadlineOperationCanceled`, submitter aborts. This one *is* covered headlessly in `test_pre_gui_hooks.py` (the confirm-callback tests), just not via xa11y.

I'll track adding a disabled-path integ case as a follow-up. Separately, integ coverage for the pre-/post-**submission** hooks is blocked on the shared mock backend not implementing a `CreateJob` route (those hooks only fire on the real Submit path), which is why the whole suite presses Export bundle — also a follow-up.

test/integ/fixtures/pregui_hooks/pregui_hook.py
```diff
@@ -0,0 +1,34 @@
...
+    json.load(sys.stdin)
+except Exception:
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in 37fb84f: narrowed the bare `except Exception` to `except (json.JSONDecodeError, ValueError)` and added an explanatory comment in the handler, so this is no longer an empty/undocumented except. Behavior is unchanged — the fixture still emits its fixed output whether or not stdin carries valid JSON.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -1018,12 +1042,41 @@ def on_create_job_bundle_callback(
...
+                submitter_name="cinema4d",
+                parameters=dict(shared_parameter_values),
+            ),
+            confirm_callback=_pre_gui_hook_confirm_callback(parent),
+        )
+    except DeadlineOperationCanceled:
+        # The user declined the hook confirmation prompt. This is a normal cancellation, not an
+        # error, so abort opening the submitter silently by returning None; show_submitter skips
+        # the dialog. Without this, the exception would surface as a spurious "Deadline UI launch
+        # failed" error for what is a deliberate "No" click.
```
karthikbekalp, line 1189: Declining the hook prompt makes _show_submitter return None, so the submitter silently disappears. Could we show an actionable cancellation message explaining how to deactivate environment hooks, or allow continuing without hooks? 
Or because this is an opt-in feature this is the expected flow?

karthikbekalp (commented):
Just 1 curious comment.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -933,6 +944,19 @@ def export_to_temp_folder(temp_dir: str, asset_references: AssetReferences) -> N
...
 
 
+def _pre_gui_hook_confirm_callback(parent):
+    """Choose the confirmation callback for pre-GUI hooks based on the auto_accept setting.
+
+    Returns ``None`` (run hooks without prompting) when ``settings.auto_accept`` is enabled,
+    otherwise the standard Qt confirmation dialog from ``qt_hook_confirmation``. Kept as a small
+    helper so the auto_accept branch can be unit-tested headlessly.
+    """
+    if str2bool(get_setting("settings.auto_accept")):
```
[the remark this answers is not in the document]
leon-li-inspire: Yup, it is intentional to execute the hook confirmation without user approval again if the `settings.auto_accept` is true.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
leon-li-inspire: It's the expected flow that the submitter disappear if user cancel the hook confirmation prompt .

karthikbekalp: Cool.

test/integ/fixtures/pregui_hooks/hooks.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Pre-GUI hook manifest for the Cinema 4D xa11y integ test. Discovered via DEADLINE_HOOKS_DIR
+# (gated by settings.allow_environment_hooks); version "1.0" is the schema the installed
+# deadline-cloud validator accepts. The submitter runs the preGUI hook before the dialog opens.
+version: "1.0"
+preGUI:
+  - command: python
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `4186f8b`. You're right that `python` isn't a portable interpreter and that a resolution failure masquerades as an output-wiring failure. deadline-cloud resolves a hook `command` via abs-path / hooks-dir / `shutil.which` with no env-var expansion, so a static manifest can't name a portable interpreter. The test now generates `hooks.yaml` per-run with `command = sys.executable` (absolute, always resolves; the hook is stdlib-only), and `pregui_hook.py` writes a marker the test asserts up front — so "hook never launched" is now distinct from "hook ran but output wasn't applied."

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -1018,12 +1042,41 @@ def on_create_job_bundle_callback(
...
+        # error, so abort opening the submitter silently by returning None; show_submitter skips
+        # the dialog. Without this, the exception would surface as a spurious "Deadline UI launch
+        # failed" error for what is a deliberate "No" click.
+        return None
+    # RenderSubmitterUISettings has no `.parameters` list, so apply_pre_gui_output routes
+    # name/description onto it and every hook parameter into shared_parameter_values.
+    # run_pre_gui_hooks returns {} when no hooks run (and raises DeadlineOperationCanceled, handled
+    # above, if the user declines); `or {}` is defensive against any future contract change so the
+    # common no-hooks path can never pass a falsy value into apply_pre_gui_output.
+    apply_pre_gui_output(pre_gui_output or {}, render_settings, shared_parameter_values)
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `1f96eea`. The submitter snapshots the pre-hook `name`/`description`, applies the hook output, records which of those the hook actually changed, and — via the new `_pre_gui_hook_sticky_baseline` context manager in `create_job_bundle` — restores those baselines for the `save_sticky_settings` write only (unless the user edited the field in the dialog, in which case their value is what persists). So hook output no longer lands in `<scene>.deadline_render_settings.json`, and it no longer feeds back as `jobName` on the next launch.

test/integ/fixtures/pregui_hooks/pregui_hook.py
```diff
@@ -0,0 +1,54 @@
...
+# Signal that this hook actually executed. The integ test points DEADLINE_CLOUD_PREGUI_MARKER at a
+# path it checks after export, so a run that yields the wrong bundle can be told apart as "the hook
+# never launched" (discovery / interpreter resolution failure) rather than "the hook ran but its
+# output wasn't wired in". Best-effort: marker I/O must never break the hook's stdout contract.
+_marker = os.environ.get("DEADLINE_CLOUD_PREGUI_MARKER")
+if _marker:
+    try:
+        with open(_marker, "w", encoding="utf-8") as _fh:
+            _fh.write("ran")
+    except OSError:
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `1f96eea` — added an explanatory comment to the best-effort marker-write `except OSError` in `pregui_hook.py`, so it's no longer an undocumented empty `except`: the marker is best-effort diagnostics and a filesystem error writing it must not change the hook's stdout contract, so the error is intentionally ignored.

test/integ/test_cinema4d.py
```diff
@@ -186,6 +193,34 @@ def _build_launch_env(
...
+    ``settings.allow_environment_hooks`` is ``true``; both default to ``false``.
+    ``settings.auto_accept`` = ``true`` makes ``_pre_gui_hook_confirm_callback``
+    return ``None`` so hooks run without the Qt confirmation prompt — the prompt's
+    behaviour is covered by the unit tests, and skipping it keeps this UI test
+    focused on the export-output contract. Both live in the plain ``[settings]``
+    section (no profile prefix), so we read-modify-write that section on the config
+    file named by ``DEADLINE_CONFIG_FILE_PATH`` in the launch env overlay.
+    """
+    config_path = Path(env_overlay["DEADLINE_CONFIG_FILE_PATH"])
+    config = ConfigParser()
```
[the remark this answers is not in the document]
leon-li-inspire: Already addressed in an earlier revision — `_enable_environment_hooks` no longer hand-rolls a `ConfigParser` edit. It writes through `deadline.client.config.set_setting` (with `DEADLINE_CONFIG_FILE_PATH` pointed at the temp config), exactly as suggested, which resolves both the `%`-interpolation re-escaping and the hardcoded `[settings]`-section assumption. Resolving.

test/integ/fixtures/pregui_hooks/pregui_hook.py
```diff
@@ -0,0 +1,37 @@
...
+
+deadline-cloud runs this as a subprocess before the submitter dialog is built (see
+``deadline.client.ui.pre_gui_hooks.run_pre_gui_hooks``). It receives the job metadata as JSON on
+stdin and returns the merged pre-GUI output as JSON on stdout: ``name`` / ``description`` land on
+the submitter's settings object, and every key under ``parameters`` flows into the dialog's shared
+parameter values (``deadline:priority`` sets the Priority field).
+
+The test asserts the exported job bundle reflects exactly these values, proving the C4D submitter
+wires ``run_pre_gui_hooks`` + ``apply_pre_gui_output`` correctly (PR #480). Keep the emitted values
+in sync with ``expected/job_bundle/`` for the ``pregui_hook`` case.
```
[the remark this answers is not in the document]
leon-li-inspire: Already addressed in an earlier revision — the current `pregui_hook.py` docstring reads: *"This case has no `expected/job_bundle/` golden dir; the assertions compare against the `_HOOK_*` constants in `test/integ/test_cinema4d.py` instead, so keep the emitted values in sync with those constants."* So it points maintainers at the `_HOOK_*` constants (matching `test_pre_gui_hook` and `test/AGENTS.md`); the `expected/job_bundle/` sync target is gone. Resolving.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -1018,12 +1042,41 @@ def on_create_job_bundle_callback(
...
+    try:
+        pre_gui_output = run_pre_gui_hooks(
+            PreGuiHookContext(
+                bundle_dir=None,
+                job_name=render_settings.name,
+                submitter_name="cinema4d",
+                parameters=dict(shared_parameter_values),
+            ),
+            confirm_callback=_pre_gui_hook_confirm_callback(parent),
+        )
```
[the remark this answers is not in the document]
leon-li-inspire: Partly fixed in `1f96eea`, partly deferred with reasoning:

- **Parent:** the confirmation and failure prompts now get a top-level, stay-on-top parent on Windows (`_pre_gui_hook_prompt_parent`). Cinema 4D's UI isn't Qt, so — unlike gui-submit / Maya / Nuke, which parent to the host's `QMainWindow` — there's genuinely no main window to parent to here; this mirrors the `WindowStaysOnTopHint` workaround the submitter dialog itself already uses, so the modal surfaces above Cinema 4D instead of behind it.
- **Running hooks before `setup_auto_detected_attachments`:** deliberately left as-is. Maya and Nuke also run their pre-GUI hooks *after* asset detection, and this submitter is intentionally structured to parallel them — reordering only Cinema 4D would break that cross-DCC consistency. Happy to do it as a coordinated change across all three integrations if you'd prefer that over the parallelism.

Final window-stacking behavior is verified on the Windows + Cinema 4D box.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -671,6 +704,11 @@ def create_job_bundle(
     settings.input_filenames = sorted(attachments.input_filenames)
     settings.input_directories = sorted(attachments.input_directories)
 
+    # Keep pre-GUI hook output out of the persisted sticky settings (see the helper); the job
+    # template above already captured the effective name/description, so this only scopes the
+    # sticky write.
+    _restore_pre_gui_hook_sticky_settings(settings, pre_gui_hook_sticky_reset)
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `1f96eea`. `_restore_pre_gui_hook_sticky_settings` no longer leaves the live settings object mutated: `create_job_bundle` now wraps `save_sticky_settings` in the `_pre_gui_hook_sticky_baseline` context manager, which restores the baseline for the write and then **reapplies the hook values on exit** (even if the write raises). The sticky write still excludes the hook output, but the in-memory object keeps the hook values — so a second Export/Submit in the same dialog session builds the same job name the user saw, not the scene-derived one. Unit tests cover the restore-then-reapply, user-edit, and write-raises paths.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -1026,14 +1082,58 @@ def on_create_job_bundle_callback(
...
+        pre_gui_output = run_pre_gui_hooks(
+            PreGuiHookContext(
+                bundle_dir=None,
+                job_name=render_settings.name,
+                submitter_name="cinema4d",
+                parameters=dict(shared_parameter_values),
+            ),
+            confirm_callback=_pre_gui_hook_confirm_callback(parent),
+        )
+    except DeadlineOperationCanceled:
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `1f96eea`. `_show_submitter` now catches `DeadlineOperationError` separately from the user-decline `DeadlineOperationCanceled`: on a hook failure (non-zero exit, timeout, malformed JSON) it logs a warning, shows a best-effort `QMessageBox` naming the failure, and opens the submitter **with no hook output** rather than letting the error reach `show_submitter`'s blanket handler. Pre-GUI hooks are an optional pre-population convenience, so a hook failure is no longer fatal — one typo in a studio-deployed `hooks.yaml` can't silently block all Cinema 4D submission anymore.

(The same single-`except` pattern exists in the Maya/Nuke submitters; worth a matching follow-up in those repos.)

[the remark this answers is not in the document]
leon-li-inspire: Update: reverted this change (in `bbb3f6e`). On further review the original behavior is intentional — a pre-GUI hook failure should surface rather than be silently swallowed — so `_show_submitter` again catches only `DeadlineOperationCanceled` and lets any other hook error propagate. Not changing this in this PR.

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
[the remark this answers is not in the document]
leon-li-inspire: Update: reverted this change (in `bbb3f6e`). The existing behavior is intentional — the confirmation dialog keeps using the caller's `parent`, without the stay-on-top wrapper widget. Not changing this in this PR.

karthikbekalp (dismissed):

src/deadline/cinema4d_submitter/cinema4d_render_submitter.py
```diff
@@ -1026,14 +1112,58 @@ def on_create_job_bundle_callback(
...
+    # name/description are sticky settings (data_classes.py). Snapshot the pre-hook (scene/sticky)
+    # values first, then record any the hook actually changed so create_job_bundle can scope those
+    # to this session (keep them out of the sticky write). This stops hook output from persisting
+    # as a stale default once the hook is disabled and from feeding back as `jobName` on the next
+    # launch, while any edit the user makes to the field in the dialog still persists.
+    hook_sticky_baseline = {
+        attr: getattr(render_settings, attr) for attr in ("name", "description")
+    }
+    apply_pre_gui_output(pre_gui_output or {}, render_settings, shared_parameter_values)
+    pre_gui_hook_sticky_reset = {
```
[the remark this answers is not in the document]
leon-li-inspire: Fixed in `6a19f1e`. You're right — the scoping was incomplete. `deadline:*` job properties flow into `shared_parameter_values`, and deadline-cloud's `SharedJobPropertiesWidget` (`set_parameter_value` → widgets → `update_settings`) writes them back onto the sticky fields on Export/Submit — `priority` ← `deadline:priority`, `initial_status` ← `deadline:targetTaskRunStatus`, and `max_failed_tasks_count`/`max_retries_per_task`/`max_worker_count` ← their matching keys — so `save_sticky_settings` persisted them.

The reset now covers all of those via `_HOOK_STICKY_SHARED_PARAM_FIELDS`: I snapshot the shared values before `apply_pre_gui_output` and record each mapped key the hook added/changed against its sticky field (storing the hook value so a later user edit in the dialog is still kept). It's applied only around `save_sticky_settings`, so the submitted job's priority — which comes from `queue_parameters`, not `settings.priority` — is unaffected. The reset-map construction is extracted into `_compute_pre_gui_hook_sticky_reset` with unit tests, including the `deadline:priority` path the integ fixture exercises.

karthikbekalp (approved):

PR #508, opened 2026-08-03
fix: bake OCIO view transform on non-tile renders

ansata2001:

Non-tile renders submitted through the adaptor come out **un-tone-mapped** (dark / "Raw" look) when the scene uses OCIO color management (the default in Cinema 4D 2026). The scene's configured **View Transform** (e.g. `ACES 1.0 SDR-video`) is not applied to the saved 8-bit image.

Root cause: `c4d.documents.RenderDocument`'s internal save does **not** bake the OCIO view transform into the output bitmap — a Cinema 4D SDK limitation. The `Commandline` renderer and Picture Viewer save do apply it, so artists' local renders look correct while farm renders (which use the adaptor → `RenderDocument`) come back dark.

The **tile path already works around this** — `tile_rendering.finalize_tile_render` disables the render-time bake, then bakes the view transform manually with `BakeOcioViewToBitmap` (added in #401). The normal full-frame path in `Cinema4DHandler.start_render` was never given the same treatment, so ordinary renders are affected.

Apply the same, already-proven workaround to non-tile renders:

- Before the render, disable `RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER` so the rendered bitmap holds render-space data, and snapshot the output directory.
- After the render, `BakeOcioViewToBitmap(bm, rd, SAVEBIT_NONE)`, reset the display/view color profiles, and overwrite **only the beauty file** C4D wrote.

The new `tile_rendering.bake_full_frame_beauty` reuses the tile path's helpers and, importantly, leaves the **alpha (`A_`) file, multi-pass files, and float/EXR output untouched** (EXR must stay scene-linear). It is guarded to Cinema 4D **2025.2+** and **8-bit output only**, matching the tile path's guards.

Non-tile renders now match tile renders (and local `Commandline`/Picture Viewer output): the scene's OCIO view transform is baked into 8-bit deliverables. Behavior for tile renders, EXR/float output, alpha, and multi-pass is unchanged.

- **Locally (c4dpy):** rendering a scene with `View Transform = ACES 1.0 SDR-video` via `RenderDocument` produced a mid-gray of `127`; with this fix the same frame produces `165` — matching the `Commandline` renderer and the scene's intended look. Verified the alpha file stays a clean matte and the native output filename is preserved.
- **End-to-end on a live worker:** submitted through the C4D submitter; the beauty frame came back correctly tone-mapped instead of dark.


leon-li-inspire (commented):
Thanks for closing the OCIO view-transform gap on the full-frame path — the overall shape is right: reuse the tile helpers, guard to 8-bit / 2025.2+, and leave alpha / multi-pass / EXR untouched. A few things I'd like addressed before merge.

### CI is red — formatting *(must fix to merge)*

All 12 `Python (...)` matrix jobs are failing, and it's a single root cause: the **lint** step (`hatch run lint` → `black --check --diff .`) finds `src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py` isn't black-formatted. `ruff` passes; only `black` fails. (The jobs that show as canceled after ~20s were stopped by fail-fast once the first job failed — same cause.)

Two expressions now fit on one line at the repo's 100-char limit and black wants them collapsed:

```diff
-                pre_existing_outputs = {
-                    os.path.join(beauty_dir, f) for f in os.listdir(beauty_dir)
-                }
+                pre_existing_outputs = {os.path.join(beauty_dir, f) for f in os.listdir(beauty_dir)}
```
```diff
-            tile_rendering.bake_full_frame_beauty(
-                bm, rd, self.render_data, pre_existing_outputs
-            )
+            tile_rendering.bake_full_frame_beauty(bm, rd, self.render_data, pre_existing_outputs)
```

**Fix:** run `hatch run fmt` (auto-formats), then `hatch run lint` to confirm clean, and push. That turns all 12 jobs green. Worth doing alongside the CR-4 tests below, since the same `hatch run` steps cover both.

---

Below, **CR-N ("Change Request N")** are just numbered labels so we can refer to each item in follow-up discussion. Severity is called out per item:

| # | Item | Severity |
|---|------|----------|
| CR-1 | Multi-frame chunks get corrupted by the save loop | **Blocking** |
| CR-2 | Task retries leave the output un-baked (still dark) | **Blocking** |
| CR-3 | `A_` / multi-pass exclusion relies on filename heuristics | Non-blocking |
| CR-4 | No unit tests were added (repo policy requires them) | **Blocking** |

---

### CR-1 — Multi-frame chunks get corrupted *(Blocking)*

The adaptor job template exposes **Frames Per Chunk** (`ChunkSize`) and **Target Chunk Duration**, so a single task can render a range like `10-20`. `_parse_frame_range` + `RDATA_FRAMEFROM`/`RDATA_FRAMETO` already support this, and `RenderDocument` writes one file per frame — but `bm` holds only the **last** frame's pixels. The save loop then re-saves *every* new matching file with that one bitmap, silently corrupting the sequence. (The tile path never hit this because a tile task is always one frame.)

**What to change:** guard the bake to the single-frame case — pass `start_frame`/`end_frame` into `bake_full_frame_beauty` and early-return (with a `print(...)` noting the skip) when `start_frame != end_frame`; or only bake when exactly one new beauty file is found, and warn otherwise. A cleaner long-term fix is to re-open/bake/re-save each written beauty file per frame, but a single-frame guard is an acceptable minimum since chunk size 1 is the default. This corresponds to the bot's inline finding on the save loop.

### CR-2 — Retries leave output un-baked *(Blocking)*

`pre_existing_outputs` is snapshotted *before* the render. On a task retry into a directory that already holds this frame's file, C4D overwrites it in place; because the path is already in the snapshot, `full in pre_existing` is `True` and the bake is **skipped** — reproducing the exact dark image this PR fixes, permanently across retries.

**What to change:** identify the output by the frame's known filename (derive it from `RDATA_PATH` + frame number + extension), or by an mtime captured immediately before `RenderDocument` (bake files with `mtime >= t0`), rather than a pre/post directory diff.

### CR-3 — `A_` / multi-pass exclusion is a name heuristic *(Non-blocking)*

`fn.startswith("A_")` and the multi-pass-basename check are fragile if a user's beauty filename legitimately starts with `A_`. Once CR-1/CR-2 switch to a *known-filename* match, you can bake only the exact beauty path and drop these heuristics entirely — simpler and more robust.

### CR-4 — Add unit tests *(Blocking, repo policy)*

Every code change needs tests (see `AGENTS.md`). The harness already exists:

- `test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/__init__.py` stubs `c4d`/`maxon` via `sys.modules[...] = MagicMock()`, so `import c4d` needs no real Cinema 4D.
- `TestStartRenderChunkRange.test_start_render_with_chunk_range` in `test_cinema4d_handler.py` is a ready-made template to clone (mock `doc`/`render_data`, patch `RenderDocument`/`MultipassBitmap`/`BaseTime`, drive a chunk range).
- `bake_full_frame_beauty` can also be tested directly — it takes plain args (`bm`, `rd`, `render_data`, `pre_existing`); just patch `os.listdir`/`os.path`.

Please cover at least: multi-frame chunk does not corrupt (CR-1 regression), single frame is baked once, retry still bakes (CR-2), non-8-bit/EXR is skipped, and alpha/multi-pass files are left untouched.

---

### Extra question *(not blocking)*

Unlike the tile path — which saves `orig_bake_flag` and restores it in `finalize_tile_render` — the non-tile path sets `RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER = False` but never restores it. Since the document is reused across frames in a session, should this be restored too, or is it harmless because it's re-set on every `start_render`?

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
```diff
@@ -509,3 +509,67 @@ def assemble_tiles(
...
+        full = os.path.join(beauty_dir, fn)
+        if full in pre_existing:
+            continue
+        if not fn.lower().endswith(ext.lower()):
+            continue
+        if fn.startswith("A_"):
+            continue  # alpha channel file — leave as-is
+        if mp_base and fn.startswith(mp_base):
+            continue  # multi-pass file — leave as-is
+        bm.Save(full, save_filter)
```
leon-li-inspire: **CR-1 (blocking) — multi-frame chunks get corrupted here.** `bm` holds only the last frame's pixels, but on a chunked task (`ChunkSize` > 1, e.g. frame range `10-20`) `RenderDocument` wrote one file per frame. This loop re-saves *every* new beauty file with that single bitmap, overwriting each rendered frame with the same image. Guard to the single-frame case: pass `start_frame`/`end_frame` in and early-return when they differ, or only bake when exactly one new beauty file is found (warn otherwise).

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
```diff
@@ -509,3 +509,67 @@ def assemble_tiles(
...
+    ext, save_filter = get_format_info(render_data[c4d.RDATA_FORMAT])
+
+    # Names to exclude: the C4D alpha file ("A_" prefix) and multi-pass files.
+    mp_base = ""
+    if render_data[c4d.RDATA_MULTIPASS_SAVEIMAGE] and render_data[c4d.RDATA_MULTIPASS_FILENAME]:
+        mp_base = os.path.basename(os.path.splitext(render_data[c4d.RDATA_MULTIPASS_FILENAME])[0])
+
+    for fn in os.listdir(beauty_dir):
+        full = os.path.join(beauty_dir, fn)
+        if full in pre_existing:
```
leon-li-inspire: **CR-2 (blocking) — retries skip the bake.** `pre_existing` is snapshotted *before* the render (see cinema4d_handler.py:317). On a task retry into a directory that already holds this frame's file, C4D overwrites it in place, but the path is already in `pre_existing`, so `full in pre_existing` is `True` and the bake is skipped — leaving the exact dark image this PR fixes, permanently across retries. Match by known output filename (`RDATA_PATH` + frame + ext) or by mtime captured just before `RenderDocument`, instead of a pre/post directory diff.

src/deadline/cinema4d_adaptor/Cinema4DClient/tile_rendering.py
```diff
@@ -509,3 +509,67 @@ def assemble_tiles(
...
+            continue
+        if not fn.lower().endswith(ext.lower()):
+            continue
+        if fn.startswith("A_"):
```
leon-li-inspire: **CR-3 (non-blocking) — name heuristic is fragile.** `fn.startswith("A_")` (and the multi-pass basename check below) misfire if a user's beauty filename legitimately starts with `A_`. Once CR-1/CR-2 switch to a known-filename match, bake only the exact beauty path and drop these heuristics entirely.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -309,6 +309,20 @@ def start_render(self, data: dict) -> None:
             bm = bitmaps.MultipassBitmap(width, height, c4d.COLORMODE_RGB)
         rd = self.render_data.GetDataInstance()
 
+        # Non-tile OCIO workaround: RenderDocument's internal save does not bake the
+        # OCIO View Transform (Cinema 4D SDK bug), so ordinary renders are written
+        # un-tone-mapped. Disable the render-time bake so the rendered bitmap holds
+        # render-space data we can bake manually after the render, and snapshot the
+        # output directory so we only overwrite the beauty image THIS render produces.
+        pre_existing_outputs: set = set()
```
leon-li-inspire: **CR-2 root cause.** This pre-render snapshot is what makes retries skip the bake (see the matching comment in tile_rendering.py). Capturing a pre-render timestamp here instead, or deriving the exact expected output filename, would be more robust than diffing the directory before/after.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -309,6 +309,20 @@ def start_render(self, data: dict) -> None:
...
+        if not is_tile_render and hasattr(c4d, "RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER"):
+            rd[c4d.RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER] = False
```
leon-li-inspire: **Extra question (not blocking).** Unlike the tile path — which saves `orig_bake_flag` and restores it in `finalize_tile_render` — this sets `RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER = False` but never restores it. Since the document is reused across frames in a session, should it be restored, or is it harmless because it's re-set on every `start_render`?

leon-li-inspire (commented):
Adding inline anchors so each CR item points at the exact line. Full rationale is in my earlier review summary above; CR-4 (add unit tests) has no single line so it stays there.

ansata2001:
Thanks for the review — both were real bugs. Fixed in the follow-up commit:

**Multi-frame corruption** — the non-tile path now renders one frame per `RenderDocument` call (for the 8-bit output that needs baking), so each frame's bitmap is baked into its own file. A single `RenderDocument` over a range leaves only the last frame in the bitmap, so earlier frames couldn't be baked correctly.

**Retry safety** — the beauty file is now located via C4D's own token resolver (`FilenameConvertTokens`) anchored on the resolved output base, plus newest modification time >= the render start — instead of a pre/post directory diff. A task retry that overwrites a file in place is still baked.

Also moved the helpers into a new `ocio_bake.py` (they're non-tile code, so `tile_rendering.py` is back to upstream), render into a float (`RGBf`) bitmap to keep the OCIO bake at full precision, and added unit tests in `test_ocio_bake.py`.

leon-li-inspire (commented):
Thanks for the thorough revision — this addresses the earlier findings well:

- **CR-1 (multi-frame corruption): resolved**, and better than the minimum I suggested — rendering one frame per `RenderDocument` call and baking each into its own file is the right fix, so chunked tasks now render *and* bake every frame.
- **CR-2 (retries un-baked): resolved** — token-resolved output base + newest-mtime-since-render-start is retry-safe, and `test_stale_file_not_baked` covers it.
- **CR-3 (name heuristics): resolved** — nice call moving this into `ocio_bake.py` and reverting `tile_rendering.py` to upstream; alpha/multi-pass exclusion by resolved base is much cleaner.
- **CR-4 (tests): added** — `test_ocio_bake.py` covers the important cases.
- The earlier bake-flag-restore question is moot now that the flag is set fresh each `start_render`.

Two things left before this is merge-ready — details inline:

1. **CI will fail lint** on `test_ocio_bake.py` (black). I verified locally with black at the repo's 100-char limit — the two source files are clean, only the test file needs reformatting. Heads-up that the Python matrix hasn't actually run on this revision yet (the workflows show `action_required` — a maintainer needs to approve the run for this fork PR), so the red won't appear until then. `hatch run fmt` fixes it now.
2. **Per-frame loop is 8-bit-only** — a behavioral question about the text-cache path, inline.

Neither is a correctness problem with the OCIO bake itself; the core fix looks right.

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
```diff
@@ -303,33 +315,59 @@ def start_render(self, data: dict) -> None:
...
-            raise RuntimeError("Error: unhandled render result: %s" % result)
-        if result != c4d.RENDERRESULT_OK:
-            raise RuntimeError("Error: render result: %s" % result_description)
-
-        # Post-render tile processing: OCIO bake, crop, save tile, restore paths
-        if is_tile_render and tile_ctx is not None:
-            tile_rendering.finalize_tile_render(bm, rd, tile_ctx, self.render_data, start_frame)
+        if bake_ocio:
+            self.render_data[c4d.RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER] = False
+            for frame in range(start_frame, end_frame + 1):
```
leon-li-inspire: The per-frame render loop is the right fix for CR-1. Two things worth confirming, since this changes the control flow for multi-frame chunks:

1. **Two divergent render strategies now coexist.** The 8-bit path renders frame-by-frame (N `RenderDocument` calls over the chunk), while the `else` branch (float/EXR, tiles) still renders the whole range in a single call as before. That's defensible — only 8-bit needs the per-frame bake — but it does mean chunked EXR and chunked 8-bit now exercise materially different code. Is that intended, and is the float/EXR range-render still known-good for chunks > 1?

2. **Text-cache handling.** `_cache_text_if_needed` is called once (line 320) for `start_frame` only, before the loop — matching the old one-frame-per-`start_render` model. Now that a single call can render frames `start..end` in-process, do frames after the first still get correct text caching / document reload, or can a chunk > 1 with animated-then-collapsed text render stale frames? If the per-frame text logic needs to run inside the loop, that would be a separate fix.

Not blocking the OCIO change itself — just want to make sure chunked (ChunkSize > 1) submissions are exercised, ideally with an integ or unit test that drives a 2+ frame range through this branch.

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_ocio_bake.py
```diff
@@ -0,0 +1,109 @@
...
+        bake_full_frame_beauty(bm, _rd(), _render_data(self._base(tmp_path)), MagicMock(), 5, start)
+        bm.Save.assert_not_called()
+        c4d.documents.BakeOcioViewToBitmap.assert_not_called()
+
+    def test_non_8bit_is_noop(self, tmp_path):
+        start = time.time()
+        _touch(tmp_path / "render0005.jpg", start + 10)
+        bm = MagicMock()
+        bake_full_frame_beauty(
+            bm, _rd(depth=c4d.RDATA_FORMATDEPTH_32), _render_data(self._base(tmp_path)), MagicMock(), 5, start
```
leon-li-inspire: This line is >100 chars, so `black --check` (the CI `hatch run lint` step) will fail on this file — the same lint failure that was red on the first revision, just relocated here. `cinema4d_handler.py` and `ocio_bake.py` are already black-clean; only this test file needs it. `hatch run fmt` reformats it to:

```python
        bake_full_frame_beauty(
            bm,
            _rd(depth=c4d.RDATA_FORMATDEPTH_32),
            _render_data(self._base(tmp_path)),
            MagicMock(),
            5,
            start,
        )
```

ansata2001:
Pushed a follow-up (`c547883`) addressing the latest round:

**Multi-pass exclusion edge case** (bot, `ocio_bake.py`) — fixed. A beauty file is now treated as multi-pass only when the multi-pass base is a *longer* (more specific) prefix than the beauty base. This handles both directions: beauty `render_beauty` / mp `render` no longer skips the beauty file, and beauty `render` / mp `render_mp` still excludes the multi-pass file. Added a regression test.

**Restoring `RDATA_BAKE_OCIO_VIEW_TRANSFORM_RENDER`** — done. It's now saved and restored in a `try/finally` around the per-frame loop, mirroring how the tile path saves/restores it in `finalize_tile_render`.

**Two divergent render strategies / EXR chunks** — the `else` branch (float/EXR and tiles) is unchanged from the previous behavior: same render-data instance, same bitmap, same flags, single range-render call. Only the 8-bit path renders frame-by-frame, since only 8-bit display output needs the view transform baked. So chunked EXR is exactly as before this PR.

**Text caching for chunks > 1** — this is unchanged from the current behavior too: `_cache_text_if_needed` is still called once for the start frame before rendering, exactly as the single-call range render did before. A chunk > 1 renders with text pinned to the start frame in both the old and new code, so this isn't a regression from the per-frame change — per-frame text caching would be a separate improvement (and applies equally to the EXR path).

test/unit/deadline_adaptor_for_cinema4d/Cinema4DClient/test_ocio_bake.py
ansata2001: Ok, this one is fixed

src/deadline/cinema4d_adaptor/Cinema4DClient/ocio_bake.py
```diff
@@ -0,0 +1,152 @@
...
+    # correct extension, not a multi-pass file, and modified at/after this render's
+    # start (never re-bakes a stale file; a retry that overwrites in place is caught).
+    target = None
+    target_mtime = render_start_time
+    for fn in os.listdir(beauty_dir):
+        if not fn.startswith(beauty_prefix):
+            continue
+        if not fn.lower().endswith(ext.lower()):
+            continue
+        if mp_prefix and mp_prefix != beauty_prefix and fn.startswith(mp_prefix):
```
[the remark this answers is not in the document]
ansata2001: Fixed

src/deadline/cinema4d_adaptor/Cinema4DClient/cinema4d_handler.py
ansata2001: Please see the comment below.

leon-li-inspire (approved):

PR #513, opened 2026-08-10
test: Run integration tests across Cinema 4D versions (2024,2025,2026)

karthikbekalp:

We need to run integration tests on all the versions of Cinema 4d that we currently support and not just 2026. 

Updated the tests to run on all the versions that we support i.e. 2024 to 2026. 

We still have issues with Redshift tests failing because of a Maxon bug so marked those tests as XFAIL so that even if it flakes it does not break the workflow. 

We have higher coverage and make sure our changes are backwards compatible. 

Ran the tests here: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/actions/runs/31415004267/job/93541820105

No documentation changes required. Only test changes. 

No. 

----


karthikbekalp:
Resolving this in favor of : https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/515

